PR #12, opened 2018-08-09
Revamp Initializable and remove Migratable

frangio:
Fixes https://github.com/zeppelinos/zos/issues/2.

Enhances `Initializable` to make it usable with multiple inheritance, with the purpose of removing `Migratable` (which was our current solution) because it leads to a confusing developer experience.

Marking `in progress` because I should update the docs to reflect the new usage, but the code and tests can already be reviewed.

- [x] Fix contracts
- [x] Fix rest of test suite
- [x] Document changes


packages/lib/test/contracts/Initializable.test.js
```diff
@@ -0,0 +1,80 @@
...
+        await this.contract.initializeNested();
+      });
+
+      it('initializer has run', async function () {
+        assert.isTrue(await this.contract.initializerRan());
+      });
+    });
+  });
+
+  describe('complex testing with inheritance', function () {
```
spalladino: Can we add another suite where mother inherits from gramps as well, so we test a diamond-like inheritance graph?

frangio:
Rebased on `next`.

frangio:
@spalladino There's something we didn't think through: if we're removing `Migratable`, what should we do about `upgradeAndCall`? I apparently broke the tests for `upgradeAndCall` because they were using a `Migratable` contract as the mock implementation. This makes sense because the function you call via `upgradeAndCall` needs some mechanism to prevent it running more than once, and `Migratable` was that mechanism.

For the record, we had agreed on removing `Migratable` on the basis that the need for migrations really hasn't come up yet and we don't want to design it in advance of knowing the requirements of a real situation.

What do you suggest we do?

spalladino:
@frangio `upgradeAndCall` still makes sense, the same way `createAndCall` makes sense even if we didn't provide any `Initializable` options. It now becomes the user's responsibility to ensure that a migration method is only called once, since we are not imposing any "best practice" to do so. We can keep the `upgradeAndCall` tests which call a regular `migrate` method, that simply relies on a `hasBeenMigrated` flag managed manually, instead of an `isMigration(,,)` modifier.

frangio:
@spalladino I think the only error remaining is one in `complex-example` which I'm not sure how to fix. (To be honest I haven't looked into it yet but it looks kinda hairy.)

```
Error: Could not find zos-lib/contracts/migrations/Migratable.sol from any sources; imported from openzeppelin-zos/contracts/ownership/Ownable.sol
```

frangio:
I added `Migratable` back in because due to Truffle's wrong handling of nested dependencies, there was no way to make the `openzeppelin-zos` package in `lib-complex` compile.

I removed the bump to 2.0.0-alpha because that wasn't going to fix anything.

One last thing. Currently `Initializable` is in the `contracts/migrations/` directory. We cannot rename the directory because of the same reason we had to keep `Migratable`, but we could move `Initializable` somewhere else. @spalladino What do you think?

spalladino:
+1 on moving Initializable to the root of Contracts, while keeping the
deprecated Migratable in its old location

PR #39, opened 2018-08-23
Status command for multiple dependencies

spalladino:
Re-enables the `status` command that had been disabled for multiple dependencies on #17 .

### DIFF
--- a/packages/cli/src/scripts/status.js
+++ b/packages/cli/src/scripts/status.js
@@ -10,7 +10,7 @@ export default async function status({ network, txParams = {}, networkFile = und
 
   if (!(await rootInfo(controller))) return;
   if (!(await versionInfo(controller.networkFile))) return;
-  await stdlibInfo(controller.networkFile);
+  await depsInfo(controller.networkFile);
   await contractsInfo(controller);
   await proxiesInfo(controller.networkFile);
 }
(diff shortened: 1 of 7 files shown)


packages/cli/src/scripts/status.js
```diff
@@ -81,44 +80,46 @@ async function contractsInfo(controller) {
     .forEach(contractAlias => log.warn(`- ${contractAlias} will be removed on next push`));
 }
 
-async function stdlibInfo(networkFile) {
+async function depsInfo(networkFile) {
```
jcarpanelli: As we are using `dependencies`/`dependency` everywhere, we should rename this to `dependenciesInfo` for consistency.

packages/cli/src/scripts/status.js
```diff
@@ -10,7 +10,7 @@ export default async function status({ network, txParams = {}, networkFile = und
 
   if (!(await rootInfo(controller))) return;
   if (!(await versionInfo(controller.networkFile))) return;
-  await stdlibInfo(controller.networkFile);
+  await depsInfo(controller.networkFile);
```
jcarpanelli: `dependenciesInfo`

packages/cli/src/scripts/status.js
spalladino: @jcarpanelli updated!

PR #47, opened 2018-08-27
Change admin command

### LINKED ISSUE #3: New CLI command to change admin ownership of a proxy
Add a new command (`transfer`?) to the CLI, that allows delegating ownership of a proxy to another address. 

The command should use the same convention as `update` to indicate which proxy/proxies are to be transferred: _Provide the [alias] you added your contract with, or use --all flag to upgrade all. If no [address] is provided, all instances of that contract class will be upgraded_.

Given that this is a potentially irreversible operation, add a confirmation before executing (but after listing all the proxies that would be transferred). Add support for a `-y/--yes` flag to skip confirmation.

Once the proxy was transferred, add a new attribute `owner` to it in the `zos.network.json` file. All subsequent calls to `update` should skip any proxies with an owner different than the current app.

spalladino:
Add new `set-admin` command to change admin ownership of a proxy. The command requires a `-y` or `--yes` flag to be actually run, otherwise it cowardly refuses. It can be run either with an address, or a contract name (optionally namespaced by a package). For instance:

```bash
$ zos session ...
$ zos set-admin MyContract $NEWADMIN -y
$ zos set-admin OpenZeppelin/ERC20 $NEWADMIN -y
$ zos set-admin 0x40 $NEWADMIN -y
```

Fixes #3

### DIFF
--- a/packages/cli/src/commands/set-admin.js
+++ b/packages/cli/src/commands/set-admin.js
@@ -1,0 +1,37 @@
+'use strict';
+
+import setAdmin from '../scripts/set-admin'
+import runWithTruffle from '../utils/runWithTruffle'
+import { fromContractFullName } from '../utils/naming'
+import _ from 'lodash'
+
+const name = 'set-admin'
+const signature = `${name} [alias-or-address] [new-admin-address]`
+const description = 'change upgradeability admin of a contract instance. Provide the [alias] or [package]/[alias] you added your contract with, or its [address]. Note that if you transfer to an incorrect address, you may irreversibly lose control over upgrading your contract.'
+
+const register = program => program
+  .command(signature, { noHelp: true })
+  .usage('[alias-or-address] [new-admin-address] --network <network> [options]')
+  .description(description)
+  .option('-y, --yes', 'accept transferring admin rights (required)')
+  .withNetworkOptions()
+  .action(action)
+
+async function action(contractFullNameOrAddress, newAdmin, options) {
+  const { yes } = options
+  if (!yes) {
+    throw Error("This is a potentially irreversible operation: if you specify an incorrect admin address, you may lose the ability to upgrade your contract forever.\nPlease double check all parameters, and run the same command with --yes.")
+  }  
+
+  let proxyAddress, contractAlias, packageName
+  if (contractFullNameOrAddress && contractFullNameOrAddress.startsWith('0x')) {
+    proxyAddress = contractFullNameOrAddress
+  } else if (contractFullNameOrAddress) {
+    ({ contract: contractAlias, package: packageName } = fromContractFullName(contractFullNameOrAddress))
+  }
+  
+  const args = _.pickBy({ contractAlias, packageName, proxyAddress, newAdmin })
+  await runWithTruffle(async (opts) => await setAdmin({ ... args, ... opts }), options)
+}
+
+export default { name, signature, description, register, action }
--- a/packages/cli/src/models/network/NetworkAppController.js
+++ b/packages/cli/src/models/network/NetworkAppController.js
@@ -93,15 +93,24 @@ export default class NetworkAppController extends NetworkBaseController {
     log.error(`Possible initialization method 'initialize' found in contract. Make sure you initialize your instance.`);
   }
 
-  async upgradeProxies(packageName, contractAlias, proxyAddress, initMethod, initArgs) {
-    // Fetch all proxies that match the filters provided
-    const proxies = this.networkFile.getProxies({ package: packageName, contract: contractAlias, address: proxyAddress})
-    if (_.isEmpty(proxies)) {
-      log.info('No proxies to update were found');
-      return [];
-    }
+  async setProxiesAdmin(packageName, contractAlias, proxyAddress, newAdmin) {
+    const proxies = this._fetchOwnedProxies(packageName, contractAlias, proxyAddress)
+    if (proxies.length === 0) return [];
+    await this.fetch();
 
-    // Load project
+    await allPromisesOrError(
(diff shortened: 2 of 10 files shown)


packages/cli/src/models/network/NetworkAppController.js
```diff
@@ -148,6 +164,28 @@ export default class NetworkAppController extends NetworkBaseController {
     log.error(`Possible migration method 'migrate' found in contract ${contractClass.contractName}. Remember running the migration after deploying it.`);
   }
 
+  _fetchOwnedProxies(packageName, contractAlias, proxyAddress) {
```
facuspagnuolo: is there a need to specify `owned`? is there a chance for a project to manage other proxies than the ones it owns?

packages/cli/src/models/network/NetworkAppController.js
```diff
@@ -148,6 +164,28 @@ export default class NetworkAppController extends NetworkBaseController {
...
+    const proxies = this.networkFile.getProxies({ 
+      package: packageName || (contractAlias ? this.packageFile.name : undefined),
+      contract: contractAlias, 
+      address: proxyAddress
+    })
+
+    if (_.isEmpty(proxies)) {
+      log.info('No owned contract instances that match were found');
```
facuspagnuolo: Shouldn't this be `No matching contract instances were found`? 
We could also log what we were looking for to be clearer

packages/cli/src/models/network/NetworkAppController.js
```diff
@@ -93,15 +93,25 @@ export default class NetworkAppController extends NetworkBaseController {
...
+    const proxies = this._fetchOwnedProxies(packageName, contractAlias, proxyAddress)
+    if (proxies.length === 0) return [];
+    await this.fetch();
 
-    // Load project
+    await allPromisesOrError(
+      _.map(proxies, async (proxy) => {
+        log.info(`Changing admin of proxy ${proxy.address} to ${newAdmin}`)
+        await this.project.changeProxyAdmin(proxy.address, newAdmin)
+        this.networkFile.updateProxy(proxy, proxy => ({ ... proxy, admin: newAdmin }))
```
facuspagnuolo: shouldn't we set this value also when creating a new instance?

packages/cli/src/models/network/NetworkAppController.js
facuspagnuolo: Note that you are repeating this validation in line 181, I think you can just delete this one since the `filter` above won't bother you in case no proxies were found at this point

packages/cli/src/models/network/NetworkAppController.js
```diff
@@ -93,15 +93,25 @@ export default class NetworkAppController extends NetworkBaseController {
...
-  async upgradeProxies(packageName, contractAlias, proxyAddress, initMethod, initArgs) {
-    // Fetch all proxies that match the filters provided
-    const proxies = this.networkFile.getProxies({ package: packageName, contract: contractAlias, address: proxyAddress})
-    if (_.isEmpty(proxies)) {
-      log.info('No proxies to update were found');
-      return [];
-    }
+  async setProxiesAdmin(packageName, contractAlias, proxyAddress, newAdmin) {
+    const proxies = this._fetchOwnedProxies(packageName, contractAlias, proxyAddress)
+    if (proxies.length === 0) return [];
```
facuspagnuolo: can we use `lodash` here to be consistent?

packages/cli/src/models/network/NetworkAppController.js
```diff
@@ -148,6 +164,28 @@ export default class NetworkAppController extends NetworkBaseController {
...
+    if (_.isEmpty(proxies)) {
+      log.info('No owned contract instances that match were found');
+      return [];
+    }
+
+    const ownedProxies = proxies.filter(proxy => !proxy.admin || proxy.admin === this.appAddress);
+
+    if (_.isEmpty(ownedProxies)) {
+      log.info('No matching contract instances are owned by this application');
+      return [];
```
facuspagnuolo: I think there is no need to specify different returns giving the one above will work as you are expecting in case no proxies were found

packages/cli/src/commands/set-admin.js
```diff
@@ -0,0 +1,37 @@
+'use strict';
+
+import setAdmin from '../scripts/set-admin'
+import runWithTruffle from '../utils/runWithTruffle'
+import { fromContractFullName } from '../utils/naming'
+import _ from 'lodash'
+
+const name = 'set-admin'
+const signature = `${name} [alias-or-address] [new-admin-address]`
+const description = 'change upgradeability admin of a contract instance. Provide the [alias] or [package]/[alias] you added your contract with, or its [address]. Note that if you transfer to an incorrect address, you may irreversibly lose control over upgrading your contract.'
```
facuspagnuolo: I would make clear that in case of an alias we are going to set the admin of all the instances for that alias

packages/cli/src/models/network/NetworkAppController.js
spalladino: After changing ownership of a proxy, the project will still keep track of it, and they will still be listed in the network file. It doesn't manage them, but it does have a reference to them.

packages/cli/src/models/network/NetworkAppController.js
spalladino: Validation in line 181 is about `ownedProxies`, not `proxies`

packages/cli/src/models/network/NetworkAppController.js
spalladino: We can safely assume that by default the admin of a new instance is the app

PR #49, opened 2018-08-28
Add CLI command to remove dependencies

jcarpanelli:
Implemented a new cli command for unlinking a list of multiple dependencies. 

_Example usage:_

```console
foo@bar:~$ zos unlink openzeppelin-zos gnosis-safe
```

### DIFF
--- a/packages/cli/src/commands/unlink.js
+++ b/packages/cli/src/commands/unlink.js
@@ -1,0 +1,22 @@
+'use strict'
+
+import unlink from '../scripts/unlink'
+import push from './push'
+
+const name = 'unlink'
+const signature = `${name} [libraries...]`
+const description = 'unlinks libraries from the project. Provide a list of whitespace-separated library names'
+
+const register = program => program
+  .command(signature, { noHelp: true })
+  .usage('[dependencyName1 ... dependencyNameN]')
+  .description(description)
+  .withPushOptions()
+  .action(action)
+
+async function action(libNames, options) {
+  await unlink({ libNames })
+  await push.tryAction(options)
+}
+
+export default { name, signature, description, register, action }
--- a/packages/cli/test/scripts/unlink.test.js
+++ b/packages/cli/test/scripts/unlink.test.js
@@ -9,37 +9,49 @@ contract('unlink script', function() {
     this.packageFile = new ZosPackageFile('test/mocks/packages/package-with-multiple-stdlibs.zos.json')
   })
 
-  it('throws an error if no libs are provided', async function () {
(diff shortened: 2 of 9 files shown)


packages/cli/src/commands/unlink.js
```diff
@@ -0,0 +1,22 @@
+'use strict'
+
+import unlink from '../scripts/unlink'
+import push from './push'
+
+const name = 'unlink'
+const signature = `${name} [libraries...]`
+const description = 'unlinks libraries from the project. Provide a list of whitespace-separated library names'
```
facuspagnuolo: maybe you have already thought about it, and it's not due to this PR but I noticed we are still using many terminologies for the same idea: stdlib, dependency, and library.. we should stick to only one or at least be clear with the difference between them
cc @spalladino

packages/cli/test/scripts/unlink.test.js
```diff
@@ -0,0 +1,45 @@
...
+
+  it('throws an error if project is a library', async function () {
+    this.packageFile.lib = true
+    const libNames = ['mock-stdlib', 'mock-stdlib-2']
+    await unlink({ libNames, packageFile: this.packageFile })
+      .should.be.rejectedWith('Libraries do not use stdlibs.')
+  })
+
+  it('throws an error if project library does not exist', async function () {
+    const libName = 'bulbasaur-lib2'
```
facuspagnuolo: 😂

packages/cli/test/scripts/unlink.test.js
```diff
@@ -0,0 +1,45 @@
...
+    await unlink({ libNames: [libName], packageFile: this.packageFile })
+    dependencies.should.not.have.property(libName)
+  })
+
+  it('unlinks multiple dependencies', async function () {
+    const { dependencies } = this.packageFile
+    const libNames = ['mock-stdlib', 'mock-stdlib-2']
+    await unlink({ libNames, packageFile: this.packageFile })
+    dependencies.should.not.have.property(...libNames)
+  })
```
facuspagnuolo: I would add a test to ensure that the dependencies you are not unlinking are still linked

packages/cli/src/commands/unlink.js
jcarpanelli: Agree. It's a bit annoying and at this point the terminology has become a bit confusing. 
+1 If you want to discuss which one we should pick and why!

facuspagnuolo:
There are some conflicts @jcarpanelli, feel free to fix them and merge!

PR #50, opened 2018-08-28
Extend link CLI command to support multiple dependencies

jcarpanelli:
Example usage:
```console
foo@bar:~$ zos link openzeppelin-zos gnosis-safe
```
or
```console
foo@bar:~$ zos init MyUpgradeableProjectWithDependencies 0.0.1 --link openzeppelin-zos,gnosis-safe
```

### DIFF
--- a/packages/cli/src/commands/init.js
+++ b/packages/cli/src/commands/init.js
@@ -20,13 +20,14 @@ const register = program => program
   .action(action)
 
 async function action(name, version, options) {
-  const { force } = options
+  const { force, link, install: installLibs } = options
+
   if (options.lib) {
-    if (options.link) throw Error('Cannot set a stdlib in a library project')
+    if (link) throw Error('Cannot set a stdlib in a library project')
     await initLib({ name, version, force })
   } else {
-    const { link: libNameVersion, install: installLib } = options
-    await init({ name, version, libNameVersion, installLib, force })
+    const libs = link ? link.split(',') : []
+    await init({ name, version, libs, installLibs, force })
   }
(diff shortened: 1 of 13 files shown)


packages/cli/src/commands/init.js
```diff
@@ -20,13 +20,14 @@ const register = program => program
...
+  const { force, link, install: installLibs } = options
+
   if (options.lib) {
-    if (options.link) throw Error('Cannot set a stdlib in a library project')
+    if (link) throw Error('Cannot set a stdlib in a library project')
     await initLib({ name, version, force })
   } else {
-    const { link: libNameVersion, install: installLib } = options
-    await init({ name, version, libNameVersion, installLib, force })
+    const libs = link ? link.split(',') : []
```
facuspagnuolo: we could use `utils/input#parseArgs` I think

packages/cli/test/scripts/create.test.js
```diff
@@ -7,7 +7,7 @@ import { Contracts, Logger } from 'zos-lib';
 import add from '../../src/scripts/add.js';
 import push from '../../src/scripts/push.js';
 import createProxy from '../../src/scripts/create.js';
-import linkLib from '../../src/scripts/link.js';
+import linkLibs from '../../src/scripts/link.js';
```
facuspagnuolo: I would stick to just `link`

jcarpanelli: Agree. Let's leave this up until we do the stdlib naming refactor.

Issue #42, opened 2018-08-24
Design lightweight functions for creating upgradeable instances


spalladino:
Today, using ZeppelinOS involves spinning up an App with Factory, with a Package, with an ImplementationProvider where the implementations are registered, before a proxy can be created. Some users have expressed concern as to locking-in to this structure, and are interested only in using the proxies themselves.

Design a set of functions to just create an upgradeable instance of a contract, without requiring setting up a ZeppelinOS app. Consider integrating these functions with existing frameworks, such as web3 or truffle-contracts.

Derived from https://github.com/zeppelinos/zos/issues/10

facuspagnuolo:
Closing this issue since https://github.com/zeppelinos/zos/pull/59 was already merged

PR #59, opened 2018-08-30
Design document for lightweight upgradeability functions

spalladino:
Design for #42


spalladino:
We discussed offline to add a section on implementing something closer to the current API, so migrating to a full-implementation is easier.

facuspagnuolo:
@spalladino do we want to maintain this kind of documents in the repo? we will work on further documentation after we develop the feature anyway, we can merge it and drop it later when the documentation is done if you think it is helpful for now

spalladino:
I'd merge it just to keep it handy, as you say, we can drop it eventually.

PR #76, opened 2018-09-06
Integration tests

### LINKED ISSUE #22: Setup simple integration tests
We are detecting issues that only arise with geth, parity, infura, or certain web3 providers. These issues are uncaught by our current unit tests. We need a set of tests that run on different infrastructures, installing the CLI and using it directly from the command line. These tests probably don't need to run on every commit, but either every X days or on every tag.

spalladino:
Setup integration tests with a set of contracts, running add, push, link, create, and update operations. Tests are run against a local geth dev node, both directly and via a truffle hd wallet provider.

Also introduces travis build stages to split package tests correctly. An initial job runs lib tests, then CLI tests (on Linux and OSX) are run in parallel with tests from examples, and only then the integration tests are run.

Fixes #22


.travis.yml
```diff
@@ -1,25 +1,94 @@
...
+      script: npx lerna run test --concurrency=1 --stream=true --scope=zos-lib
+    
+    # Test stage for CLI on Linux
+    - stage: tests
+      name: CLI tests on Linux
+      os: linux
+      install: npx lerna bootstrap --loglevel=debug
+      script: npx lerna run test --concurrency=1 --stream=true --scope=zos
+    
+    # Test stage for CLI on OSC
```
facuspagnuolo: `OSX`?

integration/cli-app/.gitignore
```diff
@@ -0,0 +1,8 @@
+node_modules
+.zos.session
+zos.*
+workdir/build/*
+workdir/node_modules
+workdir/zos*
+workdir/package*
+workdir/contracts/*
```
facuspagnuolo: What are these directories?

integration/cli-app/files/Samples.sol
```diff
@@ -0,0 +1,19 @@
...
+  function say() pure returns(string) {
+    return "Bar";
+  }
+}
+
+contract Baz {
+  function say() pure returns(string) {
+    return "Baz";
+  }
+}
```
facuspagnuolo: What about renaming to `integration` to `test`? IMO it will be clearer about its content, do you agree?

integration/cli-app/files/package.json.v2
```diff
@@ -0,0 +1,16 @@
...
+  "license": "MIT",
+  "dependencies": {
+    "truffle": "^4.1.14",
+    "zos": "1.4.1",
+    "zos-lib": "1.4.1",
+    "openzeppelin-zos": "1.9.4",
+    "bignumber.js": "^7.2.0",
+    "truffle-hdwallet-provider": "0.0.6"
+  }
+}
```
facuspagnuolo: How does the `.v2` works in the filename?

integration/cli-app/test/integration.test.js
```diff
@@ -0,0 +1,107 @@
...
+
+  it('upgrades a dependency', function () {
+    copy('WithTokenV2.sol', 'contracts/WithToken.sol')
+    run(`npx truffle compile`)
+    run(`npx zos push --deploy-stdlib --network ${network} --from ${this.from} --skip-compile`)
+    run(`npx zos update MintableERC721Token --network ${network} --from ${this.from}`)
+    run(`npx zos update WithToken --network ${network} --from ${this.from}`)
+    truffleExec(`run.js WithToken 0 isERC165 --network ${network}`).should.eq('true')
+  })
+});
```
facuspagnuolo: looks really good :)

integration/cli-app/test/integration.test.js
```diff
@@ -0,0 +1,107 @@
...
+
+function skipAfter() { }
```
facuspagnuolo: is this function declaration necessary here?

integration/cli-app/.gitignore
spalladino: `workdir` is the working directory where the test project is executed, so it is mostly gitignored

integration/cli-app/files/package.json.v2
spalladino: It's just for having two different versions that are copied during the test suite. Also, by not naming the file `package.json`, we are preventing lerna from picking it up as another project in the workspace.

spalladino:
@facuspagnuolo just updated this PR with the changes requested, especially renaming integration to tests (agree in that it's a better name). Please review again!

PR #78, opened 2018-09-06
Add clarification on adding storage vars to base contracts

spalladino:
Per suggestion of @frangio

### DIFF
--- a/packages/docs/docs/docs/writing_contracts.md
+++ b/packages/docs/docs/docs/writing_contracts.md
@@ -249,4 +249,25 @@ Then modifying `MyContract` by swapping the order in which the base contracts ar
 contract MyContract is B, A { }
 ```
 
-Violating any of these storage layout restrictions will cause the upgraded version of the contract to have its storage values mixed up, and can lead to critical errors in your application.
+You also cannot add new variables to base contracts, if the child has any variables of its own. Given the following scenario:
+```js
+contract Base {
+  uint256 base1;
+}
+
+contract Child {
+  uint256 child;
+}
+```
+
+And if `Base` is modified to add an extra variable:
+```js
+contract Base {
+  uint256 base1;
+  uint256 base2;
+}
+```
+
+Then `base2` whould be assigned the same slot as `child`. A workaround for this is to declare unused variables on base contracts that you may want to extend in the future, as a means of "reserving" those slots. Note that this trick does not involve increased gas usage.
+
+> Violating any of these storage layout restrictions will cause the upgraded version of the contract to have its storage values mixed up, and can lead to critical errors in your application.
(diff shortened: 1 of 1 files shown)


packages/docs/docs/docs/writing_contracts.md
```diff
@@ -249,4 +249,25 @@ Then modifying `MyContract` by swapping the order in which the base contracts ar
...
 
-Violating any of these storage layout restrictions will cause the upgraded version of the contract to have its storage values mixed up, and can lead to critical errors in your application.
\ No newline at end of file
+You also cannot add new variables to base contracts, if the child has any variables of its own. Given the following scenario:
+```js
+contract Base {
+  uint256 base1;
+}
+
+contract Child {
```
frangio: Missing the inheritance from `Base`.

packages/docs/docs/docs/writing_contracts.md
```diff
@@ -249,4 +249,25 @@ Then modifying `MyContract` by swapping the order in which the base contracts ar
...
+```js
+contract Base {
+  uint256 base1;
+  uint256 base2;
+}
+```
+
+Then `base2` whould be assigned the same slot as `child`. A workaround for this is to declare unused variables on base contracts that you may want to extend in the future, as a means of "reserving" those slots. Note that this trick does not involve increased gas usage.
+
+> Violating any of these storage layout restrictions will cause the upgraded version of the contract to have its storage values mixed up, and can lead to critical errors in your application.
```
frangio: Is the `>` on purpose?

packages/docs/docs/docs/writing_contracts.md
```diff
@@ -249,4 +249,25 @@ Then modifying `MyContract` by swapping the order in which the base contracts ar
...
+contract Base {
+  uint256 base1;
+}
+
+contract Child {
+  uint256 child;
+}
+```
+
+And if `Base` is modified to add an extra variable:
```
frangio: I find the "and" at the beginning of the sentence a bit confusing when the section is read as a whole.

packages/docs/docs/docs/writing_contracts.md
```diff
@@ -249,4 +249,25 @@ Then modifying `MyContract` by swapping the order in which the base contracts ar
...
+```js
+contract Base {
+  uint256 base1;
+  uint256 base2;
+}
+```
+
+Then `base2` whould be assigned the same slot as `child`. A workaround for this is to declare unused variables on base contracts that you may want to extend in the future, as a means of "reserving" those slots. Note that this trick does not involve increased gas usage.
```
frangio: I think the first sentence may be confusing. I would change it to read "... the slot `child` where was stored in the previous version" or something like that.

packages/docs/docs/docs/writing_contracts.md
spalladino: Yep, it's displayed as a highlighted note in our site

spalladino:
@frangio thanks for the review! Just updated the doc.

PR #82, opened 2018-09-07
TestApp fix to work with multiple deps models

jcarpanelli:
I've renamed the `TestApp` model to `TestHelper`. It now works fine with app projects that have multiple stdlibs, as well as lib projects.


packages/cli/src/models/ProjectApp.js
```diff
@@ -8,7 +8,8 @@ import ControllerFor from '../models/network/ControllerFor';
  */
 export default async function testApp(txParams = {}, networkFile = undefined) {
```
spalladino: I'd rename it to TestProject instead of ProjectApp, and consider supporting libs as well as apps

PR #83, opened 2018-09-08
Lightweight project implementation

### LINKED ISSUE #70: Implement Project subclass for lightweight upgrades
Following https://github.com/zeppelinos/zos/pull/59, implement a subclass of Project in `zos/lib` that creates upgradeable contracts _without_ relying on App, Package, Provider, and Factory. The interface should cover at least:
```js
async createProxy(contractClass, { packageName, contractName, initMethod, initArgs })
async upgradeProxy(proxyAddress, contractClass, { packageName, contractName, initMethod, initArgs })
async changeProxyAdmin(proxyAddress, newAdmin) 
```

Sample usage:
```js
const Project = require('zos-lib').Project
const project = new Project()
const MyContract = artifacts.require("MyContract")
const myInstance = project.createProxy(MyContract, { initMethod: 'initialize', initArgs: [42] })
```

spalladino:
Provides an implementation named SimpleProject for a lib/Project that can manage proxies without setting up a Package or App contracts at all. Note that SimpleProject requires to be provided with *two* sender accounts, one for managing proxies and one for initializing them, since the TransparentProxy pattern makes it impossible for the admin of the proxy to invoke `initialize` (or any other function of the underlying instance).

Fixes #70


packages/lib/src/project/AppProject.js
```diff
@@ -1,8 +1,9 @@
-import Project from "./Project";
+import BasePackageProject from "./BasePackageProject";
 import VersionedApp from "../app/VersionedApp";
 import Package from "../package/Package";
+import _ from 'lodash';
 
-export default class AppProject extends Project {
+export default class AppProject extends BasePackageProject {
```
facuspagnuolo: I think we can use just `PackageProject`

packages/lib/src/project/AppProject.js
```diff
@@ -60,31 +61,29 @@ export default class AppProject extends Project {
...
     if (!packageName) packageName = this.name
     return this.app.createContract(contractClass, packageName, contractName, initMethod, initArgs)
   }
 
-  // TODO: Testme
-  async createProxy(contractClass, { packageName, contractName, initMethod, initArgs }) {
+  async createProxy(contractClass, { packageName, contractName, initMethod, initArgs } = {}) {
     if (!contractName) contractName = contractClass.contractName
     if (!packageName) packageName = this.name
+    if (!_.isEmpty(initArgs) && !initMethod) initMethod = 'initialize'
```
facuspagnuolo: I think this check was already performed at an upper level, isn't it? If so, do you think it still make sense to have it duplicated?

packages/lib/src/project/SimpleProject.js
```diff
@@ -0,0 +1,62 @@
...
+  async _deployImplementation(contractClass) {
+    log.info(`Deploying logic contract for ${contractClass.contractName}`);
+    const implementation = await deploy(contractClass, [], this.txParamsAdmin);
+    return implementation;
+  }
+
+  async _tryInitializeProxy(proxy, contractClass, initMethodName, initArgs) {
+    if (!initMethodName) return;
+    if (this.txParamsInitializer.from === this.txParamsAdmin.from) {
+      throw Error(`Cannot initialize the proxy from the same address as its admin address. Make sure you use a different 'from' account for the initialization transaction params.`)
```
facuspagnuolo: I see why we need to do this but it will be completely weird for the users and I'm sure they won't understand it at once. If we move the initialization method call to the constructor, this shouldn't be a problem anymore, am I right?

packages/lib/src/project/SimpleProject.js
```diff
@@ -0,0 +1,62 @@
...
+    log.info(`Instance at ${proxyAddress} upgraded`)
+    return new contractClass(proxyAddress);
+  }
+
+  async changeProxyAdmin(proxyAddress, newAdmin) {
+    const proxy = Proxy.at(proxyAddress, this.txParamsAdmin)
+    await proxy.changeAdmin(newAdmin)
+    log.info(`Proxy admin changed to ${newAdmin}`)
+    return proxy
+  }
```
facuspagnuolo: In general, I'd love to expose an interface that abstracts you from the idea of having a proxy under the hood, even more if are planning to deal with non-upgradeable instances. Honestly, I don't have better options for naming now, we can think more about them later together

packages/lib/src/index.js
```diff
@@ -18,13 +17,14 @@ const assertions = helpers.assertions
 const assertRevert = helpers.assertRevert
 
 // model objects
+import Proxy from './proxy/Proxy'
```
facuspagnuolo: I think we don't need to import this one right?

packages/lib/src/project/AppProject.js
spalladino: Hmm I actually prefer to make it clear which are base classes, and which are leaves that can be actually instantiated.

packages/lib/src/project/AppProject.js
spalladino: It may be done at an upper level on the CLI, but since we may want to start promoting usage of the `lib` directly, we need to have it here as well.

spalladino: Actually I scanned through the NetworkAppController, scripts/create, and commands/create, and didn't found it. It could be that I already removed it when making this change.

packages/lib/src/project/SimpleProject.js
spalladino: Indeed, that's why I was interested in pushing that change; we should work on it ASAP. Still, keep in mind that the user would still need a 2nd account to _interact_ with the proxy after creating it, but at least the Project can work with a single account.

packages/lib/src/index.js
spalladino: It's exported in L40 of this file

spalladino:
@facuspagnuolo could you please re-review? I've only rebased on top of `next`

PR #84, opened 2018-09-10
Tests/add dependency tests

jcarpanelli:
- Removed Stdlib deprecated tests
- Added tests for new Dependency model


packages/cli/test/models/Dependency.test.js
```diff
@@ -0,0 +1,129 @@
...
+        const nameAndVersion = 'mock-stdlib@1.1.0'
+        const npmParams = { cwd: process.cwd(), save: true }
+        await this.dependency.install()
+        npmInstallStub.should.have.been.calledWithExactly([nameAndVersion], npmParams)
+        sinon.restore()
+      })
+    })
+  })
+
+})
```
facuspagnuolo: that neatness ❤️ love it

jcarpanelli:
@spalladino  would you please take a look to this so we can merge it?

PR #92, opened 2018-09-13
Validate storage layout

### LINKED ISSUE #37: Validate that an upgraded contract keeps same storage layout as the original one
(follows from https://github.com/zeppelinos/zos-lib/issues/63)

## The problem

For an upgrade to be safe, we need to ensure that the new contract shares the same storage layout as the previous version, and only appends new variables, without modifying the existing ones.

For instance, given:
```solidity
contract V1 {
  uint a;
  uint b;
}
```

The following upgrade is valid, since it's only adding a new variable at the end of the storage:
```solidity
contract V2 {
  uint a;
  uint b;
  uint c;
}
```

The following ones, on the other hand, are all invalid:
```solidity
contract V2 {
  uint a;
  uint c; // New variable is inserted in the slot previously assigned to `b`
  uint b;
}

contract Base {
  uint base; // New `base` variable will be assigned the first slot
}
contract V2 is Base { 
  uint a;
  uint b;
}

contract V2 {
  uint a;
  string b; // Type of a variable is changed
}
```

Other potential breaking changes include modifying the fields of a struct, or changing inheritance order. Anything that alters the storage layout can potentially introduce bugs, and needs to be detected. To detect this, we need to figure out the storage layout, store it, and compare it when a new version is set.

## Infering storage layout

The first step is to get the storage layout given a contract. The safest option to do this is to keep building on https://github.com/ethereum/solidity/pull/4017, which is an enhancement to the Solidity compiler to provide this information directly. This requires considerable effort, and we'd need to wait until the new version of the compiler is released before we can make use of the feature.

A more lightweight alternative would be to extract this information from the AST. Assuming that the order of variables in storage is the same as the order in which they are declared (this is a safe assumption to make), we can walk the AST looking for nodes that correspond to state variables (`"nodeType": "VariableDeclaration", "stateVariable": true`), and store their order and type. Note that we'd also need to walk the `linearizedBaseContracts` (in the order specified) and also retrieve their state variables, to build a complete picture of the storage.

Ideally, we could output this information in a format as similar as possible as the one described [here](https://github.com/ethereum/solidity/pull/4017#issuecomment-410331891), so the transition to the information provided by `solc` is as smooth as possible.

## Comparing storage layouts

Next step is to build a function that, given two storage layouts as extracted in the previous step, returns whether one is a valid extension of the other. It needs to check that variables types and names haven't changed, taking into account complex types (mappings, structs, etc) as well. 

The output needs to be human-readable, so we can provide the user with information on where the problem is. Make sure the error is understandable when caused by a change in an ancestor: if a base contract changes its storage layout, it's something quite hidden to the user, so we need to be extra clear in those scenarios.

## Recording storage info for a contract and checking

In order to compare a version with a new one, we need to store the storage layout information somewhere, and when the user proposes a new version, we perform the check. Ideally, we could to this when the user registers the new changes, but we need to have a [git-like stage](https://github.com/zeppelinos/zos-cli/issues/119) in place for that.

Instead, we can do that when the user runs `zos push` on a new version. On the first `zos push`, we store the storage layout info in the `zos.network.json` file (along with the bytecode hash). On subsequent ones, we recalculate the storage layout for the new contract, and compare it with the stored one. If it's ok, we allow the push and overwrite the storage layout. If not, we show the errors to the user. We'd also need to include an `--ignore-warns-on-storage-layout` (better name needed) flag so the user can `push` anyway, to work around false positives.

spalladino:
When running zos-push, validate that the storage layout is compatible with the previous version of the contract, and abort push otherwise. All warnings are displayed to the user, eg:

```
$ zos push
Validating contract AnotherImplV1 before push
Variable 'uint256 value22' in contract ImplV1 was renamed to foo in contracts/mocks/ImplV1.sol:1.
Note that the new variable foo will have the value of value22 after upgrading. If this is not the desired behavior, add a new variable foo at the end of your contract instead.
New variable 'uint256 bar' was added in contract ImplV1 in contracts/mocks/ImplV1.sol:1
This pushes all variables after bar to a higher position in storage, causing the updated contract to read incorrect initial values. Only add new variables at the end of your contract to prevent this issue.
```

Storage layout is extracted from the contract's AST by the Storage class in lib, and is stored along with other contract data (such as bytecode hash) in the zos.network.json file. It is then retrieved for comparison on the next zos push. The format is a tuple `(storage, types)`, similar to the format discussed in ethereum/solidity#4017, where `storage` contains the array of variables, with references to the `types` dictionary with detailed info on each type.

Comparison is handled by the Layout class in lib, and uses a modified levenshtein algorithm for calculating the minimum change between both storages. Changes are identified as additions (either at the end or mid-contract, with additions at the end computed at zero cost for the algorithm), deletions (end or mid-contract), or substitutions (renames, typechanges, or both).

Fixes #37 

Pending:
- [x] Handle changes in complex types (structs)
- [ ] Adding more tests for Layout
- [x] Improve and refactor CLI output
- [ ] Test CLI output

### DIFF
--- a/packages/cli/src/models/files/ZosNetworkFile.js
+++ b/packages/cli/src/models/files/ZosNetworkFile.js
@@ -192,12 +192,13 @@ export default class ZosNetworkFile {
     this.setDependency(name, fn(this.getDependency(name)));
   }
 
-  addContract(alias, instance) {
+  addContract(alias, instance, extraData = {}) {
     this.setContract(alias, {
       address: instance.address,
       constructorCode: constructorCode(instance),
       bodyBytecodeHash: bytecodeDigest(bodyCode(instance)),
       bytecodeHash: bytecodeDigest(instance.constructor.bytecode),
+      ... extraData
     })
   }
 
--- a/packages/cli/src/models/files/ZosNetworkFile.js
+++ b/packages/cli/src/models/files/ZosNetworkFile.js
@@ -192,13 +192,14 @@ export default class ZosNetworkFile {
     this.setDependency(name, fn(this.getDependency(name)));
   }
 
-  addContract(alias, instance, extraData = {}) {
+  addContract(alias, instance, { types, storage }) {
     this.setContract(alias, {
       address: instance.address,
       constructorCode: constructorCode(instance),
       bodyBytecodeHash: bytecodeDigest(bodyCode(instance)),
       bytecodeHash: bytecodeDigest(instance.constructor.bytecode),
-      ... extraData
+      types,
+      storage
     })
   }
 
--- a/packages/cli/src/models/network/NetworkBaseController.js
+++ b/packages/cli/src/models/network/NetworkBaseController.js
@@ -6,6 +6,7 @@ import Verifier from '../Verifier'
 import { flattenSourceCode } from '../../utils/contracts'
 import { allPromisesOrError } from '../../utils/async';
 import { getStorageLayout, getBuildArtifacts, compareStorageLayouts } from 'zos-lib/src';
+import { logStorageLayoutDiffs } from '../../interface/Validations';
 
 const log = new Logger('NetworkController');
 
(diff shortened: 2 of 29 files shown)


facuspagnuolo:
Let's please create some issues for the pending items

packages/cli/src/models/files/ZosNetworkFile.js
```diff
@@ -192,12 +192,13 @@ export default class ZosNetworkFile {
     this.setDependency(name, fn(this.getDependency(name)));
   }
 
-  addContract(alias, instance) {
+  addContract(alias, instance, extraData = {}) {
```
facuspagnuolo: I like the idea of using object arguments when those are kind of irrelevant to the rest of the parameters and if they are some type of configuration for the function being called. In this case, note that that `extraData` has the same importance as the rest of the parameters, we are deconstructing it as is, and it is impossible for the reader to know what data is actually being handled here. It forces the reader to look up for calls to this function to understand how it should be called. It is definitely not a blocker, but it is one of the things we should decide and standardize for the whole codebase.

packages/cli/src/models/network/NetworkBaseController.js
```diff
@@ -122,6 +140,98 @@ export default class NetworkBaseController {
...
+    const [begin] = srcFragment.split(':', 1);
+    return sourceCode.substr(0, begin).split('\n').length
+  }
+
+  _tryGetSourceFragment(sourceCode, src) {
+    if (!src || !sourceCode) return null;
+    const [begin, count] = src.split(':');
+    return sourceCode.substr(begin, count);
+  }
+
```
facuspagnuolo: I'd move all this logic outside the controller

packages/lib/src/validations/Layout.js
```diff
@@ -0,0 +1,111 @@
+import _ from 'lodash';
+import util from 'util';
+
+const SUBSTITUTION_COST = 3,
+      INSERTION_COST = 2,
+      DELETION_COST = 2;
+
+export function compareStorageLayouts(original, updated) {
+  // TODO: Check cases with empty storage (both for original and updated)
```
facuspagnuolo: let's move this `TODO` to an issue please

packages/lib/src/validations/Layout.js
```diff
@@ -0,0 +1,111 @@
...
+    return 'rename'
+  } else if (nameMatches) {
+    return 'typechange'
+  } else {
+    return 'replace'
+  }
+}
+
+// Adapted from https://gist.github.com/andrei-m/982927 by Andrei Mackenzie
+function levenshtein(originalStorage, updatedStorage, areEqualFn) {
```
facuspagnuolo: 👏

packages/lib/src/validations/Layout.js
```diff
@@ -0,0 +1,111 @@
...
+  const distanceMatrix = levenshtein(original.storage, updated.storage, areEqual)
+  const operations = walk(distanceMatrix, original.storage, updated.storage, areMatch)
+  return operations.filter(op => op.action !== 'equal')
+}
+
+function storageEntryMatches(originalVar, updatedVar, originalTypes, updatedTypes) {
+  const originalType = originalTypes[originalVar.type],
+        updatedType = updatedTypes[updatedVar.type];
+
+  // TODO: Compare complex types (structs and enums)
```
facuspagnuolo: same here, let's create another issue for that

packages/lib/src/validations/Storage.js
```diff
@@ -0,0 +1,221 @@
...
+      members: referencedNode.members.map(m => m.name)
+    }
+  }
+
+  getValueTypeInfo(node) {
+    return (node.nodeType === 'Mapping')
+      ? this.getValueTypeInfo(node.valueType)
+      : this.getTypeInfo(node)
+  }  
+}
```
facuspagnuolo: 😍

packages/lib/src/validations/Storage.js
```diff
@@ -0,0 +1,221 @@
...
+  }
+
+  getFunctionTypeInfo() {
+    // Process a reference to a function disregarding types, since we only care how much space it takes
+    return { 
+      id: 't_function', 
+      kind: 'elementary',
+      label: 'function' 
+    } 
+  }
```
facuspagnuolo: could we move this one and the one above to constants?

packages/cli/src/models/network/NetworkBaseController.js
spalladino: Absolutely!

spalladino:
@facuspagnuolo I implemented the suggested changes, and added issues #112, #113, and #114 to cover for the missing pieces (all in Backlog for now). The only thing I'm wondering is whether we may want to add a feature toggle to disable these warnings until those issues are finished, though I think we can do that later if we find out that we can't get them ready by the next release.

PR #93, opened 2018-09-13
Update simple-example to use new simple project

spalladino:
Instead of manually deploying logic contracts, proxies, and initializing them, simple-example now uses SimpleProject. Also adds a test to ensure it keeps working with the latest version of lib (side note: it wasn't).


spalladino:
@facuspagnuolo it seems I hadn't properly configured the "do not dismiss stale reviews" on the project (I was sure I had done it, but who knows). Could you re-approve if you don't mind?

PR #105, opened 2018-09-19
Remove swarm hash from bytecode before hashing it

jcarpanelli:
Fixes https://github.com/zeppelinos/zos/issues/38.
@spalladino should we add an option parameter to the `bytecodeDigest` function, e.g., `{ removeSwarmHash: true/false }` to decide whether to remove it or not?

### DIFF
--- a/packages/cli/src/utils/contracts.js
+++ b/packages/cli/src/utils/contracts.js
@@ -12,7 +12,7 @@ export function constructorCode(instance) {
 }
 
 export function bytecodeDigest(rawBytecode) {
-  const bytecode = removeSwarmHash(rawBytecode.replace(/^0x/, ''))
+  const bytecode = tryRemoveSwarmHash(rawBytecode.replace(/^0x/, ''))
   const buffer = Buffer.from(bytecode, 'hex')
   const hash = crypto.createHash('sha256')
   return hash.update(buffer).digest('hex')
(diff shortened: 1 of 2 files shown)


packages/cli/src/utils/contracts.js
```diff
@@ -26,3 +21,14 @@ export function bytecodeDigest(rawBytecode) {
...
+
+function splitCode(instance) {
+  const bytecode = instance.constructor.bytecode.replace(/^0x/, '')
+  const body = instance.constructor.deployedBytecode.replace(/^0x/, '')
+  const constructor = bytecode.substr(0, bytecode.indexOf(body))
+  return { constructor, body }
+}
+
+function removeSwarmHash(bytecode) {
+  return bytecode.replace(/a165627a7a72305820.*0029$/, '')
```
facuspagnuolo: would you mind adding a comment to explain this chunk of bytecode?

packages/cli/src/utils/contracts.js
```diff
@@ -26,3 +21,17 @@ export function bytecodeDigest(rawBytecode) {
...
+  const body = instance.constructor.deployedBytecode.replace(/^0x/, '')
+  const constructor = bytecode.substr(0, bytecode.indexOf(body))
+  return { constructor, body }
+}
+
+// Removes the last 43 bytes of the bytecode, i.e., the swarm hash that the solidity compiler appends and that
+// respects the following structure: 0xa1 0x65 'b' 'z' 'z' 'r' '0' 0x58 0x20 <32 bytes swarm hash> 0x00 0x29
+// (see https://solidity.readthedocs.io/en/v0.4.24/metadata.html#encoding-of-the-metadata-hash-in-the-bytecode)
+function removeSwarmHash(bytecode) {
+  return bytecode.replace(/a165627a7a72305820.*0029$/, '')
```
spalladino: Let's assert that the swarm hash is indeed 32 bytes. We can replace `.*` in the regex for `[a-fA-F0-9]{64}`.

spalladino: Also, it'd be interesting to fail if the replace operation does not match anything to replace, in case we stumble upon a contract with a different swarm hash format.

jcarpanelli: @spalladino what do you mean by "fail"?

jcarpanelli: already discussed it offline.

Issue #88, opened 2018-09-13
Simplify backing architecture of contracts for an upgradeable App


facuspagnuolo:
Currently we have a huge architecture of smart contracts in order to reflect the different versions of an application of smart contracts on-chain. The idea of this issue is to propose a new simpler model:

For instance, we are providing different flavors of `App`s (versioned and unversioned) when we are only providing support to work with the versioned one through the CLI. We can remove the unversioned one in favour of having an off-chain representation of the smart contracts of your application, which is very much aligned with the lightweight mode we are working on as well (https://github.com/zeppelinos/zos/issues/74).

This will basically trigger the idea of having a single `App` that could have many `Packages`, and then every `Package` can have many `ImplementationsDirectories` as we are doing now.

PR #106, opened 2018-09-19
Move proxy initialization to constructors

facuspagnuolo:
Fixes https://github.com/zeppelinos/zos/issues/100


packages/lib/contracts/application/BaseApp.sol
```diff
@@ -75,8 +73,9 @@ contract BaseApp is Ownable {
    * @return Address of the new proxy.
    */
   function create(string packageName, string contractName) public returns (AdminUpgradeabilityProxy) {
-    address implementation = getImplementation(packageName, contractName);
-    return factory.createProxy(this, implementation);
+    // TODO: should we merge this function with the createAndCall?
```
spalladino: +1

packages/lib/contracts/upgradeability/UpgradeabilityProxy.sol
```diff
@@ -26,10 +26,15 @@ contract UpgradeabilityProxy is Proxy {
...
+   * @param _data Data to send as msg.data in the low level call.
+   * It should include the signature and the parameters of the function to be called, as described in
+   * https://solidity.readthedocs.io/en/develop/abi-spec.html#function-selector-and-argument-encoding.
    */
-  constructor(address _implementation) public {
+  constructor(address _implementation, bytes _data) public payable {
     assert(IMPLEMENTATION_SLOT == keccak256("org.zeppelinos.proxy.implementation"));
-
+    if(_data.length > 0) {
+      require(_implementation.delegatecall(_data));
```
spalladino: This means we are settling with `initializer functions` and not `initializer contracts`. Should we consider instead accepting `address implementation, address initializer, bytes data`, and run `initializer.delegatecall(data)` if initializer is not zero? The App could initially use the same address for both, but leaves the door open for initializer contracts in the future. @frangio care to chime in here?

packages/lib/contracts/application/BaseApp.sol
```diff
@@ -11,17 +10,16 @@ import "openzeppelin-solidity/contracts/ownership/Ownable.sol";
...
 
   /**
    * @dev Constructor function
-   * @param _factory Proxy factory
    */
-  constructor(UpgradeabilityProxyFactory _factory) public {
-    require(address(_factory) != address(0), "Cannot set the proxy factory of an app to the zero address");
-    factory = _factory;
-  }
+  constructor() public {}
```
spalladino: I hesitated about entirely removing the ProxyFactory, since the original idea was that we could have a single ProxyFactory shared by all Apps, so each user would only need to deploy the App. However, since we are now aiming at making these contracts upgradeable, gas cost for the end user is exactly the same, since they only need to deploy a proxy to App, regardless of where the factory code for proxies is. So +1 to this change.

spalladino:
@facuspagnuolo I've just noted that there are some tests on SimpleProject failing. I started working on #88 on top of this PR without noticing that, and I'm already fixing them on my branch. We can either:

- Wait until I push those fixes, you manually copy them onto this branch, and then I rebase
- Close this PR, and then it goes in together with the fix for #88

packages/lib/contracts/upgradeability/UpgradeabilityProxy.sol
frangio: I'd like to leave the door open but I'm worried that it might be overkill for now. Would it be too disruptive to change that in a future version?

Note that if we do add the `address initializer` parameter, we should allow `_data.length == 0` because initializer contracts can run with no data.

facuspagnuolo: I'd like the idea of `initializer contracts` but in that case, I'd expect the CLI to solve the initializers functions to initializer contracts conversions and it might be an overkill for now. I think we can provide a new implementation of the proxy using `initializer contracts` with the required CLI improvements to deal with it in the future.

facuspagnuolo:
@spalladino I committed some new changes, please take another look. I also cherry-picked the commits you mentioned about the `SimpleProjet` to make sure the CI was passing and there was nothing else broken on my end. We can merge and rhen you rebase, or close and merge your PR including these changes whatever is best for you is fine for me.

packages/lib/contracts/upgradeability/UpgradeabilityProxy.sol
frangio: So the CLI change would be transparent and backwards compatible. But the `Proxy` and `App` changes would not be backwards compatible and would need a major bump.

Are we comfortable with that?

spalladino: The App change could be backwards compatible, as we could just add a new `createAndCall` method that receives the `initializer address` parameter as well as the others. The only breaking change would be around proxies, but since it only affects the ctor, I don't think it'd be breaking at all.

frangio: What do we mean by "breaking" exactly? If we do change the constructor it _could_ break deployment scripts that use the `Proxy` directly.

Issue #35, opened 2018-08-22
Preserve App, Package, and Directory addresses in aborted deployments


spalladino:
Given an initial `zos push` operation that deploys Factory, App, Package, and Directory contracts, we want to preserve them as they are being deployed, in case the deployment fails mid-run. Make sure the addresses of each contract is stored in the zos.network.json file as they are deployed, and that the push picks up from there on subsequent operations.

Related to https://github.com/zeppelinos/zos-cli/issues/319

PR #109, opened 2018-09-20
Preserve App, Package and Directory addresses

jcarpanelli:
Related to #35
A first approach to preserve addresses in `zos.network.json` if the project deployment fails at some point.

- [x] Implement an app, package and directory address tracker in lib and cli.
- [ ] Pick up already deployed contracts addresses on  future `push` runs.

### DIFF
--- a/packages/lib/src/utils/errors/DeployError.js
+++ b/packages/lib/src/utils/errors/DeployError.js
@@ -1,0 +1,17 @@
+'use strict'
+
+export class DeployError extends Error {
+  constructor(message, thepackage, directory) {
+    super(message)
+    this.package = thepackage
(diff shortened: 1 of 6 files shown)


packages/lib/src/utils/errors/DeployError.js
```diff
@@ -0,0 +1,17 @@
+'use strict'
+
+export class DeployError extends Error {
```
spalladino: How about creating a single DeployError that accepts an object, and assigns it to `this`? So we can have a generic DeployError that we use for every circumstance.

spalladino:
Looking great so far!

PR #110, opened 2018-09-20
Simplify zos-lib contracts

### LINKED ISSUE #88: Simplify backing architecture of contracts for an upgradeable App

spalladino:
- Remove VersionedApp/UnversionedApp distinction
- Remove FreezableImplementationDirectory/ImplementationDirectory

Fixes #88


packages/lib/contracts/application/App.sol
```diff
@@ -1,32 +1,92 @@
...
+  /**
+   * @dev Sets a package in a specific version as a dependency for this application. 
+   * Requires the version to be present in the package.
+   * @param packageName Name of the package to set or overwrite.
+   * @param package Address of the package to register.
+   * @param version Version of the package to use in this application.
+   */
+  function setPackage(string packageName, Package package, string version) public onlyOwner {
+    require(package.hasVersion(version), "The requested version must be registered in the given package");
+    providers[packageName] = ProviderInfo(package, version);
```
facuspagnuolo: should we check that there was no other package already set under that name?

facuspagnuolo: I'm thinking if we should force the user to unset the previous package in that case

packages/lib/test/contracts/application/App.test.js
```diff
@@ -0,0 +1,51 @@
...
+
+  describe('ownership', function () {
+    beforeEach("setting ownable", function () {
+      this.ownable = this.app
+    })
+    shouldBehaveLikeOwnable(appOwner, anotherAccount)
+  })
+
+  shouldManageProxies(accounts);
+  shouldManagePackages(accounts);
```
facuspagnuolo: 👍

packages/lib/contracts/application/App.sol
spalladino: I wouldn't, as this allows the user to use this method to also change the version of a package if they want to do so. And I'd reduce the number of txs to interact with the App as much as possible.

PR #111, opened 2018-09-21
Changes on design so it is inline with the website

marianorod:


packages/docs/docs/website/static/css/custom.css
```diff
@@ -1,6 +1,32 @@
...
+
 @media (max-width: 735px) {
+    .docMainWrapper .wrapper{ 
+    padding:0 20px;
+ 
+  }
+
+  .navigationSlider .slidingNav ul{ 
+    background-color: #1d2327 !important;
+      }
```
facuspagnuolo: I think it is not properly indented, at least my browser is showing me these lines weirdly indented. Would you mind taking a look at it?

Issue #108, opened 2018-09-19
Implement vouching tests


fiiiu:
Implement tests for the new Vouching (formerly Kernel) system.

Issue #103, opened 2018-09-18
Add TPL support to token contract


spalladino:
Add [TPL](https://tplprotocol.org/) support to the token contract from https://github.com/zeppelinos/kernel, which will be used within the Kernel in #101.

Issue #101, opened 2018-09-18
Implement Kernel contract for managing libraries stake


spalladino:
Create a Kernel contract that allows for registering new libraries, staking, and unstaking, using an ERC20 token. A library should have the following data, and could be implemented as a struct within the contract:

- The developer who runs the project (address)
- The developer pool balance in ZEP (uint256)
- The on-chain code (address of Package at major version) (discuss whether a new major spawns a new package, or we need to add a "major" attribute to Lib here, since a dev may be supporting two majors of his library)
- The source code (url + hash of source, updated always to the latest version, fires an event when updated to track changes)

The Kernel contract should support four main actions:

- Creating a new library, including an initial stake
- Updating the library info (define what info can be updated, for instance, the source code should be upgradeable, or the ownership should be transferrable)
- Staking tokens for that library (only the owner of the library)
- Removing staked tokens (only the owner)

The contract should be upgradeable. Consider reusing part of the work done in https://github.com/zeppelinos/kernel, and even developing this new version in that repo.

PR #115, opened 2018-09-21
Token and vouching mechanics

fiiiu:
This PR adds a new package implementing the `ZEP` token and the `Vouching` mechanics, closing #101 and #108.

Pending Work: 
- The contracts are not yet upgradeable, we are tackling this on a new branch on top of this one (#116).
- TPL is not yet integrated, so the token contract is some of a placeholder. Same as above regarding further work (#103).

We are opening this as is so as to get a first round of review on the vouching mechanics, which are independent of the above. 

Notes:
- Right now, an attacker can register a new dependency with a fresh name, but with an address already listed in the system. This does not open any obvious vulnerability, but might cause confusion with respect to who the owner of a dependency is. Ultimately, this stems from the fact that the indexing is done with dependency names instead of their addresses. 

(Opening on `master` since that's where we started from.. Cannot automatically merge on `next` :-/ )


packages/vouching/.gitattributes
```diff
@@ -0,0 +1 @@
+*.sol linguist-language=Solidity
```
spalladino: Should we move this to the root of the repo?

packages/vouching/contracts/Migrations.sol
```diff
@@ -0,0 +1,23 @@
+pragma solidity ^0.4.23;
+
+contract Migrations {
```
spalladino: Let's remove this contract altogether

packages/vouching/contracts/Vouching.sol
```diff
@@ -0,0 +1,101 @@
+pragma solidity ^0.4.24;
+
+
+import "./ZepToken.sol";
+import "openzeppelin-solidity/contracts/math/SafeMath.sol";
```
spalladino: Ok for now, but keep in mind that we'll need to use oz-zos and not oz-sol when tackling #116

packages/vouching/contracts/Vouching.sol
```diff
@@ -0,0 +1,101 @@
...
+
+contract Vouching {
+  using SafeMath for uint256;
+
+  struct Dependency {
```
spalladino: Not a fan of the name. `Dependency` makes sense in a context where it is a dependency of something else. I'd prefer `PackageEntry` or something like that.

packages/vouching/contracts/Vouching.sol
```diff
@@ -0,0 +1,101 @@
...
+    address owner;
+    address dependencyAddress;
+    uint256 stake;
+  }
+
+  mapping (string => Dependency) private _registry;
+  uint256 private _minimumStake;
+  ZepToken private _token;
```
spalladino: Let's use a generic ERC20 here

packages/vouching/contracts/Vouching.sol
```diff
@@ -0,0 +1,101 @@
...
+  modifier onlyDependencyOwner(string name) {
+    require(msg.sender == _registry[name].owner);
+    _;
+  }
+
+  event OwnershipTransferred(address indexed oldOwner, address indexed newOwner);
+  event DependencyCreated(string name, address indexed owner, address indexed dependencyAddress, uint256 initialStake);
+  event Vouched(string name, uint256 amount);
+  event Unvouched(string name, uint256 amount);
+  event DependencyRemoved(string name);
```
spalladino: I'd consider indexing the `name` argument in all events. The only issue would be losing track of the original name in `DependencyCreated`, but we could add a duplicated `not-indexed-name` argument there. I'd like to know what the OZ team thinks about this, though (ping @frangio, @elopio, @nventuro).

packages/vouching/contracts/Vouching.sol
```diff
@@ -0,0 +1,101 @@
...
+    return _registry[name].dependencyAddress;
+  }
+
+  function getDependencyOwner(string name) public view returns(address) {
+    return _registry[name].owner;
+  }
+
+  function getDependencyStake(string name) public view returns(uint256) {
+    return _registry[name].stake;
+  }
```
spalladino: How about replacing these 3 functions with a `getDependency` that returns a tuple `(address, address, uint256)`?

packages/vouching/contracts/Vouching.sol
spalladino: Following a similar discussion: should `DependencyCreated` include an `initialStake`, or should we fire two subsequent events of `DependencyCreated+Vouched`? I think I prefer the current approach though.

packages/vouching/migrations/1_initial_migration.js
```diff
@@ -0,0 +1,5 @@
+var Migrations = artifacts.require("./Migrations.sol");
+
+module.exports = function(deployer) {
+  deployer.deploy(Migrations);
+};
```
spalladino: Same, let's remove these truffle leftovers

packages/vouching/package.json
```diff
@@ -0,0 +1,24 @@
...
+  "main": "truffle-config.js",
+  "directories": {
+    "test": "test"
+  },
+  "scripts": {
+    "test": "scripts/test.sh"
+  },
+  "keywords": [],
+  "author": "",
+  "license": "ISC",
```
spalladino: Aren't we using MIT so far in the other packages?

packages/vouching/package.json
```diff
@@ -0,0 +1,24 @@
+{
+  "name": "zos-vouching",
+  "version": "1.0.0",
```
spalladino: Let's follow the same version as the other packages.

packages/vouching/contracts/Vouching.sol
```diff
@@ -0,0 +1,101 @@
...
+
+  event OwnershipTransferred(address indexed oldOwner, address indexed newOwner);
+  event DependencyCreated(string name, address indexed owner, address indexed dependencyAddress, uint256 initialStake);
+  event Vouched(string name, uint256 amount);
+  event Unvouched(string name, uint256 amount);
+  event DependencyRemoved(string name);
+
+  constructor(uint256 minimumStake, ZepToken token) public {
+    require(token != address(0));
```
spalladino: Even if they are not yet supported by truffle, let's add revert reasons to all `require`s.

packages/vouching/contracts/Vouching.sol
```diff
@@ -0,0 +1,101 @@
...
+    emit DependencyCreated(name, owner, dependencyAddress, initialStake);
+  }
+
+  function transferOwnership(string name, address newOwner) external onlyDependencyOwner(name) {
+    require(newOwner != address(0));
+    _registry[name].owner = newOwner;
+    emit OwnershipTransferred(msg.sender, newOwner);
+  }
+
+  function vouch(string name, uint256 amount) external onlyDependencyOwner(name) {
```
spalladino: How do we feel about accepting vouches from other addresses?

packages/vouching/test/Vouching.test.js
```diff
@@ -0,0 +1,265 @@
...
+      describe('unvouch', function () {
+        const safeUnstakeAmount = stakeAmount.minus(minStake);
+
+        it('reverts when caller is not the dependency\'s owner', async function () {
+          await assertRevert(
+            this.vouching.unvouch(dependencyName, safeUnstakeAmount, { from: vouchingOwner })
+          );
+        });
+
+        it('reverts when the remaining stake amount is less than the minimum', async function () {
```
spalladino: Can we add a test that checks attempting to unvouch more than the current stake, so we exercise the usage of safemath there?

packages/vouching/contracts/Vouching.sol
fiiiu: Also don't like `Dependency`, but `PackageEntry` is somewhat confusing.. I don't have a better proposal though, any other ideas? `Entry` on its own is too dry?

packages/vouching/contracts/Vouching.sol
fiiiu: In particular, we might need to call other specific ZEP methods when doing the TPL integration..

packages/vouching/contracts/Vouching.sol
tinchoabbate: > 
> 
> Following a similar discussion: should `DependencyCreated` include an `initialStake`, or should we fire two subsequent events of `DependencyCreated+Vouched`? I think I prefer the current approach though.


We're *creating* an entry in the registry, so firing an event that is called `Vouched` in addition to the creation event might be somewhat confusing. IMO a single event (`DependencyCreated`) that includes the `initialStake` is clearer.

tinchoabbate:
Already pushed some updates. Now waiting for the opened discussions to be settled to make the final changes.

packages/vouching/contracts/Vouching.sol
facuspagnuolo: I'm not sure if indexing strings is still being not simple, here is some research I did for the previous release (see https://github.com/zeppelinos/zos-lib/issues/199)

packages/vouching/contracts/Vouching.sol
```diff
@@ -0,0 +1,97 @@
...
+  modifier onlyDependencyOwner(string name) {
+    require(msg.sender == _registry[name].owner, "Not allowed");
+    _;
+  }
+
+  event OwnershipTransferred(address indexed oldOwner, address indexed newOwner);
+  event DependencyCreated(string name, address indexed owner, address indexed dependencyAddress, uint256 initialStake);
+  event Vouched(string name, uint256 amount);
+  event Unvouched(string name, uint256 amount);
+  event DependencyRemoved(string name);
```
facuspagnuolo: Nit, we usually place the events right under the contract name

packages/vouching/contracts/Vouching.sol
tinchoabbate: As discussed offline, we'll keep the vouching restricted to the owner.

packages/vouching/contracts/Vouching.sol
tinchoabbate: As discussed offline, we'll move to use a generic ERC20 here to keep things decoupled.

packages/vouching/contracts/Vouching.sol
fiiiu: I remembered there was something about indexing strings, that's why I didn't.. let's wait for the OZ guys input on this!

packages/vouching/test/ZepToken.test.js
```diff
@@ -0,0 +1,30 @@
...
+  });
+
+  it('has an amount of decimals', async function () {
+    const decimals = await this.zepToken.decimals();
+    assert.equal(decimals, 18);
+  });
+
+  it('has the correct total supply', async function () {
+    const totalZEP = new BigNumber('100000000e18');
+    (await this.zepToken.totalSupply()).should.be.bignumber.equal(totalZEP);
```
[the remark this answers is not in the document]
tinchoabbate: Updated tests to use chai consistently.

packages/vouching/contracts/Vouching.sol
tinchoabbate: :+1: good to know. Updated accordingly.

packages/vouching/test/helpers/assertRevert.js
```diff
@@ -0,0 +1,16 @@
...
+  } catch (error) {
+    error.message.should.include('revert', `Expected "revert", got ${error} instead`);
+    return;
+  }
+  should.fail('Expected revert not received');
+}
+
+module.exports = {
+  assertRevert,
+};
```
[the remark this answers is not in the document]
tinchoabbate: Not quite. `lib` package is using babel and we're not, so importing from `lib` won't work.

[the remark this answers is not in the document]
spalladino: That should not be an issue, but let's review that offline later

packages/vouching/test/helpers/expectEvent.js
```diff
@@ -0,0 +1,33 @@
...
+
+      for (const [k, v] of Object.entries(eventArgs)) {
+        if (e.args[k] !== v) {
+          matches = false;
+        }
+      }
+
+      if (matches) {
+        return true;
+      }
```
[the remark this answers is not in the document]
tinchoabbate: Yes, I know. I took the methods from OpenZeppelin's repo testing suite, and seemed right to keep it the same instead of changing the code for this minor thing.

packages/vouching/test/helpers/assertRevert.js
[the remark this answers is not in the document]
tinchoabbate: Done. I was getting an `unexpected token export` error when requiring the one in `lib`. That's why initially thought it was related to ES6 modules and babel. Now it's working fine.

packages/vouching/contracts/Vouching.sol
```diff
@@ -0,0 +1,96 @@
...
+    );
+  }
+
+  function create(string name, address owner, address dependencyAddress, uint256 initialStake) external {
+    require(initialStake >= _minimumStake, "Initial stake must be equal or greater than minimum stake");
+    require(owner != address(0), "Owner address cannot be zero");
+    require(dependencyAddress != address(0), "Dependency address cannot be zero");
+    require(_registry[name].dependencyAddress == address(0), "The name has already been registered");
+
+    _token.transferFrom(owner, this, initialStake);
```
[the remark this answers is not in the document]
tinchoabbate: I tackled all instances of this issue using `SafeERC20` methods.

tinchoabbate:
I'm coming up with a possible implementation for indexing dependency names. Still testing locally. When it's ready, I'll branch out from `vouching` and then create a PR to merge again.

packages/vouching/contracts/Vouching.sol
```diff
@@ -55,9 +57,9 @@ contract Vouching is Initializable {
     require(initialStake >= _minimumStake, "Initial stake must be equal or greater than minimum stake");
     require(owner != address(0), "Owner address cannot be zero");
     require(dependencyAddress != address(0), "Dependency address cannot be zero");
-    require(_registry[name].dependencyAddress == address(0), "The name has already been registered");
+    require(_registry[name].dependencyAddress == address(0), "Given dependency name was already registererd");
```
fiiiu: NIT: reads strangely, I propose "The provided name is already registered".

packages/vouching/contracts/Vouching.sol
```diff
@@ -0,0 +1,99 @@
...
+    require(initialStake >= _minimumStake, "Initial stake must be equal or greater than minimum stake");
+    require(owner != address(0), "Owner address cannot be zero");
+    require(dependencyAddress != address(0), "Dependency address cannot be zero");
+    require(_registry[name].dependencyAddress == address(0), "Given dependency name was already registererd");
+
+    _token.safeTransferFrom(owner, this, initialStake);
+
+    _registry[name] = Dependency(owner, dependencyAddress, initialStake);
+
+    emit DependencyCreated(name, owner, dependencyAddress, initialStake);
```
facuspagnuolo: One last thing here, following the [check-effects-interactions pattern](https://solidity.readthedocs.io/en/v0.4.25/security-considerations.html?highlight=effects#use-the-checks-effects-interactions-pattern), the `transferFrom` should be the last part of this function

packages/vouching/contracts/Vouching.sol
```diff
@@ -0,0 +1,99 @@
...
+  function transferOwnership(string name, address newOwner) external onlyDependencyOwner(name) {
+    require(newOwner != address(0), "New owner address cannot be zero");
+    _registry[name].owner = newOwner;
+    emit OwnershipTransferred(msg.sender, newOwner);
+  }
+
+  function vouch(string name, uint256 amount) external onlyDependencyOwner(name) {
+    _token.safeTransferFrom(msg.sender, this, amount);
+    _registry[name].stake = _registry[name].stake.add(amount);
+    emit Vouched(name, amount);
```
facuspagnuolo: same here

packages/vouching/contracts/Vouching.sol
```diff
@@ -0,0 +1,99 @@
...
+  }
+
+  function unvouch(string name, uint256 amount) external onlyDependencyOwner(name) {
+    uint256 remainingStake = _registry[name].stake.sub(amount);
+    require(remainingStake >= _minimumStake, "Remaining stake must be equal or greater than minimum stake");
+
+    _registry[name].stake = remainingStake;
+    _token.safeTransfer(msg.sender, amount);
+
+    emit Unvouched(name, amount);
```
facuspagnuolo: same here

packages/vouching/contracts/Vouching.sol
```diff
@@ -0,0 +1,99 @@
...
+  }
+
+  function remove(string name) external onlyDependencyOwner(name) {
+    // Owner surrenders _minimumStake to the system
+    _token.safeTransfer(msg.sender, _registry[name].stake.sub(_minimumStake));
+    delete _registry[name];
+    emit DependencyRemoved(name);
```
facuspagnuolo: and here

packages/vouching/contracts/ZepToken.sol
```diff
@@ -0,0 +1,29 @@
...
+  string public constant symbol = "ZEP";
+  uint8 public constant decimals = 18;
+
+  uint256 public constant TOTAL_SUPPLY = 1e8 * (10 ** uint256(decimals));
+
+  /**
+   * @dev Constructor that gives msg.sender all of existing tokens, and sets it as the owner
+   */
+  constructor() public {
+    owner = msg.sender;
```
facuspagnuolo: you shouldn't do this manually since the `Ownable` constructor is taking care of it

packages/vouching/test/helpers/assertRevert.js
facuspagnuolo: Please remove this one since you're already using the one in `zos-lib`

packages/vouching/contracts/Vouching.sol
fiiiu: not in this one, right?

packages/vouching/contracts/Vouching.sol
tinchoabbate: Okey, good to know. Thanks facu. Should be fixed now.

tinchoabbate:
Already fixed latest issues, should be ready now.

spalladino:
@facuspagnuolo do you want to update your review, as it is blocking the merge?

Issue #116, opened 2018-09-21
Make Vouching contract upgradeable


fiiiu:
The `Vouching` contract in the `vouching` package is not yet upgradable, make it so.

PR #118, opened 2018-09-24
Make Vouching upgradeable

tinchoabbate:
This PR makes the `Vouching` contract upgradeable (#116)
Changes:
- Vouching now inherits from Initializable
- Refactored constructor to `initialize` function
- Updated tests accordingly

### DIFF
--- a/packages/vouching/package.json
+++ b/packages/vouching/package.json
@@ -13,7 +13,9 @@
   "author": "",
   "license": "MIT",
   "dependencies": {
-    "openzeppelin-solidity": "^1.12.0"
+    "openzeppelin-solidity": "^1.12.0",
+    "openzeppelin-zos": "^1.9.4",
+    "zos-lib": "2.0.0-alpha.0"
   },
   "devDependencies": {
     "chai": "^4.1.2",
--- a/packages/vouching/test/Vouching.test.js
+++ b/packages/vouching/test/Vouching.test.js
@@ -17,8 +17,9 @@ contract('Vouching', function ([_, tokenOwner, vouchingOwner, developer, transfe
   const stakeAmount = minStake.times(2);
 
   it('requires a non-null token', async function () {
+    let vouching = await Vouching.new({ from: vouchingOwner });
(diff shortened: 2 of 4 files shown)


packages/vouching/test/Vouching.test.js
```diff
@@ -17,8 +17,9 @@ contract('Vouching', function ([_, tokenOwner, vouchingOwner, developer, transfe
   const stakeAmount = minStake.times(2);
 
   it('requires a non-null token', async function () {
+    let vouching = await Vouching.new({ from: vouchingOwner });
```
spalladino: Always prefer `const` over `let`

packages/vouching/package.json
```diff
@@ -13,7 +13,9 @@
   "author": "",
   "license": "MIT",
   "dependencies": {
-    "openzeppelin-solidity": "^1.12.0"
+    "openzeppelin-solidity": "^1.12.0",
```
facuspagnuolo: I think you don't need this one, please remove if so

packages/vouching/test/Vouching.test.js
facuspagnuolo: ➕

packages/vouching/package.json
tinchoabbate: The package is still using this dependency in the ZepToken contract ([https://github.com/zeppelinos/zos/blob/vouching-upgradeable/packages/vouching/contracts/ZepToken.sol#L3](https://github.com/zeppelinos/zos/blob/vouching-upgradeable/packages/vouching/contracts/ZepToken.sol#L3)). I believe it is better to keep this PR simple, and later remove the dependency once we make ZepToken upgradeable. Do you agree ?

facuspagnuolo: sure! let's do that

tinchoabbate:
All comments addressed and CI passed, thanks to @spalladino's fix :raised_hands: . Note that we're merging into `vouching` branch.

PR #120, opened 2018-09-24
Preserve App, Package and Directory addresses

### LINKED ISSUE #35: Preserve App, Package, and Directory addresses in aborted deployments

jcarpanelli:
Fixes #35
- [x] Implement an app, package and directory address tracker in lib and cli.
- [x] Pick up already deployed contracts addresses on  future `push` runs.

### DIFF
--- a/packages/cli/src/models/network/NetworkBaseController.js
+++ b/packages/cli/src/models/network/NetworkBaseController.js
@@ -72,13 +72,17 @@ export default class NetworkBaseController {
       log.info(`Creating new version ${requestedVersion}`);
       const provider = await this.newVersion(requestedVersion);
       this.networkFile.contracts = {};
-      this._registerVersion(requestedVersion, provider.address);
+      this._registerVersion(requestedVersion, provider);
     }
   }
 
-  _registerVersion(version, providerAddress) {
-    this.networkFile.provider = { address: providerAddress };
-    this.networkFile.version = version;
+  _registerPackage({ address }) {
+    this.networkFile.package = { address }
+  }
+
+  _registerVersion(version, { address }) {
+    this.networkFile.provider = { address }
+    this.networkFile.version = version
   }
 
   async newVersion(versionName) {
--- a/packages/cli/src/models/network/NetworkBaseController.js
+++ b/packages/cli/src/models/network/NetworkBaseController.js
@@ -51,29 +51,29 @@ export default class NetworkBaseController {
   }
 
   async push(reupload = false) {
-    if (this.isDeployed) {
-      await this.fetch();
-      await this.pushVersion();
-    } else {
-      await this.deploy();
-    }
+    this._checkVersion()
+    await this.fetchOrDeploy()
 
     await Promise.all([
-      this.uploadContracts(reupload), 
+      this.uploadContracts(reupload),
       this.unsetContracts()
     ])
   }
 
-  async pushVersion() {
+  _checkVersion() {
     const requestedVersion = this.packageFile.version;
     const currentVersion = this.networkFile.version;
+
     if (requestedVersion !== currentVersion) {
       log.info(`Current version ${currentVersion}`);
       log.info(`Creating new version ${requestedVersion}`);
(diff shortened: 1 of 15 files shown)


packages/lib/src/project/AppProject.js
```diff
@@ -1,26 +1,36 @@
...
+
+      if (!await app.hasPackage(name, version)) await app.setPackage(name, thepackage.address, version)
+      const project = new this(app, name, version, txParams)
+      project.directory = directory
+      project.package = thepackage
+
+      return project
+    } catch(deployError) {
+      throw new DeployError(deployError.message, { thepackage, directory, app })
+    }
```
facuspagnuolo: What do you think about splitting this method a bit to sth like:
```js
  static async fetchOrDeploy(name = 'main', version = '0.1.0', txParams = {}, { appAddress = undefined, packageAddress = undefined }) {
    try {
      const app = await this.fetchOrDeployApp(...)
      const thepackage = await this.fetchOrDeployPackage(...)
      const directory = await this.fetchOrDeployDirectory(...)
      return this.buildProject(...)
    } catch(deployError) {
      throw new DeployError(deployError.message, { thepackage, directory, app })
    }
  }
```

packages/lib/src/project/LibProject.js
```diff
@@ -1,18 +1,30 @@
...
+        ? await thepackage.getDirectory(version)
+        : await thepackage.newVersion(version)
+
+      const project = new this(thepackage, version, txParams)
+      project.directory = directory
+
+      return project
+    } catch(deployError) {
+      throw new DeployError(deployError.message, { thepackage, directory })
+    }
```
facuspagnuolo: we may want to split this one too, it is not that large tho

packages/cli/src/models/network/NetworkBaseController.js
```diff
@@ -51,38 +51,38 @@ export default class NetworkBaseController {
...
 
     await Promise.all([
-      this.uploadContracts(reupload), 
+      this.uploadContracts(reupload),
       this.unsetContracts()
     ])
   }
 
-  async pushVersion() {
+  _checkVersion() {
```
facuspagnuolo: 👍

packages/lib/src/project/AppProject.js
```diff
@@ -1,26 +1,36 @@
...
-  static async deploy(name = 'main', version = '0.1.0', txParams = {}) {
-    const thepackage = await Package.deploy(txParams)
-    const directory = await thepackage.newVersion(version)
-    const app = await App.deploy(txParams)
-    await app.setPackage(name, thepackage.address, version)
-    const project = new this(app, name, version, txParams)
-    project.directory = directory
-    project.package = thepackage
-    return project
+  static async fetchOrDeploy(name = 'main', version = '0.1.0', txParams = {}, { appAddress = undefined, packageAddress = undefined }) {
```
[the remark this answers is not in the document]
spalladino: I'd say so, it simplifies usage of the Project class directly

spalladino:
@jcarpanelli feel free to merge after applying @facuspagnuolo's suggestions

packages/lib/src/project/AppProject.js
jcarpanelli: We've discussed this offline. It's not a good idea to merge `deploy` and `fetch` methods in `Package` and `App`, as it would add complexity. 
On the other hand, separating the logic to new static methods inside `AppProject`/`LibProject` (or `BasePackageProject`) just to make the code prettier is not reason enough to do so. In case we would want to provide this methods through `zos-lib`, we can add them, but this exceeds this PR's scope.

PR #125, opened 2018-09-25
Fix status checker, comparator and fetcher

jcarpanelli:
I'm creating this pull request because Github won't let me reopen https://github.com/zeppelinos/zos/pull/99  :cat: 

---

- Adapt `StatusChecker`, `StatusComparator` and `StatusFetcher` classes to work with the new contracts architecture and lib models implemented by @spalladino to support multiple dependencies.
- Implement multiple dependencies checker, fetcher and comparator functions.

### DIFF
--- a/packages/cli/src/models/status/StatusChecker.js
+++ b/packages/cli/src/models/status/StatusChecker.js
@@ -1,4 +1,6 @@
 import _ from 'lodash'
+import { promisify } from 'util'
+
 import { Logger, LibProject, AppProject } from 'zos-lib'
 import EventsFilter from './EventsFilter'
 import StatusFetcher from './StatusFetcher'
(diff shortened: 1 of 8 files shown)


packages/cli/src/models/status/StatusChecker.js
```diff
@@ -113,9 +118,9 @@ export default class StatusChecker {
     if (address !== expected) this.visitor.onMismatchingContractAddress(expected, address, { alias, address })
   }
 
-  _checkImplementationBytecode(alias, address) {
+  async _checkImplementationBytecode(alias, address) {
     const expected = this.networkFile.contract(alias).bodyBytecodeHash
-    const observed = bytecodeDigest(web3.eth.getCode(address))
+    const observed = bytecodeDigest(await promisify(web3.eth.getCode.bind(web3.eth))(address))
```
facuspagnuolo: nice catch! we should have a common file providing this promisified functions of web3

jcarpanelli: Agree. We should also consider using [this](http://bluebirdjs.com/docs/api/promise.promisifyall.html)

PR #126, opened 2018-09-25
Remove old application/versioning test dir

facuspagnuolo:
Just removing an old dir left after refactoring the lib contracts architecture


spalladino:
Good catch!

Issue #64, opened 2018-09-03
Allow running `zos` from any subfolder


spalladino:
Most CLI apps (such as git or npm) support being run from any subfolder within the project, as they crawl the folder hierarchy upwards until they reach a folder with a `.git` or `package.json` that signal the root. Given that we have a `zos.json` that indicates the project root, we should support running `zos` from any subfolder within the project.

gitcoinbot:
Issue Status: 1. **Open** 2. Started 3. Submitted 4. Done 

<hr>

__This issue now has a funding of 190.0 DAI (190.0 USD @ $1.0/DAI)  attached to it__ as part of the Ethereum Foundation fund__.__

 * If you would like to work on this issue you can 'start work' [on the Gitcoin Issue Details page](https://gitcoin.co/issue/zeppelinos/zos/64/1274).
* Want to chip in? Add your own contribution [here](https://gitcoin.co/issue/zeppelinos/zos/64/1274).
* Questions? Checkout <a href='https://gitcoin.co/help'>Gitcoin Help</a> or the <a href='https://gitcoin.co/slack'>Gitcoin Slack</a>
* $23,090.04 more funded OSS Work available on the [Gitcoin Issue Explorer](https://gitcoin.co/explorer)

gitcoinbot:
Issue Status: 1. **Open** 2. Started 3. Submitted 4. Done 

<hr>

__Workers have applied to start work__.


These users each claimed they can complete the work by 12 months from now.
Please review their action plans below:


**1) [pvgupta24](https://gitcoin.co/profile/pvgupta24) has applied to start work _(Funders only: [approve worker](https://gitcoin.co/issue/zeppelinos/zos/64/1274?mutate_worker_action=approve&worker=pvgupta24) | [reject worker](https://gitcoin.co/issue/zeppelinos/zos/64/1274?mutate_worker_action=reject&worker=pvgupta24))_.**

- Understand the relevant code base for the zos CLI package
- Understand how other CLI apps crawl up the folder hierarchy to signal the root
- Clarifications on codebase and the issue
- Make the minimum necessary changes
- Test the changes from subfolders
- Send a PR and address the requested changes

Learn more [on the Gitcoin Issue Details page](https://gitcoin.co/issue/zeppelinos/zos/64/1274).

vs77bb:
@zachzundel You've been approved here. Please do let us know if you have any questions and a WIP PR sooner than later always helps 👍

zachzundel:
[...]

PR #130, opened 2018-09-25
Allow zos to run from anywhere in the project

zachzundel:
## Description
This PR changes the ZosPackageFile.js to crawl up to root looking for a package file. (Addressing issue #64)

## Still Needs
Tests


packages/cli/src/models/files/ZosPackageFile.js
```diff
@@ -9,7 +9,25 @@ export default class ZosPackageFile {
...
+    while(!fs.exists(this.fileName)) {
+        if(process.cwd() === "/") {
+            process.chdir(cwd)
+            break
+        }
+
+        try {
+            process.chdir("..")
+        } catch (error) {
+            console.error("couldn't find zos.json")
```
spalladino: Let's throw instead of just logging

zachzundel:
Ready for review!

packages/cli/src/models/files/ZosPackageFile.js
```diff
@@ -9,7 +9,24 @@ export default class ZosPackageFile {
 
   constructor(fileName = 'zos.json') {
```
spalladino: Note that this discovery should not apply to all commands. For instance, `zos init` should create a new project without doing any lookup. That said, this may not be the best place for this logic. Check if `ControllerFor` may work best, with a flag to prevent the lookup, used if called from zos init.

tests/cli-app/test/integration.test.js
```diff
@@ -73,6 +73,14 @@ describe(`cli-app on ${network}`, function () {
     truffleExec(`run.js WithToken 0 say --network ${network} --from ${this.from}`).should.eq('TKN')
   })
 
+  it('creates an instance from a dependency from inside a subfolder', function () {
+    run(`cd contracts`)
```
spalladino: Note that `run` automatically sets the workdir, so this call has no effect. The correct way to test it should be adding a new `cwd` option to `run`.

packages/cli/src/models/files/ZosPackageFile.js
spalladino: It may also cause issues if we want to retrieve the list of contracts for a dependency. If the zos.json is missing, then this would climb up to the project root, and find the project's zos.json file.

tests/cli-app/test/integration.test.js
zachzundel: Fixed.

packages/cli/src/models/files/ZosPackageFile.js
zachzundel: So I've added a commit that fixes your first point, if it is running an init command then it won't climb through directories looking for the zosfile. 
I'm not sure about the second point, though, how would a dependency in node_modules (which would have to have been installed by NPM) be broken enough to not have a zos.json file?

spalladino:
Great! The PR looks good. Please add that test, and check out the error in the CI, and we should be good to merge after that.

spalladino:
@zachzundel it seems there was an issue with integration tests on `master`, which PR #161 should fix. Please try rebasing your branch on top of master after that PR is fixed, to see if the CI passes.

vs77bb:
@spalladino Is this one back ready for your review? Hope all is well 🙂

spalladino:
@vs77bb there is still a conflict that prevents merging, as soon as it's fixed I think we're good to go. @zachzundel do you think you can take care of it?

zachzundel:
I can take care of it, but I'm working on an issue with the .zos.lock file.

zachzundel:
Hi @spalladino, I think I've solved the merge and lockfiles issue, please take another look. I have a lingering feeling I'm forgetting something but I can't remember what.

spalladino:
Thanks for the contribution @zachzundel!

spalladino:
One question that comes to mind now though @zachzundel: how is the `.zos.lock` lookup working? For instance: if I run a command from a subfolder, and I simultaneously attempt to run another command from another subfolder, will the `.zos.lock` effectively prevent me from doing so? Or will there be two different `.zos.lock` files, one on each subfolder, allowing me to run two commands at the same time (which is something we want to prevent)?

zachzundel:
There will be one in each sub folder. If we wanted to have there be one for the project, we would have to move the creation of the lockfile from the zos-cli.js file to somewhere like ZosPackageLock.js, and I wasn’t confident enough that code was called on every run like zos-cli.js

spalladino:
Got it. I'll open a new issue to work on that, and agree that we'd need to move that code somewhere else. Not sure which would be the best location though, will have to look into it. Thanks!

Issue #132, opened 2018-09-25
Consider indexing name argument in all Vouching events


tinchoabbate:
@spalladino [suggested in PR #115](https://github.com/zeppelinos/zos/pull/115#discussion_r219619581) indexing name arguments for all events fired in the `Vouching` contract.

I'm moving the discussion outside the PR so it can be better documented.

Related issues were the problem of indexing string arguments has been discussed:
- [in OpenZeppelin](https://github.com/OpenZeppelin/openzeppelin-solidity/issues/992)
- [in zOS](https://github.com/zeppelinos/zos-lib/issues/199)

spalladino:
Following the discussion in https://github.com/OpenZeppelin/openzeppelin-solidity/issues/992, it'd seem like the best course of action would be to make `name` indexed in all events, and add a `fullname` unindexed argument in `DependencyCreated`. WDYT?

tinchoabbate:
Reviewed the comment you left in the PR, and I've been working locally on a new implementation that addresses them in some ways.

Quick question, do we need the `name` to be a `string`? Could it instead be a `bytes32`? In my tests I'm seeing that (as you mentioned in the PR), `bytes32 indexed` event arguments are not being hashed. Furthermore, it seems that only fixed-size types can be indexed (see https://github.com/trufflesuite/truffle/issues/405#issuecomment-312067312).

PR #133, opened 2018-09-25
Index names in Vouching events using name hash

tinchoabbate:
This PR includes a possible implementation that tackles #132 . I'm opening it to start discussing the issue over real code.

Here I'm basically indexing by the keccak256 hash of the `name` argument. The actual name (not the hash) is logged in the `DependencyCreated` event. Tests were updated as well.

Not entirely sure of the efficiency of the implementation though, since we'd be hashing the `name` everytime the contract fires an event. I'm eager to hear reviews and comments on this.

~EDIT: note that I'm merging into `vouching`~

### DIFF
--- a/packages/vouching/contracts/Vouching.sol
+++ b/packages/vouching/contracts/Vouching.sol
@@ -8,10 +8,16 @@ import "openzeppelin-zos/contracts/Initializable.sol";
 
 contract Vouching is Initializable {
   event OwnershipTransferred(address indexed oldOwner, address indexed newOwner);
-  event DependencyCreated(string name, address indexed owner, address indexed dependencyAddress, uint256 initialStake);
-  event Vouched(string name, uint256 amount);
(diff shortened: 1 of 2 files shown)


spalladino:
Interesting. I noted the issue you created first, and left my suggestion there. Nice to see that the implementation actually matches the suggestion :-P

That said, note that you shouldn't need to manually hash the name before sending it to the event. Solidity should take care of that automatically, and I _think_ that it hashes only if the string is longer than 32 bytes. Either way, the language takes care of that for you.

Regarding the tests, I'd avoid testing for the hash value. You don't really care how the value is hashed, you only care that the event can be searched by the indexed value. So, instead of testing for the hashed value, I'd register two different dependencies (and get the indexed name from the events emiited by those txs) and vouch for them, use a filter to retrieve the events related to only one of them, and check that the events from the other are not returned. 

Let me know if it's clear!

PR #137, opened 2018-09-25
Track logic contracts in SimpleProject

### LINKED ISSUE #72: Track created logic contracts in lightweight project for reuse
The lightweight project from #70 should keep track of all logic contracts created, so when a new instance is requested, the same logic contract can be reused. This implies tracking not just the name of the contract, but also its bytecode hash to detect changes. We can track the same info as we do now in the CLI.

spalladino:
Reuses the same logic contract for multiple proxies. Supports tracking bytecode changes to optionally redeploy if the contract class changes. And since we were at it, use upgradeAndCall from Proxy to make atomic upgrades.

Fixes #72

### DIFF
--- a/packages/lib/package.json
+++ b/packages/lib/package.json
@@ -58,6 +58,7 @@
     "openzeppelin-solidity": "~1.10.0",
     "truffle-config": "^1.0.6",
     "truffle-contract": "^3.0.6",
+    "truffle-flattener": "github:alcuadrado/truffle-flattener#master",
     "truffle-provisioner": "^0.1.1",
     "web3": "^0.18.4"
   }
--- a/packages/lib/package.json
+++ b/packages/lib/package.json
@@ -58,7 +58,7 @@
     "openzeppelin-solidity": "~1.10.0",
     "truffle-config": "^1.0.6",
     "truffle-contract": "^3.0.6",
-    "truffle-flattener": "github:alcuadrado/truffle-flattener#master",
+    "truffle-flattener": "^1.2.8",
     "truffle-provisioner": "^0.1.1",
     "web3": "^0.18.4"
   }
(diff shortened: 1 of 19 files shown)


packages/lib/src/project/SimpleProject.js
```diff
@@ -32,12 +33,12 @@ export default class SimpleProject  {
...
 
-  async upgradeProxy(proxyAddress, contractClass, { initMethod, initArgs, initFrom } = {}) {
+  async upgradeProxy(proxyAddress, contractClass, { contractName, initMethod, initArgs, initFrom, redeployIfChanged } = {}) {
     proxyAddress = toAddress(proxyAddress)
-    const implementation = await this._deployImplementation(contractClass);    
+    const implementation = await this._getOrDeployImplementation(contractClass, contractName, redeployIfChanged);    
     log.info(`Upgrading proxy to new logic contract at ${implementation.address}`)
     const proxy = Proxy.at(proxyAddress, this.txParams)
-    await proxy.upgradeTo(implementation)    
+    await proxy.upgradeTo(implementation) // TODO: Use upgradeToAndCall!
```
facuspagnuolo: we should do sth like we are doing in `App#upgradeProxy` right? I mean, if there is no `initMethod` we should call `upgradeTo`, if there is then we should call `upgradeToAndCall`, am I right?

packages/lib/src/proxy/Proxy.js
```diff
@@ -21,9 +21,11 @@ export default class Proxy {
     this.txParams = txParams
   }
 
-  async upgradeTo(address) {
+  async upgradeTo(address, migrateData) {
```
facuspagnuolo: awesome

packages/lib/package.json
```diff
@@ -57,6 +57,7 @@
     "openzeppelin-solidity": "~1.10.0",
     "truffle-config": "^1.0.6",
     "truffle-contract": "^3.0.6",
+    "truffle-flattener": "github:alcuadrado/truffle-flattener#master",
```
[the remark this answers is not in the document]
jcarpanelli: @spalladino, the latest released version of this library includes module exporting. We should be ok to switch to `truffle-flattener@1.2.8` now.

facuspagnuolo:
@spalladino the CI is complaining because of the version requested for `truffle-flattener` I think

spalladino:
Forcing merge as Travis has passed but Github hasn't noticed

PR #140, opened 2018-09-26
Common interface for Package contract

spalladino:


packages/lib/contracts/application/App.sol
```diff
@@ -85,7 +85,7 @@ contract App is Ownable {
   function unsetPackage(string packageName) public onlyOwner {
     require(address(providers[packageName].package) != address(0), "Package to unset not found");
     delete providers[packageName];
-    emit PackageChanged(packageName, address(0), "");
+    emit PackageChanged(packageName, address(0), [uint64(0), uint64(0), uint64(0)]);
```
facuspagnuolo: What do you think about having sth like `PackageSet` and `PackageUnset` events instead?

packages/lib/contracts/application/Package.sol
```diff
@@ -11,55 +11,98 @@ import "openzeppelin-solidity/contracts/ownership/Ownable.sol";
...
+    Version storage version = versions[majorToLatestVersion[major]];
+    return (version.semanticVersion, version.contractAddress, version.contentURI); 
+  }
+
+  function semanticVersionHash(uint64[3] version) internal pure returns (bytes32) {
+    return keccak256(abi.encodePacked(version[0], version[1], version[2]));
+  }
+
+  function semanticVersionIsZero(uint64[3] version) internal pure returns (bool) {
+    return version[0] == 0 && version[1] == 0 && version[2] == 0;
```
facuspagnuolo: I know that we work towards simplifying the architecture of contracts, but it may make sense having the `interface` in a separate file to be clearer

packages/lib/contracts/application/App.sol
spalladino: I think that having a single event is more useful in terms of watching the contract. If you are tracking the status of a particular package (or all packages), tracking only `PackageSet` events does not work, since you'd need the `Unset` as well. In other words, it wouldn't make sense to filter by Set events and not by Unset events at the same time; so it's just simpler for the client to merge them in a single event. OTOH, it feels odd to be filling args with empty values.

Anyway, I have no strong preference. I'd keep it as `Changed` just because it's the same pattern we're using in the ImplementationDirectory as well, and I don't see a reason to change that interface. But I'm ok with making the change if you think it's for the best.

spalladino:
> I'd add tests for the new functions and the improvements for addVersion

Absolutely, will be working on that today

spalladino:
@facuspagnuolo ready to review

packages/docs/docs/docs/api_application_Package.md
```diff
@@ -0,0 +1,6 @@
...
...</dt><dd>true if the version is registered in this package, false otherwise.</dd></dl></div></div></li><li><div class="item function"><span id="semanticVersionHash" class="anchor-marker"></span><h4 class="name">semanticVersionHash</h4><div class="body"><code class="signature">function <strong>semanticVersionHash</strong><span>(uint64[] version) </span><span>internal </span><span>pure </span><span>returns  (bytes32) </span></code><hr/><dl><dt><span class="label-parameters">Parameters:</span></dt><dd><div><code>version</code> - uint64[]</div></dd><dt><span class="label-return">Returns:</span></dt><dd>bytes32</dd></dl></div></div></li><li><div class="item function"><span id="semanticVersionIsZero" class="anchor-marker"></span><h4 class="name">semanticVersionIsZero</h4><div class="body"><code class="signature">function <strong>semanticVersionIsZero</strong><span>(uint64[] version) </span><span>internal </span><span>pure </span><span>returns  (bool) </span></code><hr/><dl><dt><span class="label-parameters">Parameters:</span></dt><dd><div><code>version</code> - uint64[]</div></dd><dt><span class="label-return">Returns:</span></dt><dd>bool</dd></dl></div></div></li></ul></div></div></div>
```
facuspagnuolo: 😨

PR #142, opened 2018-09-28
Check for initializer to call base contract initializers.

### LINKED ISSUE #52: Warn if a contract initializer fails to call a parent initializer
When writing an initializer function, all parent initializers must be manually called. We could use the AST to check if there are any base contracts, and if they declare their own `initialize` functions, and then check if the contract's `initialize` function does call them. If it doesn't, we can show a warning to the user.

For instance:

```solidity
contract BaseContract {
  uint256 public y;

  function initialize() public {
    y = 42;
  }
}

contract MyContract is BaseContract {
  uint256 public x;

  function initialize(uint256 _x)  public {
    BaseContract.initialize(); // Warn if this call is missing!
    x = _x;
  }
}
```

Destiner:
Implemented `getUninitializedBaseContracts` function, which checks whether given contract has any Base Contracts with `initialize()` that weren't initialized in the child contract. Added function to the `add`.

Fixes #52.


packages/cli/src/models/local/LocalBaseController.js
```diff
@@ -110,6 +115,57 @@ export default class LocalBaseController {
...
+        baseContractsWithInitialize[baseContractName] = false
+      }
+    }
+    // Find "initialize" function in the AST
+    const contractDefinition = contractJson.ast.nodes
+      .find(n => n.nodeType === "ContractDefinition" && n.name === contractJson.contractName)
+    const initializeFunction = contractDefinition.nodes
+      .find(n => n.nodeType === "FunctionDefinition" && n.name === "initialize")
+    if (initializeFunction === undefined) {
+      return []
```
spalladino: This is an interesting scenario: what if the child contract has no initialize function, but its parent(s) do?

If the child contract has a single parent, then the user can call that `initialize` directly, as the child will inherit it. But what if it has multiple parents, each with its own `initialize`? In that case, it'd seem like the child contract **must** have an `initialize` to call its parents', otherwise only one of the parents would get initialized.

Using contracts like the ones you defined:
```solidity
contract ShouldHaveInitialize is WithInitialize, AnotherWithInitialize {
}

contract DoesNotNeedAnInitialize is WithInitialize {
}
```

Let me know if you agree, or if I'm missing something!

packages/cli/src/models/local/LocalBaseController.js
```diff
@@ -110,6 +115,57 @@ export default class LocalBaseController {
...
+    if (baseContracts.length == 0) {
+      return []
+    }
+    // Make a dict of base contracts that have "initialize" function
+    const baseContractsWithInitialize = {}
+    for (const baseContract of baseContracts) {
+      const baseContractName = baseContract.baseName.name
+      const baseContractPath = Contracts.getLocalPath(baseContractName)
+      const baseContractAbi = fs.parseJson(baseContractPath).abi
+      if (baseContractAbi.find(fn => fn.type === "function" && fn.name === "initialize")) {
```
spalladino: Could we generalize this check to also include functions that have a modifier named [`initializer`](https://github.com/zeppelinos/zos/blob/master/packages/lib/contracts/Initializable.sol#L31), even if the function name is not `initialize`?

packages/cli/src/models/local/LocalBaseController.js
```diff
@@ -110,6 +115,57 @@ export default class LocalBaseController {
...
+        continue
+      }
+      if (statement.expression.nodeType !== "FunctionCall") {
+        continue
+      }
+      if (statement.expression.expression.memberName !== "initialize") {
+        continue
+      }
+      const baseContractName = statement.expression.expression.expression.name
+      initializedContracts[baseContractName] = true
```
spalladino: Minor style comment: could we rephrase these checks as the following?
```js
if (statement.nodeType === "ExpressionStatement" && statement.expression.nodeType === "FunctionCall" && statement.expression.expression.memberName === "initialize") {
  const baseContractName = statement.expression.expression.expression.name
  initializedContracts[baseContractName] = true
}
```

packages/cli/src/models/local/LocalBaseController.js
Destiner: That's right!

packages/cli/src/models/local/LocalBaseController.js
Destiner: Yes, I'll add check for this.

packages/cli/src/models/local/LocalBaseController.js
Destiner: Sure!

Destiner:
Made updates based on the review. Here's a quick overview of what's changed.
Code:
* Now initializer can be defined either by name or by modifier, and we'll find it with `getContractInitializer`. This is true for base contracts and a child contract.
* Fixed code style issue.

Tests:
* Add contracts to test extended functionality, as you suggested (`ShouldHaveInitialize` and `DoesNotNeedAnInitialize`).
* `AnotherWithInitialize` now initialized with `initializer` modifier! Thus we can test both types of initializers easily.

spalladino:
Great work @Destiner! Can I trouble you for one more test, which I'm not sure if it'll pass? 

Having the following contracts:
```solidity
contract ShouldHaveInitialize is WithInitialize, AnotherWithInitialize {
  uint public someValue;
}

contract ExtendsFromShouldHaveInitialize is ShouldHaveInitialize {
  uint public someOtherValue;
}
```

Then adding `ExtendsFromShouldHaveInitialize` to zos should display the same warning as if you were adding `ShouldHaveInitialize`.

Destiner:
Yeah, I thought a lot about this  case. It actually won't show the warning, you're right. We can show the warning, but it'll require recursive check of all base contracts. More over, we'll see the warning anyway when we'll add `ShouldHaveInitialize`, so it may be a bit redundant anyway.

What do you think?

spalladino:
> More over, we'll see the warning anyway when we'll add ShouldHaveInitialize, so it may be a bit redundant anyway.

The thing is that you may choose *not* to add ShouldHaveInitialize to your zos project, as you will probably just want to `add` the childrenmost contracts in your inheritance chain. Actually, the scenario where you add both parent and child should be the most infrequent.

How difficult do you think it'd be to add the recursive check you mention? I think there is value in it. In the meantime, I'll merge the PR so we can keep working on top of it.

Destiner:
It's actually quite easy to implement. I'll test my implementation and will submit later. For now, I can say that it works on test case that you provided and doesn't introduce regressions.

As long as most of the time we don't use `add --all`, I think we're safe.

Though recursion always introduce complexity. Say, we have `contract A is B, C`, then `contract B is D`, `contract C is D`, and finally `contract D is WithInitialize, WithAnotherInitialize`. `A`, `B`, `C`, `D` doesn't have initializer. Even if we only `add A`, we'll receive, I guess, 2 warnings about D.

So, as long as we don't want to start catching cases like this, I'm OK. Or, at least we need to make a list of such cases beforehand :)

spalladino:
> Even if we only add A, we'll receive, I guess, 2 warnings about D.

Can we recursively collect all the contracts that would throw a warning, run a `_.uniq`, and then display each warning just once? If so, I'd go with this approach.

spalladino:
Sounds great, thanks @Destiner! And yes please, open a new PR on master. Thanks!!

PR #145, opened 2018-09-28
Add TPL integration tests

facuspagnuolo:
Fixes https://github.com/zeppelinos/zos/issues/135


packages/vouching/test/contracts/ZEPToken.test.js
```diff
@@ -2,18 +2,27 @@ import { encodeCall, assertRevert } from 'zos-lib'
...
+const BasicJurisdiction = artifacts.require('BasicJurisdiction');
 
 require('chai')
   .use(require('chai-bignumber')(BigNumber))
   .should();
 
-contract('ZEPToken', ([ _, owner, another, jurisdiction]) => {
-  const attributeID = 0;
+contract('ZEPToken', ([ _, tokenOwner, another, jurisdictionOwner, validatorOwner, zeppelin ]) => {
+  const receiveTokensAttributeID = 0
```
spalladino: I'd use a non-zero ID for testing, because EVM

packages/vouching/test/contracts/ZEPToken.test.js
```diff
@@ -37,10 +46,122 @@ contract('ZEPToken', ([ _, owner, another, jurisdiction]) => {
...
+
+      await this.jurisdiction.addValidator(this.validator.address, "ZEP Validator", { from: jurisdictionOwner })
+      await this.jurisdiction.addAttributeType(receiveTokensAttributeID, false, false, ZERO_ADDRESS, 0, 0, 0, "can transfer", { from: jurisdictionOwner })
+      await this.jurisdiction.addValidatorApproval(this.validator.address, receiveTokensAttributeID, { from: jurisdictionOwner })
+      await this.validator.addOrganization(zeppelin, 100, "ZEP Org", { from: validatorOwner })
+    })
+
+    function assertTokensCannotBeTransferred() {
+      it('cannot transfer', async function () {
+        assert.equal(await this.zepToken.canTransfer(recipient, amount, { from: sender }), false)
```
spalladino: Minor: why not `should`?

packages/vouching/test/contracts/ZEPToken.test.js
```diff
@@ -37,10 +46,122 @@ contract('ZEPToken', ([ _, owner, another, jurisdiction]) => {
...
+        await this.zepToken.approve(recipient, amount, { from: sender })
+
+        assert(await this.zepToken.canTransferFrom(sender, recipient, amount, { from: recipient }))
+        await this.zepToken.transferFrom(sender, recipient, amount, { from: recipient })
+
+        assert((await this.zepToken.balanceOf(recipient)).eq(amount))
+      })
+    }
+
+    describe('when the no one was not allowed to receive tokens', function () {
```
spalladino: "the no one" sounds like a creature out of H P Lovecraft :-P

spalladino:
Looks good! Left some very minor comments. Something that came to my mind though: if the tokens do reach an unintended owner, we have no way to actually freeze them. We can only control recipients in txs, but not senders. Should we consider adding a check in TPLToken that the whitelist attribute is required for both sender and recipient, and not just recipient?

packages/vouching/test/contracts/ZEPToken.test.js
facuspagnuolo: bad copy paste 😂

packages/vouching/test/contracts/ZEPToken.test.js
facuspagnuolo: For some reason, the token approvals fail when using `should`

spalladino:
@facuspagnuolo note there are conflicts on `Vouching.test.js`, I think we can merge right after they are resolved

Issue #159, opened 2018-10-02
Detect incompatible Package versions


facuspagnuolo:
Given that we are breaking backward compatibility with the `Package` implementation for the next release, we will need to think a way for the CLI to detect and warn the user in case they try to link to a lib (e.g. OZ-zos 1.9.x) that is backed by a  previous `Package` version.

A possible solution could be to detect usages of old `Package`s with the new version of the CLI, and also provide a patch for `zos@1.9.x` to detect links to libs backed by the new implementation of `Package`.

spalladino:
Will implement via `zosversion` field check in zos.json files as discussed in https://github.com/zeppelinos/zos/issues/131

spalladino:
Fixed in #162 and #163

Issue #131, opened 2018-09-25
Handle breaking changes from 2.0 via json version check


spalladino:
We are introducing breaking changes as part of version 2, which may cause both contracts and zos.json files from previous versions to be incompatible with the current one, and viceversa. The main breaking changes are:

1. Support for multiple dependencies: this affects the `App` contract, which is incompatible with the one from 1.x. It also changes the format of zos.json files (both local and network), since the dependencies list is now an array, and not a single `stdlib` entry.
2. New Package contract interface, for compatibility with aOS: Packages will enforce semver, and have a different interface.

Note that 2 is still not implemented, but if it is, we would need to:
- [x] Add a flag to all json files from version 2.0 indicating that they belong to the new version (implemented as `zosversion: '2'`)
- [x] Release a patch of 1.4.x that checks whether that flag is present in any zos.json file (local or network, from the project or a stdlib), and if it is, aborts execution.
- [x] On version 2.x (master), check that zosversion is '2' on all zos json files, and refuse to use them otherwise.

spalladino:
Fixed in #162 and #163

PR #162, opened 2018-10-02
Enforce zosversion check when loading zos json files

spalladino:
When loading a zos.json or zos.network.json file, either local or from a dependency, it is now checked that the json file contains a zosversion equal to 2, or an error message is thrown.

This PR is part of the fixes for issues #159 and #131. Another PR is pending to implement the same logic in 1.4.x.


spalladino:
CI is failing with the following error:
> zos version identifier not found in node_modules/openzeppelin-zos/zos.json. This means the project was built with an older version of zos (1.x), and needs to be upgraded. Please refer to the documentation at https://docs.zeppelinos.org for more info.

I hate it when my code is smarter than me.

PR #176, opened 2018-10-04
Update package locks

facuspagnuolo:
Output of running `npx lerna bootstrap`


spalladino:
Interesting, the output of running `npx lerna bootstrap` on my machine are the package-lock files with specific versions, instead of ranges. Are you running `lerna v3.3.2`?

facuspagnuolo:
I'm running `lerna v3.1.4`, what about `npm`? I'm using `6.1.0`

spalladino:
It was my bad, I was on npm 5.x.

PR #177, opened 2018-10-04
Export commands and scripts

### LINKED ISSUE #40: Export cli scripts for using them programmatically.
It would be great (and pretty straightforward to implement) if we provide a programmatic way of using the cli by exporting all the scripts. Users would be able to do stuff like:

```javascript
import zos from 'zos' // or import { init, add, push } from 'zos'
...
zos.init({ version: '0.0.1', name: 'myAwesomeProject'})
zos.add({ contractData: [{ name: 'myContract', alias: 'awesomeContract' }]})
zos.push({ network: 'ropsten' })
...
```
Also, as this would be a new functionality, we should _add the appropiate documentation to `docs`_.

facuspagnuolo:
Fixes #40 

I exported both scripts and commands

### DIFF
--- a/packages/cli/src/index.js
+++ b/packages/cli/src/index.js
@@ -2,19 +2,17 @@
 const version = 'v' + require('../package.json').version
 
 // model objects
+import commands from './commands'
+import scripts from './scripts'
+import local from './models/local'
+import network from './models/network'
 import TestHelper from './models/TestHelper'
-import ControllerFor from './models/local/ControllerFor'
-import LocalAppController from './models/local/LocalAppController'
-import LocalLibController from './models/local/LocalLibController'
-import NetworkAppController from './models/network/NetworkAppController'
-import NetworkLibController from './models/network/NetworkLibController'
 
 export {
   version,
+  local,
+  network,
+  commands,
+  scripts,
   TestHelper,
-  ControllerFor,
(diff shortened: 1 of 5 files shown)


packages/cli/src/commands/index.js
```diff
@@ -1,16 +1,31 @@
...
+import link from './link'
+import push from './push'
 import remove from './remove'
-import verify from './verify'
+import session from './session'
 import setAdmin from './set-admin'
+import status from './status'
+import unlink from './unlink'
+import update from './update'
+import verify from './verify'
```
spalladino: Dat OCD

packages/cli/src/index.js
```diff
@@ -2,19 +2,17 @@
...
-import ControllerFor from './models/local/ControllerFor'
-import LocalAppController from './models/local/LocalAppController'
-import LocalLibController from './models/local/LocalLibController'
-import NetworkAppController from './models/network/NetworkAppController'
-import NetworkLibController from './models/network/NetworkLibController'
 
 export {
   version,
+  local,
+  network,
```
spalladino: I like the idea of exporting the Controllers as well!

facuspagnuolo:
Let's add `runWithTruffle`, agree :)

Issue #165, opened 2018-10-03
Make CLI lightweight by default providing a migration path


facuspagnuolo:
Following up the changes made in https://github.com/zeppelinos/zos/pull/146, the CLI does not work in a lightweight mode by default, it can either work with `SimpleProjects` or `BasePackageProjects` but not with both. The idea would be to improve the CLI to work in lightweight mode by default (simple project) and provide a way to migrate `SimpleProjects` to `BasePackageProjects`

Here are some things that we may want to consider that also follow the lightweight idea:
- The `bump` command should only modify the version of the `zos.<network>.json` file working as an of-chain piece of information.
- The new `publish` command should deploy the whole zos-lib architecture to the network updating the `zos.<network>.json` file with the respective addresses of the app and package contracts
- We should also make sure that `add`, `push`, `create` and `update` commands work properly after publishing an `App`. 
- What about `status` and `freeze`? Maybe these commands should not be available for lightweight projects.
- In this case `publish` will be used by users only once, does that make sense? maybe there is a way this could be integrated with `bump`, or maybe we may want to merge both `publish` and `push` commands in a future

PR #195, opened 2018-10-08
Add --full-app flag to push command

spalladino:
Migrates a lightweight app to a full app.
Related to #165 (pending making lightweight app the default).

### DIFF
--- a/packages/cli/src/commands/push.js
+++ b/packages/cli/src/commands/push.js
@@ -16,13 +16,14 @@ const register = program => program
   .option('-d, --deploy-libs', 'deploys libraries to the network if there is no existing deployment')
   .option('--reset', 'redeploys all contracts (not only the ones that changed)')
   .option('-f, --force', 'ignores validation errors and deploys contracts')
+  .option('--full-app', 'deploys an App contract to manage the project')
   .withNetworkOptions()
   .action(action)
 
 async function action(options) {
-  const { skipCompile, deployLibs, force, reset: reupload } = options
+  const { skipCompile, deployLibs, force, reset: reupload, fullApp } = options
   await runWithTruffle(
-    async (opts) => await push({ force, deployLibs, reupload, ... opts }),
+    async (opts) => await push({ force, deployLibs, reupload, fullApp, ... opts }),
     { compile: !skipCompile, ... options }
   )
 }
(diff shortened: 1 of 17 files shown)


packages/cli/src/models/network/NetworkAppController.js
```diff
@@ -9,8 +9,14 @@ import { AppProjectDeployer, SimpleProjectDeployer } from './ProjectDeployer';
 const log = new Logger('NetworkAppController');
 
 export default class NetworkAppController extends NetworkBaseController {
+  get isLightweight() {
+    return this.packageFile.isLightweight && !this.appAddress;
```
facuspagnuolo: I think we can rely only on the app address

packages/lib/src/project/SimpleProject.js
```diff
@@ -58,7 +58,7 @@ export default class SimpleProject  {
     delete this.implementations[contractName]
   }
 
-  registerImplementation(contractName, { address, bytecodeHash }) {
+  async registerImplementation(contractName, { address, bytecodeHash }) {
```
facuspagnuolo: is it `async`?

packages/cli/src/commands/push.js
```diff
@@ -16,13 +16,14 @@ const register = program => program
   .option('-d, --deploy-libs', 'deploys libraries to the network if there is no existing deployment')
   .option('--reset', 'redeploys all contracts (not only the ones that changed)')
   .option('-f, --force', 'ignores validation errors and deploys contracts')
+  .option('--full-app', 'deploys an App contract to manage the project')
```
facuspagnuolo: leaving this comment to remind us that we are thinking of moving this flag to a new `publish` command

packages/lib/src/project/SimpleProject.js
spalladino: It's just to be consistent with the equivalent method in AppProject

packages/cli/src/models/network/NetworkAppController.js
spalladino: isLightweight will return true by default soon!

spalladino:
Closing as was already merged as part of https://github.com/zeppelinos/zos/pull/231

PR #197, opened 2018-10-09
Improve ABI parsing for initialization

### LINKED ISSUE #194: Handle multiple initializers with same arguments length during proxies initialization
Currently, when a user runs `zos create MyContract --init myInitMethod --args a1,a2,...,aN` we have a lookup function that fetches the initialization method to be called from the ABI matching the given name and the number of arguments. The thing is that there could be contracts inheriting from other initializable contracts having different initialization methods with the same number of arguments but different types. This is the case for `ZepToken` and `ERC20Detailed`. And the problem is that this information is flattened in the ABI output, so we cannot tell which is the method that a user wants to call actually.

In order to fix this, I suggest fetching the initialize function assuming that every user is willing to call the initialize function of the most derived contract of the hierarchy. We can walk the AST to have that information and then fetch the function of the ABI matching the input types as well.

spalladino:
Fixes #194

### DIFF
--- a/packages/cli/src/models/local/LocalBaseController.js
+++ b/packages/cli/src/models/local/LocalBaseController.js
@@ -200,6 +200,7 @@ export default class LocalBaseController {
     return false
   }
 
+  // TODO: Use ABIs.getFunction here
   getContractInitializer(contractJson) {
     const contractDefinition = contractJson.ast.nodes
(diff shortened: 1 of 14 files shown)


facuspagnuolo:
@spalladino please fix the CI and merge then

PR #214, opened 2018-10-09
Add skip-compile option to push options

jcarpanelli:
Related to https://github.com/zeppelinos/zos/issues/67, this PR doesn't fix it though, as there is nothing to fix, but adds the `skip-compile` option to the `withPushOptions` function, to have the posibility to whether run `truffle compile` or not.

### DIFF
--- a/packages/cli/src/bin/options.js
+++ b/packages/cli/src/bin/options.js
@@ -16,6 +16,7 @@ program.Command.prototype.withNetworkOptions = function () {
 program.Command.prototype.withPushOptions = function () {
   return this
     .option('--push [network]', 'push all changes to the specified network')
+    .option('--skip-compile', 'skips contract compilation')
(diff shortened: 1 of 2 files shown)


packages/cli/src/bin/options.js
```diff
@@ -16,6 +16,7 @@ program.Command.prototype.withNetworkOptions = function () {
 program.Command.prototype.withPushOptions = function () {
   return this
     .option('--push [network]', 'push all changes to the specified network')
+    .option('--skip-compile', 'skips contract compilation')
```
spalladino: Note that `add` already has a skip-compile option, and adding it here as well caused the option to be listed twice on `add --help`. We should remove it from the add command.

spalladino:
> LGTM, but shouldn't we remove it from push too?

Nope, `push` uses `withNetworkOptions`, not `withPushOptions`, so we're good.

PR #215, opened 2018-10-10
Added a storage gap at the end of Initializable's storage layout.

nventuro:
Created from the discussion in https://github.com/OpenZeppelin/openzeppelin-zos/pull/29

This is the same gap all `openzeppelin-zos` contracts have. My editor also removed some trailing whitespace, but I can undo that if you have strong feelings about it :stuck_out_tongue:

### DIFF
--- a/packages/lib/test/contracts/upgradeability/AdminUpgradeabilityProxy.test.js
+++ b/packages/lib/test/contracts/upgradeability/AdminUpgradeabilityProxy.test.js
@@ -130,8 +130,12 @@ contract('AdminUpgradeabilityProxy', ([_, admin, anotherAccount]) => {
           })
 
           it('uses the storage of the proxy', async function () {
-            // fetch the x value of Migratable at position 0 of the storage
-            const storedValue = await Proxy.at(this.proxyAddress).getStorageAt(1);
+            // storage layout should look as follows:
+            //  - 0: Initializable storage
+            //  - 1-50: Initailizable reserved storage (50 slots)
+            //  - 51: initializerRan
+            //  - 52: x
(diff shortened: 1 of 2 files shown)


spalladino:
@nventuro it seems there was a test that was dependent on storage positions. Could you fix them?

nventuro:
Done! I also improved the wording of the comment, since it was very outdated.

packages/lib/test/contracts/upgradeability/AdminUpgradeabilityProxy.test.js
```diff
@@ -130,8 +130,12 @@ contract('AdminUpgradeabilityProxy', ([_, admin, anotherAccount]) => {
           })
 
           it('uses the storage of the proxy', async function () {
-            // fetch the x value of Migratable at position 0 of the storage
-            const storedValue = await Proxy.at(this.proxyAddress).getStorageAt(1);
+            // storage layout should look as follows:
+            //  - 0: Initializable storage
+            //  - 1-50: Initailizable reserved storage (50 slots)
+            //  - 51: initializerRan
+            //  - 52: x
```
spalladino: Shouldn't it be...?

```
//  - 0: Initializable#initialized
//  - 1: Initializable#initializing
//  - 2-51: Initailizable#gap
//  - 52: x
```

nventuro: No, both `initialized` and `initializing` are `bool`, so they are packed together into the same storage slot. The test uses `InitiazableMock`, which has `bool initializerRan` and `uint256 x`, those are not packed due to `x` using the full slot.

PR #217, opened 2018-10-10
Removed dependency names can no longer be reused.

### LINKED ISSUE #202: Fix vouching removals to disallow reusing dependency names
### [zOS2V-M01]
The `Vouching` contract keeps a registry of dependencies identified by names. If a dependency is deleted, anyone can add malicious package with the same dependency name. A developer using that dependency may indarvertly use the malicious package.

nventuro:
Fixes #202.

I considered leaving a sentinel value to signal a name had been taken, but both `owner` and `address` need to be obviously zeroed out, and `stake` can be 0 (if `minimalStake` is 0), so I opted for adding an extra field. This should have a minimal impact on gas costs, since `used` is packed alongside `owner`, emitting a single `SSTORE`/`SLOAD` opcode when accessing (this was verified experimentally).

I'm not 100% sure on the name (maybe `nameTaken`?), but this is an internal detail and not part of the API.

### DIFF
--- a/packages/vouching/contracts/Vouching.sol
+++ b/packages/vouching/contracts/Vouching.sol
@@ -23,13 +23,13 @@ contract Vouching is Initializable {
   using SafeERC20 for ERC20;
 
   struct Dependency {
-    bool used;
     address owner;
     address dependencyAddress;
     uint256 stake;
   }
 
   mapping (string => Dependency) private _registry;
+  mapping (string => bool) private _takenDependencyNames;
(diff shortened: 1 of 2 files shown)


facuspagnuolo:
An alternative could be to have a separate mapping to tell whether a given dependency name was used or not: `mapping (string => bool) dependencyUsed`
Then we can full delete the Dependency struct and check for that mapping when registering new ones. What do you think?

Issue #205, opened 2018-10-09
Address Nomic comments and recommendations to lib contracts


facuspagnuolo:
### [zOS2L-O01]
- [ ] Change `AdminUpgradeabilityProxy#upgradeToAndCall` to work with a delegate call instead of a regular call. 

### [zOS2L-O02] & [zOS2L-O03]
- [ ] Remove `Migratable` contract and related tests

### [zOS2L-O04] 
- [ ] Do not index `App`’s `ProxyCreated` event parameter

PR #219, opened 2018-10-10
upgradeToAndCall now delegatecalls instead of externally calling.

nventuro:
Fixes issue [zOS2L-O01] of the Nomic audit (part of #205).

There were no tests for this, but I'm not sure what they'd look like, since we'd be testing the existence of the transparent proxy mechanism: maybe have the call query `admin()` and check that it reverts if the proxy is not its own admin, and doesn't otherwise?


alcuadrado:
One way of testing this is to make a proxy its own owner and upgrade itself with `upgradeToAndCall`. If an external call is used that will revert.

Issue #94, opened 2018-09-14
Update zos json files with new CLI lightweight mode


facuspagnuolo:
After improving the CLI to work in a lightweight mode by default, the serialization of every project in the `zos.<network>.json` and `zos.json` files may change. We should update its documentation after.

spalladino:
Note that there is already a section in the docs on this (covered [here](https://github.com/zeppelinos/zos-cli/issues/36)). We may want to review it rather than writing one from scratch.

facuspagnuolo:
True, I'll add more details to the description, sorry for that

ajsantander:
Addressed by #222

PR #222, opened 2018-10-10
Updates documentation about json files regarding lightmode.

ajsantander:
Addresses #94 *Update zos json files with new CLI lightweight mode*


packages/docs/docs/docs/advanced.md
```diff
@@ -227,6 +229,8 @@ ZeppelinOS will also generate a file for each of the networks you work on (`loca
...
+Another thing to notice in these files are the version numbers (or names!). The `<appVersion>` keeps track of the latest app version, and matches `<version>` from `zos.json`. The `<proxy-i-version>`s, on the other hand, keep track of which version of the contracts the proxies are pointing to. Say you deploy a contract in your app version 1.0, and then bump the version to 1.1 and push some upgraded code for that same contract. This will be reflected in the `<contract-i-address>`, but not yet in the proxy, which will display 1.0 in `<proxy-i-version>` and the old logic contract address in `<implementation-i-address>`. Once you run `zos update` to your contract, `<proxy-i-version>` will show the new 1.1 version, and `<implementation-i-address>` will point to the new `<contract-i-address>`.
```
elopio: Above you removed 'version name'. In here you can remove '(or names!)'.

packages/docs/docs/docs/advanced.md
```diff
@@ -227,6 +229,8 @@ ZeppelinOS will also generate a file for each of the networks you work on (`loca
...
+
+Also, notice the fields `<app>` and `<package>`. These contain the addresses of contracts that ZeppelinOS uses to facilitate the creation of proxies and the management of different versions of your contracts. Unless ZeppelinOS is used in "light" mode (using the `--light` option when running the `zos init` command), it will make use of these two contracts. An `App` contains a `Package`, which keeps track of versioned `ImplementationDirectory`'s using semantic versioning. An `ImplementationDirectory` simply provides different implementations for each of your app's contracts. Note that both `<app>` and `<package>` will not exist if the ZeppelinOS is initialized in light mode.
```
elopio: This looks good to me. But, there are a few things said today that confuse me.
- @palla said lightweight will be the default, so the order of the paragraph should be reverted. First mention what lightweight uses, and then mention what full app uses.
- @maraoz said we shouldn't mention "lightweight" to our users for devcon. I'm not sure if that is solved by mentioning a "default mode" and a "full mode"

I'm adding Palla as a reviewer here, because I don't have enough info to approve/reject.

spalladino: @elopio @ajsantander indeed, lightweight mode will now be the default **for apps** (libs cannot be lightweight). App, Package, and Provider will be deployed only if the user chooses to do so when running a `push --full`, or on the first push if the user had initialized the app as `init --full`.

ajsantander: Reworded a bit, pls review again @elopio .

packages/docs/docs/docs/advanced.md
```diff
@@ -227,6 +229,8 @@ ZeppelinOS will also generate a file for each of the networks you work on (`loca
...
+
+Also, notice the fields `<app>` and `<package>`. These contain the addresses of contracts that ZeppelinOS uses to facilitate the creation of proxies and the management of different versions of your contracts. These contracts will only be deployed when ZeppelinOS is used in "full" mode (using the `--full` option when running the `zos init` or `zos push` commands). That is, your project will not have an app or package unless explicitly using the `--full` option. Note that this mode is required for projects that produce an EVM package. For more info on the architecture of the "full" mode, see the [Contract Architecture](https://docs.zeppelinos.org/docs/architecture.html) section.
```
facuspagnuolo: Note that we don't have the `full` flag anymore, we implemented the `publish` command instead.
I would say something like: 
```
Also, notice [...]. These contracts will only be deployed once you publish your project to a desired 
network. That is, your project will not have an `app` or `package` unless explicitly running the 
`publish` command. Note that this step is required for projects that produce an EVM package. 
To read more about the architecture of contracts we are using to publish your project on-chain 
please refer to the [Contract Architecture](https://docs.zeppelinos.org/docs/architecture.html) section.
```

Note that I'm talking about a "step" to refer the moment you publish your code, instead of "mode" to avoid the user from thinking that ZeppelinOS can work in different modes

Issue #203, opened 2018-10-09
Address Nomic comments and recommendations to kernel contracts


facuspagnuolo:
### [zOS2V-O01]
- [x] Improve ZEPToken-TPL integration tests to ensure that if a recipient's attribute is revoked they cannot receive tokens but can still transfer to other allowed recipients. 
- [x] Add brief inline documentation telling how TPL works for the ZEPToken

### [zOS2V-O02]
- [x]  Update [ZEPToken inline documentation](https://github.com/zeppelinos/zos/blob/b2d3987d32ddf6c5f3291c960d5797291d60c6aa/packages/vouching/contracts/ZEPToken.sol#L10)

### [zOS2V-O03] 
- [x] Use `AddressUtils#isContract` in `Vouching#create`

### [zOS2V-O04] 
- [x] Fix [unvouch test](https://github.com/zeppelinos/zos/blob/c5019dd48153e7dec681a224bc8d328446eafd27/packages/vouching/test/contracts/Vouching.test.js#L220) to work with `stakeAmount.plus(1)` instead.

nventuro:
Closed via #225, #227 and #236.

PR #225, opened 2018-10-10
Vouching.create now requires the dependency address to be a contract.

nventuro:
Fixes issue [zOS2V-O03] of the Nomic audit (part of #203).


facuspagnuolo:
@nventuro failing vouching test, please review

nventuro:
Ha, that's what I get for not running tests after merging for conflicts.

PR #226, opened 2018-10-10
Renamed "standard libraries" to "EVM packages"

ajsantander:
Addresses https://github.com/zeppelinos/zos-docs/issues/93


packages/docs/docs/docs/using.md
```diff
@@ -1,18 +1,18 @@
 ---
 id: using
-title: Using the stdlib in your app
-sidebar_label: Using the stdlib
+title: Using the EVM package in your app
```
elopio: What about "Using EVM packages in your app"?

ajsantander: Good catch. Something I missed when I systematically searched and replaced. I fixed this one and looked for other situations where the same error could have occurred and actually found a few.

facuspagnuolo:
@elopio we are tackling pending renames on https://github.com/zeppelinos/zos/pull/240

PR #229, opened 2018-10-10
Enhance cli outputs on error

jcarpanelli:
Fixes  https://github.com/zeppelinos/zos-cli/issues/337.

- Changed `successfully written zos.network.json` to a less _possitive-everything-is-fine-wohoo!_ message (`Created zos.network.json` or `Updated zos.network.json`).
- Only writing `zos.network.json` file if needed, whence only showing this message when the file is written.
- Moved the error message right next to the action that caused it.

### DIFF
--- a/packages/cli/src/scripts/create.js
+++ b/packages/cli/src/scripts/create.js
@@ -1,8 +1,6 @@
 import stdout from '../utils/stdout';
 import ControllerFor from '../models/network/ControllerFor';
-import { Logger } from 'zos-lib'
-
-const log = new Logger('scripts/create')
+import ScriptError from '../models/errors/ScriptError'
 
 export default async function createProxy({ packageName, contractAlias, initMethod, initArgs, network, txParams = {}, force = false, networkFile = undefined }) {
   if (!contractAlias) throw Error('A contract alias must be provided to create a new proxy.')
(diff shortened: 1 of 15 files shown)


packages/cli/src/scripts/create.js
```diff
@@ -1,16 +1,25 @@
...
-  await controller.checkContractDeployed(packageName, contractAlias, !force);
-  const proxy = await controller.createProxy(packageName, contractAlias, initMethod, initArgs);
+  try {
+    await controller.checkContractDeployed(packageName, contractAlias, !force);
+    const proxy = await controller.createProxy(packageName, contractAlias, initMethod, initArgs);
+    stdout(proxy.address);
+
+    return proxy;
+  } catch(error) {
+    log.error(error.message)
```
facuspagnuolo: shouldn't we rethrow here? otherwise, we will be swallowing that error, for example, we are using this scripts programmatically in other projects and we expect them to fail if sth went wrong, we should write and rethrow I think

jcarpanelli:
I've implemented a new `ScriptError` class, that inherits from `Error`. This is not very generic though, but with some small changes (sending the second parameter of `ScriptError` constructor as an object), we can add other functionalities that we may want to execute after an error is thrown (e.g., writing a `zos.error.log`).

PR #230, opened 2018-10-10
Add new check command

spalladino:
Adds a new check command that outputs validation errors for all contracts in the project, or allows for a single contract to be specified.

Depends on pull request from branch feature/validate-contract-on-push. Merge after that one.


spalladino:
No need to remove, check does not call `withPushOptions`

PR #240, opened 2018-10-12
Stdlib/lib/dependency rename to package

jcarpanelli:
Renamed stdlib/lib/dependency to package in messages, labels, descriptions and json attributes.

Fixes https://github.com/zeppelinos/zos/issues/238.
Related to https://github.com/zeppelinos/zos/issues/272.


spalladino:
Awesome work Jota! I didn't even know there were still references to stdlib around the codebase.

The only thing I'm not sold on is removing the term `dependencies`. In many cases I think it makes sense to keep them, especially since using `packages` arbitrarily is confusing with the project package itself. For example, the `zos.network.json` file will now have a `package` entry (for the project package) and a `packages` one (for the dependencies). Also, using `package: true` and `isPackage: true` in the JSON files feels awkward, especially having a `package: { address: '0x...' }` entry as well.

As for the commands, I'd try alternating between dependencies and packages depending on the context, for instance: `--deploy-packages` => `--deploy-dependencies`.

@elopio or @maraoz could you chime in? I don't want to change anything you guys have already agreed upon regarding naming, but I'd appreciate if you could review it.

maraoz:
Agree 100% with @spalladino on the "dependencies" vs "packages" comment. Actually I had this concern in my mind when we decided to rename "libraries" to "packages", happy to see it addressed.

jcarpanelli:
Note: I squashed my previous commits to avoid dealing with conflicts in each commit while rebasing, sorry for that.
Please let me know if I'm missing something.

packages/cli/src/commands/link.js
```diff
@@ -4,14 +4,14 @@ import push from './push'
 import linkStdlib from '../scripts/link'
 
 const name = 'link'
-const signature = `${name} [libraries...]`
-const description = 'links project with a list of libraries located in each library npm package'
+const signature = `${name} [dependencies...]`
+const description = 'links project with a list of dependencies located in its npm package'
```
spalladino:
```suggestion
const description = 'links project with a list of dependencies each located in its npm package'
```

spalladino: Look ma, I'm using Github suggestions!!

packages/cli/src/commands/freeze.js
```diff
@@ -5,7 +5,7 @@ import runWithTruffle from '../utils/runWithTruffle'
 
 const name = 'freeze'
 const signature = name
-const description = 'freeze current release version of your stdlib project'
+const description = 'freeze current release version of your package project'
```
spalladino:
```suggestion
const description = 'freeze current release version of your published project'
```

Now we can freeze any kind of project, as long as its published.

PR #241, opened 2018-10-12
Detect initial values in fields declarations

jcarpanelli:
Fixes https://github.com/zeppelinos/zos/issues/80.

### DIFF
--- a/packages/lib/src/validations/InitialValues.js
+++ b/packages/lib/src/validations/InitialValues.js
@@ -1,0 +1,28 @@
+'use strict'
+
+import _ from 'lodash'
+import Contracts from "../utils/Contracts"
+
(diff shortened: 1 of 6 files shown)


packages/lib/src/validations/InitialValues.js
```diff
@@ -0,0 +1,28 @@
...
+
+import _ from 'lodash'
+import Contracts from "../utils/Contracts"
+
+export function hasInitialValuesInDeclarations(contractClass) {
+  return detectInitialValues(contractClass)
+}
+
+function detectInitialValues(contractClass) {
+  const nodes = contractClass.ast.nodes.filter((n) => n.name === contractClass.contractName)
```
spalladino: Can be a find instead of filter, as there should be only one.

PR #253, opened 2018-10-14
Drop libs distinction on init

### LINKED ISSUE #207: Consider removing difference between app and lib until publishing
Today a zOS user needs to decide up front whether he'll be using zOS to create a shared lib or an app, and that decision cannot be changed later. 

We can leverage lightweight mode to allow for postponing this decision: initially, all development is in lightweight mode, but we can give our users the chance to `publish` their work as an app or lib (or both) later down the road, based on whether they want their work to be reusable by other teams.

spalladino:
Fixes  #207

### DIFF
--- a/packages/cli/src/commands/init.js
+++ b/packages/cli/src/commands/init.js
@@ -2,7 +2,6 @@
 
 import push from './push'
 import init from '../scripts/init'
-import initLib from '../scripts/init-lib'
 
 const name = 'init'
 const signature = `${name} <project-name> [version]`
@@ -12,8 +11,7 @@ const register = program => program
   .command(signature, { noHelp: true })
   .usage('<project-name> [version]')
   .description(description)
-  .option('--full', 'create an application, package, and implementation directory contracts to manage your project (always true for libs)')
-  .option('--lib', 'create a standard library instead of an application')
+  .option('--full', 'create an application, package, and implementation directory contracts to manage your project')
   .option('--force', 'overwrite existing project if there is one')
   .option('--link <stdlib>', 'link to a standard library')
   .option('--no-install', 'skip installing stdlib dependencies locally')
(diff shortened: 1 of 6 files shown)


facuspagnuolo:
Could we consider this PR fixes https://github.com/zeppelinos/zos/issues/207?

packages/cli/src/commands/init.js
```diff
@@ -23,13 +21,8 @@ const register = program => program
...
 
-  if (options.lib) {
-    if (link) throw Error('Cannot set a stdlib in a library project')
-    await initLib({ name, version, force })
-  } else {
-    const libs = link ? link.split(',') : []
-    await init({ name, version, libs, installLibs, force, full })
-  }
+  const libs = link ? link.split(',') : []
+  await init({ name, version, libs, installLibs, force, full })
```
facuspagnuolo: 👏

packages/cli/src/models/network/NetworkAppController.js
```diff
@@ -257,4 +257,10 @@ export default class NetworkAppController extends NetworkBaseController {
       fs.writeJson(path, data)
     }
   }
+
+  async freeze() {
+    await this.fetchOrDeploy(this.currentVersion)
+    await this.project.freeze()
+    this.networkFile.frozen = true
+  }
```
facuspagnuolo: ❄️

packages/cli/src/models/network/NetworkBaseController.js
```diff
@@ -86,6 +86,7 @@ export default class NetworkBaseController {
     if (this._newVersionRequired()) {
       log.info(`Current version ${this.currentVersion}`);
       log.info(`Creating new version ${this.packageVersion}`);
+      this.networkFile.frozen = false;
```
facuspagnuolo: we can remove the overload from `NetworkLibController` then, I know that we are not removing lib code in this PR tho

Issue #249, opened 2018-10-12
Error when checking for uninitialized base contracts


spalladino:
The uninitialized base contract check (see #52) is failing even on simple cases with an `Cannot read property 'name' of undefined` error . Either identify the cause of the error and fix it, or remove the check altogether (drop L17 from `lib/src/validations/index.js`).

Destiner:
Hi there! Want to try to tackle this, but can't launch zos from source.

I'm executing `node zos/packages/cli/src/bin/zos-cli.js`, but getting 
```
import { Logger } from 'zos-lib'
       ^

SyntaxError: Unexpected token {
```

spalladino:
Removing feature altogether in #258 and moving this bug to backlog.

Destiner:
Can you please provide an example that fails? I pushed some contracts and they seem to work fine. I ensured that "uninit base contract" check is running.

spalladino:
Closing due to #1300

PR #258, opened 2018-10-15
Temporarily remove validation of uninit base contracts

spalladino:
Re-enable once #249 is fixed.

### DIFF
--- a/packages/lib/src/validations/index.js
+++ b/packages/lib/src/validations/index.js
@@ -10,11 +10,9 @@ import { hasInitialValuesInDeclarations } from './InitialValues';
 
 const log = new Logger('validate')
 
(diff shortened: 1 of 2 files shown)


packages/lib/src/validations/index.js
```diff
@@ -84,7 +82,7 @@ function tryGetUninitializedBaseContracts(contractClass) {
   try {
     return _(getUninitializedBaseContracts(contractClass)).values().flatten().uniq().value()
   } catch (error) {
-    log.error(`Skipping uninitialized base contracts validation due to error: ${error.message}`)
+    log.error(`- Skipping uninitialized base contracts validation due to error: ${error.message}`)
```
facuspagnuolo: This line will be never executed, however, I would use a `warn` level here

PR #276, opened 2018-10-17
Clean the main README

### LINKED ISSUE #232: Update readmes for 2.0 release

elopio:
Fixes #232

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,78 +1,83 @@
-# ZeppelinOS
+# ZeppelinOS _(zos)_
+
+[![standard-readme compliant](https://img.shields.io/badge/readme%20style-standard-brightgreen.svg)](https://github.com/RichardLitt/standard-readme)
 [![Build Status](https://travis-ci.org/zeppelinos/zos.svg?branch=master)](https://travis-ci.org/zeppelinos/zos)
 [![lerna](https://img.shields.io/badge/maintained%20with-lerna-cc00ff.svg)](https://lernajs.io/)
 
-Multi-package repository for ZeppelinOS [library](https://github.com/zeppelinos/zos/tree/master/packages/lib#readme), [CLI](https://github.com/zeppelinos/zos/tree/master/packages/cli#readme), [vouching](https://github.com/zeppelinos/zos/tree/master/packages/vouching#readme), and [docs](https://github.com/zeppelinos/zos/tree/master/packages/docs#readme).
+> ZeppelinOS smart contract platform.
 
-ZeppelinOS is a platform to develop, manage and operate smart contract applications in Ethereum. It can be used to create smart contract systems that can be fixed and improved over time, enabling developers to opt-in to mutability for their deployed code through [upgradeability patterns](https://blog.zeppelinos.org/proxy-patterns/).
+ZeppelinOS is a platform to develop, deploy and operate smart contract
+projects on Ethereum and every other EVM and eWASM-powered blockchain.
 
-## Getting started
+This repository includes the ZeppelinOS
+[Command-Line Interface](https://github.com/zeppelinos/zos/tree/master/packages/cli#readme),
+[JavaScript Library](https://github.com/zeppelinos/zos/tree/master/packages/lib#readme),
+[Vouching Smart Contracts](https://github.com/zeppelinos/zos/tree/master/packages/vouching#readme),
+and [the sources for the Documentation Website](https://github.com/zeppelinos/zos/tree/master/packages/docs#readme).
 
-Install ZeppelinOS and setup your project as described in the [Setup guide](https://docs.zeppelinos.org/docs/setup.html).
+## Install
 
-In order to build an upgradeable application with ZeppelinOS, follow our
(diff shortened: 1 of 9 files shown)


README.md
```diff
@@ -1,78 +1,83 @@
-# ZeppelinOS
+# ZeppelinOS _(zos)_
```
facuspagnuolo: I like the idea of clarifying this, but it looks weird in markdown. Maybe using a new line for `(zos)`?

README.md
```diff
@@ -1,78 +1,83 @@
...
-## Development setup
+* [@facuspagnuolo](https://github.com/facuspagnuolo/)
+* [@spalladino](https://github.com/spalladino)
 
-This monorepo is managed via [lerna](https://lernajs.io/). To get everything up and running, install lerna 3.x (either globally or locally from the root of the repository), and run `lerna bootstrap`. This will compile all packages, install all dependencies, and create symlinks for local dependencies (for instance, `packages/cli/node_modules/zos-lib` will be symlinked to `packages/lib`).
+## Contribute
 
-Note that, since `cli` and `lib` packages need to be precompiled, for any change in `lib` to impact `cli` you'll need to run a new `lerna bootstrap`.
+To contribute, join our
+[community channel on Telegram](https://t.me/zeppelinos) where you can talk to
```
facuspagnuolo: please validate with @martriay to make sure we are not missing any other channel

TERMINOLOGY.md
```diff
@@ -1,5 +1,5 @@
 * ZeppelinOS (zOS for short): a platform to develop, deploy and operate smart
-  contract applications on Ethereum and every other EVM and eWASM-powered
+  contract projects on Ethereum and every other EVM and eWASM-powered
```
facuspagnuolo: 👍

packages/docs/README.md
```diff
@@ -1,26 +1,60 @@
...
-`git clone https://github.com/zeppelinos/zos-docs.git`
+## Development setup
+
+First, install [Node.js](http://nodejs.org/), [npm](https://npmjs.com/) and
+[git](https://git-scm.com/).
+
+Then, clone the repository:
+
+```sh
+git clone https://github.com/zeppelinos/zos-docs.git
```
facuspagnuolo: shouldn't we ask the user to clone the main `zos` repo and navigate to `packages/docs` instead?
```
git clone https://github.com/zeppelinos/zos.git
cd ./packages/docs
```

packages/docs/README.md
```diff
@@ -1,26 +1,60 @@
...
+```
 
 Install Docusaurus:
 
-`npm install -g docusaurus-init`
+```sh
+npm install --global docusaurus-init
+```
 
 Navigate to the `docs/website` directory and run:
```
facuspagnuolo: I think this is `docs/docs/website`

packages/docs/README.md
```diff
@@ -1,26 +1,60 @@
...
 
-`npm install`
+```sh
```
facuspagnuolo: we can add `cd /docs/docs/website` here to be explicit

packages/lib/README.md
```diff
@@ -1,50 +1,79 @@
...
-  - [Examples](#examples)
-- [Links](#links)
-- [Security](#security)
-- [License](#license)
+This is the repository for the ZeppelinOS JavaScript library. It is mainly used
+by the
+[`zos` command-line interface](https://github.com/zeppelinos/zos/tree/master/packages/cli#zeppelinos-command-line-interface),
+which is the recommended way to use ZeppelinOS; but this library can also be
+used directly to operate ZeppelinOS projects when a programmatic interface is
+preferred or more flexibility and lower-level access is required.
```
facuspagnuolo: Loved this!

facuspagnuolo:
The only one missing would be to work on the vouching readme as well, I don't if you preferred to leave that one for later

README.md
elopio: This is the recommended style of standard readme:
https://github.com/RichardLitt/standard-readme/blob/master/spec.md#title
I would prefer to follow it even in small details. My argument is that if we find a way to improve it, we should propose it upstream for everybody to benefit. If it's not worth it, we will be just discussing about subjective preferences, which is what I want to avoid following the standard.

packages/docs/README.md
elopio: :thinking: it depends what is the reference directory.
from the root of the repo, this is packageds/docs/docs/website. But from the directory where this README is located, it's docs/website.
This is very confusing :D any preference?

packages/docs/README.md
elopio: ah, yes, good idea.

README.md
facuspagnuolo: Makes sense!

packages/docs/README.md
facuspagnuolo: @elopio IMO this step needs to be fixed since we are not using that repo anymore

packages/docs/README.md
facuspagnuolo: Note that I suggested navigating to the package above, so if we apply that change this should be just /docs/website I think

Issue #96, opened 2018-09-17
Update guide to onboard ERC20 tokens


facuspagnuolo:
Once OZ 2.0 is released, we should update the [onboarding guide](https://docs.zeppelinos.org/docs/erc20_onboarding.html) for ERC20 tokens using the new [ERC20Migrator]( https://github.com/OpenZeppelin/openzeppelin-solidity/blob/release-v2.0.0/contracts/drafts/ERC20Migrator.sol) implementation. 

We should also decide whether [this example repo](https://github.com/zeppelinos/erc20-opt-in-onboarding) is still necessary or if we can rely only on the documentation guide.

EDIT: @elopio's comment on a duplicated issue
> The page https://docs.zeppelinos.org/docs/erc20_onboarding.html needs to be updated to use the most recent version of openzeppelin-eth.
> Also it refers to some outdated things from zos, like --deploy-libs.

PR #277, opened 2018-10-17
docs: remove basil and crafty demos

elopio:
The demos are outdated. For 2.0, they should probably be replaced with
simpler examples.

Fixes: #96

[CI skip]


elopio:
Yes, I'm not deleting the files for the old versions, like: https://github.com/zeppelinos/zos/blob/master/packages/docs/docs/website/versioned_docs/version-1.0.0/basil.md
This will only apply when generating the new tag.

PR #288, opened 2018-10-17
Adds a new section to the documentation on "proxy patterns"

### LINKED ISSUE #151: Add upgradeability pattern documentation
We've written a lot about upgradeability patterns in our blog but there is no consolidated documentation in our doc site. We should add a section explaining every detail of the pattern we decided to use for ZeppelinOS. We can build something based on our [proxy patterns post](https://blog.zeppelinos.org/proxy-patterns/)

ajsantander:
(WIP)

Fixes #151


packages/docs/docs/docs/proxies.md
```diff
@@ -0,0 +1,68 @@
+---
+id: proxies
+title: Proxy Pattern
```
elopio: I would like to brand this:
The ZeppelinOS Upgrade Pattern

The problem here is that the ideas for this come from many different places, so maybe at the end we can add some generic thanks to everybody researching about upgradeability in ethereum.

@martriay what do you think?

packages/docs/docs/docs/proxies.md
```diff
@@ -0,0 +1,68 @@
...
+---
+
+This article describes the "unstructured storage" proxy pattern, the fundamental building block of ZeppelinOS's contract upgradeability feature.
```
elopio: ZeppelinOS's contract upgradeability feature -> ZeppelinOS upgrades.

packages/docs/docs/docs/proxies.md
```diff
@@ -0,0 +1,68 @@
...
+
+Note: For a more in depth read, please see [blog.zeppelinos.org/proxy-patterns](https://blog.zeppelinos.org/proxy-patterns/), which discusses the need for proxies, goes into more technical detail on the subject, elaborates on other possible proxy patterns that were considered for zOS, and more.
+
+## Why upgradeability?
+
+By design, smart contracts are immutable. On the other hand, software quality heavily depends on the ability to upgrade and patch source code in order to make iterative releases. Even though blockchain based software profits signifficantly from the technology's immutability, still a certain degree of mutability is needed for bugfixing and potential product improvements. ZeppelinOS solves this apparent contradiction by providing an easy to use, simple and robust upgrade mechanism for smart contracts that can be controlled by any type of governance, be it a multi-sig wallet, a simple address or a complex DAO.
```
elopio: We could mention here the word "opt-in". Maybe, "simple, robust and opt-in mechanism..."

packages/docs/docs/docs/proxies.md
```diff
@@ -0,0 +1,68 @@
...
+
+## Upgradeability via the proxy pattern
+
+The basic idea behind using the proxy pattern to solve the problem of pugradeability, is to use two contracts instead of one. The first contract is a simple wrapper or "proxy" which users interact with directly and is in charge of forwarding transactions to and from the second contract, which contains the logic of the pair. The key concept to understand is that the logic contract can be replaced while the proxy is never changed. Both contracts are still immutable, in the sense that their code cannot be changed, but the logic contract can simply be swapped by another. The wrapper can thus point to a different logic implementation and in doing so, the software is "upgraded".
```
elopio: to solve the problem of pugradeability -> for upgrades

elopio: I like this explanation a lot!

packages/docs/docs/docs/proxies.md
```diff
@@ -0,0 +1,68 @@
...
+
+## Proxy forwarding
+
+The most immediate problem that proxies need to solve is how the proxy exposes the entire interface of the logic contract without requiring a one to one mapping of the entire logic contract's interface. This would be difficult to maintain, prone to errors, and would make the interface itself not upgradeable. Hence, a dynamic forwarding mechanism is required. The basics of such mechanism can be summarized in the code below:
```
elopio: This -> That

packages/docs/docs/docs/proxies.md
```diff
@@ -0,0 +1,68 @@
...
+assembly {
+    let ptr := mload(0x40)
+    calldatacopy(ptr, 0, calldatasize)
+    let result := delegatecall(gas, _impl, ptr, calldatasize, 0, 0)
+    let size := returndatasize
+    returndatacopy(ptr, 0, size)
+
+    switch result
+    case 0 { revert(ptr, size) }
+    default { return(ptr, size) }
```
elopio: I prefer when complex code is commented inline, instead than at the bottom.
However, that could affect the readability of the paragraph, I'm not sure here.
At least, it would be good to comment in line with (1), (2), (3) and (4) to make it easier to understand what parts of the code you are talking about.

packages/docs/docs/docs/proxies.md
```diff
@@ -0,0 +1,68 @@
...
+ }
+ ```
+
+This code can be put in the fallback function of a proxy, and will forward any call to any function with any set of parameters to the logic contract without it needing to know anything in particular of the logic contract's interface. In essence, (1) the `calldata` is copied to memory, (2) the call is forwarded to the logic contract, (3) the return data from the call to the logic contract is retrieved, and (4) the returned data is forwarded back to the caller.
```
elopio: link to the solidity docs of fallback function.
It would be good to explain why this is assembly and not solidity code.

packages/docs/docs/docs/proxies.md
```diff
@@ -0,0 +1,68 @@
...
+
+A very important thing to note is that the code makes use of `delegatecall` which executes the callee's code in the context of the caller's state. That is, the logic contract controls the proxy's state and the logic contract's state is meaningless. Thus, the proxy doesn't forward transactions to and from the logic contract, but also represents the pair's state. The state is in the proxy and the logic is in the particular implementation that the proxy points to. To put it in yet another way, the "what" is in the proxy contract and the "how" is in the logic contract.
```
elopio: Link to delegatecall's docs.
doesn't forward -> doesn't only forward

elopio: I'm not sure I like the last sentence about "what" and "how". What would you think of mentioning instead that the state variables are in the proxy, and the functions are in the logic?

packages/docs/docs/docs/proxies.md
```diff
@@ -0,0 +1,68 @@
...
+
+As a result, a logic contract doesn't need to care about overwritting any of the proxy's variables.
+
+TODO: add diagrams 
+
+## Storage collisions
+
+Despite the unstructured storage's approach on avoiding collisions between the logic contract's variables and the proxy's variables, storage collisions between different versions of the logic contracts' variables can occur. In this case, imagine that the first implementation of the logic contract stores `address public _owner` at the first storage slot and a logic contract that replaces it on an upgrade stores `address public _lastContributor` at the same first slot. When the updated logic attempts to write to `_lastContributor` it will be using the same storage position where the previous value for `_owner` was being stored and hence overwrite it.
+
+The unstructured storage proxy mechanism doesn't safeguard against this. It is up to the user to have new versions of a logic contract extend previous versions, or otherwise guarantee that the storage hirarchy is always appended to but not changed. ZeppelinOS' CLI does however detect such collisions, and warns the developer about them.
```
elopio: hirarchy -> hierarchy

ajsantander:
@elopio In this last commit I removed a few sections from advanced.md and merged them into proxies.md. My criteria was "incorporate what the items in advanced.md describe that is not described in the items in proxies.md and then delete them."

packages/docs/docs/docs/proxies.md
ajsantander: Keep in mind that the title has to fit in the sidebar =D

packages/docs/docs/docs/proxies.md
ajsantander: I don't like that because it may confuse the user. One may think that variables need to be declared in the proxy in <code> and functions in the implementation.

Removing my sentence and replacing it with nothing for now.

packages/docs/docs/docs/proxies.md
```diff
@@ -0,0 +1,150 @@
...
+|...                       |...                      |
+|...                       |                         |
+|...                       |                         |
+|...                       |                         |
+|...                       |                         |
+|address _implementation   |                         | <=== Randomized slot.
+|...                       |                         |
+|...                       |                         |
+   
+An example of how the randomized storage is achieved:
```
facuspagnuolo: I wouldn't use the word random, it sounds not-robust, I'd rather use something like `fixed`, explaining that we choose a specific position [making sure that Solidity won't ever use it](https://github.com/zeppelinos/labs/issues/82)

Issue #56, opened 2018-08-29
Estimate gas price when sending transactions


spalladino:
Truffle's mainnet default is 100 GWei, which is typically 20x-100x higher than the gas price required. If the user hasn't explicitly set a gas price, use a gas price API (such as [eth gas station](https://ethgasstation.info/)) or the value from previous blocks (similar to what the [gasPrice subprovider](https://github.com/MetaMask/provider-engine/blob/master/subproviders/gasprice.js) does) to estimate a more reasonable value.

Note that newer versions of truffle/web3 **may** handle this automatically, if we upgrade.

Another possible enhancement could be to check the actual value of the tx (in either ETH or USD), and warn the user if it exceeds a certain threshold.

gitcoinbot:
Issue Status: 1. **Open** 2. Started 3. Submitted 4. Done 

<hr>

__Workers have applied to start work__.


These users each claimed they can complete the work by 12 months from now.
Please review their action plans below:


**1) [Anish-Agnihotri](https://gitcoin.co/profile/Anish-Agnihotri) has applied to start work _(Funders only: [approve worker](https://gitcoin.co/issue/zeppelinos/zos/56/1273?mutate_worker_action=approve&worker=Anish-Agnihotri) | [reject worker](https://gitcoin.co/issue/zeppelinos/zos/56/1273?mutate_worker_action=reject&worker=Anish-Agnihotri))_.**

Estimating gas pricing should be a fairly doable job using an external API or using the method that you've mentioned (comparing past transactions). Essentially, I'd setup an async query to an external API that stores the averaged Gas pricing for the last x blocks in a variable. Then, that same estimated gas pricing could be offered to the user.

If this sounds somewhat in the range of what you're looking for, I'd love to have a chat and discuss this further.

Learn more [on the Gitcoin Issue Details page](https://gitcoin.co/issue/zeppelinos/zos/56/1273).

spalladino:
@Anish-Agnihotri that sounds good! There is a `Transactions` module that centralizes all transactions and deployments we make, and is already taking care of gas estimations. I think this would be the best place to hook the gas price estimator. 

A few notes:
- If the user has supplied their own gas price via truffle.js, we should not attempt to estimate gas price automatically.
- Since most CLI operations are short-lived, we can estimate the gas price just once, and cache it on the Transactions object, without needing to invalidate it
[...]

PR #299, opened 2018-10-18
Estimate gas price

zachzundel:
So far, I have the gas price being queried from the API if the default is being used. I still need to fix the tests and maybe add a `--gasPrice` flag to the CLI. This is for issue #56 .


zachzundel:
@spalladino Could you take a look at this and let me know if it seems reasonable so far?

spalladino:
@zachzundel looks good so far! A few comments:

- Try to cache the gas price obtained from the API in the `state` object, to avoid having to wait one query per each tx.
- Double-check the units in which the API returns the gas price, you may need to add a few zeroes to the value you obtain.
- Make sure to add a few tests later down the road (I know this is still WIP).
- And a minor comment: these lines are repeated several times, see if it's possible to refactor them:
```js
   let fixedGasPrice = checkGasPrice(txParams)
   if (fixedGasPrice) {
      txParams.gasPrice = fixedGasPrice
  }
```

spalladino:
@zachzundel looks good so far! A few comments:

- Try to cache the gas price obtained from the API in the `state` object, to avoid having to wait one query per each tx.
- Double-check the units in which the API returns the gas price, you may need to add a few zeroes to the value you obtain.
- Make sure to add a few tests later down the road (I know this is still WIP).
- And a minor comment: these lines are repeated several times, see if it's possible to refactor them:
```js
   let fixedGasPrice = checkGasPrice(txParams)
   if (fixedGasPrice) {
      txParams.gasPrice = fixedGasPrice
  }
```

spalladino:
@zachzundel looks good so far! A few comments:

- Try to cache the gas price obtained from the API in the `state` object, to avoid having to wait one query per each tx.
- Double-check the units in which the API returns the gas price, you may need to add a few zeroes to the value you obtain.
- Make sure to add a few tests later down the road (I know this is still WIP).
- And a minor comment: these lines are repeated several times, see if it's possible to refactor them:
```js
   let fixedGasPrice = checkGasPrice(txParams)
   if (fixedGasPrice) {
      txParams.gasPrice = fixedGasPrice
  }
```

spalladino:
@zachzundel looks good so far! A few comments:

- Try to cache the gas price obtained from the API in the `state` object,
to avoid having to wait one query per each tx.
- Double-check the units in which the API returns the gas price, you may
need to add a few zeroes to the value you obtain.
- Make sure to add a few tests later down the road (I know this is still
WIP).
- And a minor comment: these lines are repeated several times, see if it's
possible to refactor them:
```js
   let fixedGasPrice = checkGasPrice(txParams)
   if (fixedGasPrice) {
      txParams.gasPrice = fixedGasPrice
  }
```

(sorry if you see this comment several times, I'm getting an error when
posting it)

zachzundel:
Hi @spalladino -- I've updated my PR and I think I've fixed everything you pointed out. There was already a test for estimateGas that seems to cover my use case, please let me know if that's not correct and I will write another one!

packages/lib/src/utils/Transactions.js
```diff
@@ -152,11 +156,28 @@ async function getNodeVersion () {
...
-    throw new Error(`Cowardly refusing to execute transaction with excessively high default gas price of 100 gwei. Consider explicitly setting a different gasPrice value in your truffle.js config file. You can check reasonable values for gas price in https://ethgasstation.info/.`);
+    if (state.gasPrice) {
+        txParams.gasPrice = state.gasPrice;
+        return txParams;
+    }
+
+    try {
+      let apiResponse = await axios.get(GAS_API_URL);
+      let gasPriceGwei = apiResponse.safeLow / 10;
+      let gasPrice = BN(gasPriceGwei * 1000000000)
```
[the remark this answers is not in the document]
zachzundel: Can do! Out of curiousity, can you point me to a writeup on why one should prefer const over let?

zachzundel:
How can we tell testnet vs. mainnet here? Is it something in `txParams`? Do I need to use Truffle?

packages/lib/src/utils/Transactions.js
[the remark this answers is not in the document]
spalladino: Sure! From [ESLint](https://eslint.org/docs/rules/prefer-const):

> const declaration tells readers, “this variable is never reassigned,” reducing cognitive load and improving maintainability.

zachzundel:
@spalladino I've added in tests, this is ready for review!

package.json
```diff
@@ -3,5 +3,8 @@
   "private": true,
   "devDependencies": {
     "lerna": "~3.3.2"
+  },
+  "dependencies": {
+    "8": "0.0.1"
```
spalladino: What is this dependency for? Please remove if unneeded.

packages/lib/src/utils/Transactions.js
```diff
@@ -152,12 +156,48 @@ async function getNodeVersion () {
...
+}
+
+async function fixGasPrice(txParams) {
+  const network = await getNetwork();
+
   const gasPrice = txParams.gasPrice || Contracts.artifactsDefaults().gasPrice;
-  if (TRUFFLE_DEFAULT_GAS_PRICE.eq(gasPrice) || !gasPrice) {
-    throw new Error(`Cowardly refusing to execute transaction with excessively high default gas price of 100 gwei. Consider explicitly setting a different gasPrice value in your truffle.js config file. You can check reasonable values for gas price in https://ethgasstation.info/.`);
+
+  if (TRUFFLE_DEFAULT_GAS_PRICE == gasPrice || !gasPrice) {
```
spalladino: Let's use bignumber equality functions for this, ie `TRUFFLE_DEFAULT_GAS_PRICE.eq(gasPrice)`

packages/lib/src/utils/Transactions.js
```diff
@@ -152,12 +156,48 @@ async function getNodeVersion () {
...
+    if (network != '1') {
```
spalladino: We shouldn't need to do anything if network id is not mainnet, I think we can just leave gasPrice empty and let truffle pick the default. We can just `return` if `network != 1`.

packages/lib/src/utils/Transactions.js
```diff
@@ -152,12 +156,48 @@ async function getNodeVersion () {
...
+      txParams.gasPrice = TRUFFLE_DEFAULT_GAS_PRICE.toNumber();
+      return txParams;
+    }
+
+    txParams.gasPrice = await getETHGasStationPrice()
+
+    if (txParams.gasPrice >= TRUFFLE_DEFAULT_GAS_PRICE) {
```
spalladino: Same here, let's use `TRUFFLE_DEFAULT_GAS_PRICE.lte(txParams.gasPrice)`

packages/lib/src/utils/Transactions.js
```diff
@@ -32,7 +36,7 @@ const TRUFFLE_DEFAULT_GAS_PRICE = BN(100000000000);
  * @param retries number of transaction retries
  */
 export async function sendTransaction(contractFn, args = [], txParams = {}, retries = RETRY_COUNT) {
-  await checkGasPrice(txParams);
+  txParams = await fixGasPrice(txParams)
```
spalladino: `fixGasPrice` is both modifying `txParams` and returning it, which may be confusing to users of the function. Let's either: 1- remove the return value, and just call it as `await fixGasPrice(txParams)` (without the need for assigning to txParams)
2- or have fixGasPrice return a copy of txParams with gasPrice set (without modifying the parameter, so you'd need to make a copy of the txParams parameter among the first things you do in fixGasPrice)

packages/lib/src/utils/Transactions.js
```diff
@@ -152,12 +156,48 @@ async function getNodeVersion () {
...
+  if (TRUFFLE_DEFAULT_GAS_PRICE == gasPrice || !gasPrice) {
+    if (network != '1') {
+      txParams.gasPrice = TRUFFLE_DEFAULT_GAS_PRICE.toNumber();
+      return txParams;
+    }
+
+    txParams.gasPrice = await getETHGasStationPrice()
+
+    if (txParams.gasPrice >= TRUFFLE_DEFAULT_GAS_PRICE) {
+        throw new Error(`Gas price API gave very high value (>100gwei), please manually provide a gas price.`)
```
spalladino: Let's rephrase to "The current gas price estimate from ethgasstation.info is over 100 gwei. If you do want to send a transaction with a gas price this high, please set it manually in your truffle.js configuration file."

packages/lib/src/utils/Transactions.js
```diff
@@ -152,12 +156,48 @@ async function getNodeVersion () {
...
-  if (await isGanacheNode()) return;
+async function getNetwork() {
+  if (!state.network) {
+    state.network = await promisify(web3.version.getNetwork.bind(global.web3.version))();
+  }
+  return state.network;
+}
+
+async function getETHGasStationPrice() {
+    if (state.gasPrice) return state.gasPrice;
```
spalladino: Please use two spaces for indentation, for consistency

packages/lib/test/src/utils/Transactions.test.js
```diff
@@ -67,6 +73,34 @@ contract('Transactions', function([_account1, account2]) {
       assertGasLt(tx, 1000000);
     });
 
+    it('uses gas price API when gas not specified', async function () {
+        const oldNetwork = state.network
```
spalladino: Please use two spaces for indentation, for consistency

packages/lib/test/src/utils/Transactions.test.js
```diff
@@ -67,6 +73,34 @@ contract('Transactions', function([_account1, account2]) {
...
+
+        state.network = "1";
+        const stub = sinon.stub(axios, 'get').resolves({ average: 49 })
```
spalladino: Let's move these two lines to a beforeEach block. If needed, we can wrap these two tests in a "describe" block, so the beforeEach only affects them.

packages/lib/test/src/utils/Transactions.test.js
```diff
@@ -67,6 +73,34 @@ contract('Transactions', function([_account1, account2]) {
...
+
+        const { tx } = await sendTransaction(this.instance.initialize, [42, 'foo', [1,2,3]]);
+      
+        assertGasPrice(tx, 49 * 1e8);
+
+        stub.restore()
+        state.network = oldNetwork
```
spalladino: Same here but with an afterEach

zachzundel:
I've pushed commits to address all your changes!

PR #304, opened 2018-10-19
docs: avoid the truffle gas limit check using the test env

elopio:
I also added one parragraph to explain what's in the docs.

[CI skip]


elopio:
NODE_ENV=test is to workaround https://github.com/zeppelinos/zos/issues/303
A guide using a script to run with truffle exec, as in the examples, will be in the docs.

PR #309, opened 2018-10-22
Check package json version instead of zos version for dependencies

### LINKED ISSUE #308: Dependencies validations may be misleading
Let's say you are developing an EVM package and publish a version `1.0.0`, but then you noticed that there was a problem with something that has nothing to do with the contracts you are providing, e.g. fixing a deployment script or adding a file that was npm-ignored or so. Then, you will have to bump your `package.json` and publish again. This will make your `package.json` and `zos.json` to get unsynched, which means that people trying to link your package as `zos link your-evm-package@1.0.1` will fail saying you requested version `1.0.1` but the one in its `zos.json` is `1.0.0`, and the way to circumvent this is to link it using a semver as `zos link your-evm-package@^1.0.0`.

Which should be the expected behaviour in this kind of scenario? Should we expect the users to bump their zOS project too? What can be improved from the CLI to avoid users publishing unsynched versions of `package.json` and `zos.json` files?

spalladino:
Changes all dependency version validations to check against the package.json version instead of the zos.json version. This allows publishing an update of the npm package without modifying the zos.json versions. The downside is that we are coupling the zos.json declared dependency with npm versioning (the dependencies node of the zos.json file will now list dependencies with their npm version, and not their zos version), but we can review this in the future when we allow other package managers to be used. We can discuss whether this is the best approach though, as it allows to completely unsync the npm version and the zos version for a dependency, and whether to merge it now in case it introduces potential new issues.

We do use the zos.json version to check the zos.network.json files. If a dependency updates its zos.json version, but fails to update a zos.network.json, the consumer will reject using the dependency on that network.

Fixes #308

### DIFF
--- a/packages/cli/src/models/dependency/Dependency.js
+++ b/packages/cli/src/models/dependency/Dependency.js
@@ -15,7 +15,9 @@ export default class Dependency {
   }
 
   static satisfiesVersion(version, requirement) {
-    return !requirement || version === requirement || semver.satisfies(version, requirement);
+    return !requirement 
+      || version === requirement 
+      || semver.satisfies(version, requirement);
   }
 
   static async install(nameAndVersion) {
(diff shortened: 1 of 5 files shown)


facuspagnuolo:
We already discussed this offline, but I'm dropping my comment here anyway. I don't like the idea of coupling the package.json version to the zos.json. I'd prefer working on some improvements that prevent the user to get those files unsynched, additionally, we should reconsider the versions validations we are doing to see if there is a way to handle this scenario improving the user experience.

PR #310, opened 2018-10-22
Use contract method `estimateGas` function instead of `web3.eth.estimateGas` for contract method calls

### LINKED ISSUE #297: Fail to deploy to rinkeby using `truffle-hdwallet-provider` when having dependencies linked.
I'm trying to deploy an unpublished project with a linked dependency (eth-did-registry), and I'm getting the following error:
```console
foo@bar:~$ npx zos push --network infura_rinkeby --skip-compile --deploy-dependencies --verbose
[NetworkAppController] Deploying ethr-did-registry contracts
[Package] Deploying new Package...
[Package] Deployed Package 0xb1d6f84fbcd4d497fb7d4d3a4b9bf4535216a0a8
[Package] Adding new version...
[ImplementationDirectory] Deploying new ImplementationDirectory...
[ImplementationDirectory] Deployed ImplementationDirectory at 0xca9557b6e2140e7eaf1fd9ce249f2294ea486a28
[Error] Error: Failed deployment of dependency ethr-did-registry with error: Error: ProviderError: gas required exceeds allowance or always failing transaction
    at allPromisesOrError (/home/jota/work/zeppelin/zos/packages/cli/lib/utils/async.js:36:11)
    at process._tickCallback (internal/process/next_tick.js:68:7)

```

`truffle.js` config:

```javascript
infura_rinkeby: {
  name: "infura_rinkeby",
  network_id: "4",
  provider:  () =>  {
    return new HDWalletProvider(process.env.WALLET_MNEMONIC, `https://rinkeby.infura.io/v3/${process.env.INFURA_API_KEY}`)
  },
  gas: 8000000,
  gasPrice: 22000000000,
  from: '0x1e...'
},

```

I'm running the same command using a rinkeby geth node with the same gas/gasPrice and it works fine.

EDIT: I'm having the same problem using `truffle-hdwallet-provider` and the geth  node, so I would assume that the problem is not infura, but `truffle-hdwallet-provider`.

jcarpanelli:
Fixes #297 and #441 

The problem was not related with dependencies as mentioned in #297, but with contract method calls when using `truffle-hdwallet-provider`. For some reason, `web3.eth.estimateGas` fails to execute in this specific case, but the `estimateGas` function from the method (i.e., `mycontract.myFunction.estimateGas`) doesn't.

### DIFF
--- a/packages/lib/src/utils/Transactions.js
+++ b/packages/lib/src/utils/Transactions.js
@@ -93,8 +93,8 @@ async function _sendTransaction(contractFn, args = [], txParams = {}) {
   }
 
   // Estimate gas for the call
-  const estimateGasTxParams = { ... txParams, ... contractFn.request(...args).params[0] };
-  const gas = await estimateActualGas(estimateGasTxParams);
+  const gas = await estimateActualGasFnCall(contractFn, args, txParams)
+
   return contractFn(...args, { gas, ... txParams });
 }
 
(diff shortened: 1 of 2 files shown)


spalladino:
We are putting this PR on hold since we could not reproduce the issue on the integration tests. We'll look for more info on the cause of the problem before applying the fix.

PR #331, opened 2018-10-23
Add a new 'ZEPToken' section to the documentation

### LINKED ISSUE #150: Add guides about the token and vouching process
Add a section explaining how to interact with the token and vouching contracts through the CLI.
We may also want to add some details about those contracts and where are they published.

ajsantander:
Fix #150


packages/docs/docs/docs/zeptoken.md
```diff
@@ -0,0 +1,62 @@
...
+sidebar_label: The ZEPToken
+---
+
+ZeppelinOS' native ZEP token is used to incentivize the creation and maintenance of a healthy ecosystem of EVM packages. It does so via a series of mechanisms that are detailed in the following sections of this article.
+
+## Compatibility standards
+
+ZEP is an ERC20 token, controlled by a TPL (Transaction Permission Layer) jurisdiction (TODO: add link). This allows regional jurisdictions to control which addresses can hold or exchange ZEP using KYC, or whatever other validation process are found to be adequate for the region.
+
+## EVM package staking
```
facuspagnuolo: @elopio we already decided to use `vouch` instead right?

packages/docs/docs/docs/zeptoken.md
```diff
@@ -0,0 +1,62 @@
...
+_NOTE: This feature is not implemented in the current ZeppelinOS release._
+
+## Marketplace
+TODO
+
+## Scheduler
+TODO
+
+## Governance
+TODO
```
facuspagnuolo: I wouldn't add a section about the marketplace, the scheduler and governance for now

packages/docs/docs/docs/zeptoken.md
```diff
@@ -0,0 +1,62 @@
...
+
+If a contribution is assimilated without the provision of a reward, the contribution's author may challenge the EVM package.
+
+_NOTE: This feature is not implemented in the current ZeppelinOS release._
+
+## EVM package linking fees
+
+EVM package developers may choose to set up a linking fee in ZEP, in which case anyone who links to the EVM package will need to provide a one time payment for such linkage to occur.
+
+_NOTE: This feature is not implemented in the current ZeppelinOS release._
```
facuspagnuolo: I'd move the `NOTE`s right under the section's title

packages/docs/docs/docs/zeptoken.md
ajsantander: Leo and I discussed that we will not document anything that is not yet implemented, so I will be removing the sections you suggest, plus others.

packages/docs/docs/docs/zeptoken.md
ajsantander: As stated, we will not document non-implemented features, so these notes will be removed.

packages/docs/docs/docs/zeptoken.md
```diff
@@ -0,0 +1,38 @@
...
+
+The current version of this contract currently supports the following features:
+- Registering a new EVM package.
+- Vouching for an EVM package (only package owners for now).
+- Removing stake from an EVM package.
+- Removing an EVM package from the registry.
+
+## ZEP token private beta
+
+Zeppelin is currently distributing mainnet ZEP to projects interested in adopting ZeppelinOS. This is not a sale, but beta testing for the aforementioned token dynamics. To apply to the private beta distribution, please submit your project's details in the [ZEP Token's private beta submission form](https://google.com)(TODO add link).
```
spalladino:
```suggestion
Zeppelin is currently distributing mainnet ZEP to projects interested in adopting ZeppelinOS. This is not a sale, but beta testing for the aforementioned token dynamics. To apply to the private beta distribution, please submit your project's details in the [ZEP Token's private beta submission form](https://beta-registration.zeppelinos.org).
```

packages/docs/docs/docs/zeptoken.md
```diff
@@ -0,0 +1,38 @@
+---
+id: zeptoken
+title: The ZEPToken
+sidebar_label: The ZEPToken
+---
+
+ZeppelinOS' native ZEP token is used to incentivize the creation and maintenance of a healthy ecosystem of EVM packages. It does so via a series of mechanisms that are detailed in the following sections of this article.
```
spalladino: I'd mention in the very first paragraph that the ZEP token is **not** needed for using ZeppelinOS, only for participating in this ecosystem.

packages/docs/docs/docs/zeptoken.md
```diff
@@ -0,0 +1,38 @@
...
+
+## Compatibility standards
```
spalladino: I'd move this section to the end, and merge it with the private beta. This way, after mentioning that the ZEP token is managed via TPL, we can explain that in order to be accepted into the TPL jursidiction the user can apply to the private beta.

packages/docs/docs/docs/zeptoken.md
```diff
@@ -0,0 +1,38 @@
...
+
+ZEP is an ERC20 token, controlled by a [TPL](https://github.com/TPL-protocol) (Transaction Permission Layer) jurisdiction. This allows regional jurisdictions to control which addresses can hold or exchange ZEP using KYC, or whatever other validation process are found to be adequate for the region.
+
+The ZEP token's source code can be found at [`zos/packages/vouching/contracts/ZEPToken.sol`](https://github.com/zeppelinos/zos/blob/master/packages/vouching/contracts/ZEPToken.sol).
+
+## EVM package vouching
+
+Anyone can create ZeppelinOS EVM packages for free. More precisely, you don't need to hold or spend ZEP to create an EVM package. 
+
+However, ZeppelinOS provides a mechanism in which an EVM package can be registered and vouched for using ZEP. The collected stake of an EVM package can be challenged by other ZEP holders whenever a deficiency in the EVM package is presented for evaluation. In such a situation, the package's stake could be slashed in favor of the challenger.
```
spalladino: I think @elopio had the goal to use `vouch` instead of `stake`, to avoid confusion. If so, we should change all mentions to stake in this doc.

packages/docs/docs/docs/zeptoken.md
```diff
@@ -0,0 +1,38 @@
...
+
+This simple mechanism allows ZEP that is vouched for an EVM package to represent:
```
spalladino: It looks odd to list a set of features here, only to have a following paragraph pointing out which is the _actual_ set of features. I'd reprhase this sentence mentioning that the _end goal_ for the ZEP token vouching is x, y, and z, to then mention the actions available now.

packages/docs/docs/docs/zeptoken.md
ajsantander: Yes, I am aware of the preference of `vouch` vs `stake`. However, I find it hard to put into sentences because the latter is quantitative and the former is qualitative.

Some examples:
"The collected stake of an EVM package" => "The collected vouch of an EVM package"
"A package's stake could be slashed..." => "A package's vouch could be slashed..."

spalladino: We could use:
- "The tokens vouched for an EVM Package"
- "The vouched tokens for a package could be slashed..."

But I'll defer that decision to you and @elopio, to remain consistent with the rest of the site.

Issue #43, opened 2018-08-24
Track owner addresses in zos.network.json files


spalladino:
We should track the owner address of the App, Package, and Provider contracts deployed to a network in the zos.network.json file. This way, we can validate whether the `from` address actually has permissions to run a particular action, before attempting to actually execute it, and save the user from wasting time and gas into a failing tx.

This would complement the owner tracking of individual proxies in https://github.com/zeppelinos/zos/issues/3.

vs77bb:
@mahim23 You've been approved! A WIP PR sooner than later helps, please let us know if you have any questions along the way 🙂 cc @spalladino

gitcoinbot:
@mahim23 Hello from Gitcoin Core - are you still working on this issue? Please submit a WIP PR or comment back within the next 3 days or you will be removed from this ticket and it will be returned to an ‘Open’ status. Please let us know if you have questions!
* [x] warning (3 days)
* [ ] escalation to mods (6 days)

Funders only: Snooze warnings for [1 day](https://gitcoin.co/issue/zeppelinos/zos/43/1271?snooze=1) | [3 days](https://gitcoin.co/issue/zeppelinos/zos/43/1271?snooze=3) | [5 days](https://gitcoin.co/issue/zeppelinos/zos/43/1271?snooze=5) | [10 days](https://gitcoin.co/issue/zeppelinos/zos/43/1271?snooze=10) | [100 days](https://gitcoin.co/issue/zeppelinos/zos/43/1271?snooze=100)

mahim23:
Hey! I have started looking through the codebase and tried a few things. Will clarify any doubts and send a WIP PR soon.

mahim23:
@spalladino I went through the code for the cli and the `zos.network.json` file is being handled by the `ContollerFor` network class. So, I should add the owner attribute to contract/app/package and change it in the `networkFile` attribute in the `NetworkBaseController` object, whenever it is being updated.

Is this approach right or am I missing something?

spalladino:
@mahim23 sounds good! The `ZosNetworkFile` is only handled by the `NetworkBaseController` (via its children `NetworkApp` and `NetworkLib`), so you should be able to injet these values on the `fetchOrDeploy` method. 

The next step, which is actually checking that the owner matches the sender address before sending every tx in the future, is more tricky. The sender may not be available in the `txParams`, and we may need to query the node to actually get the default sender address. We'd also need to be careful to check that these addresses match in every public method of the `NetworkController`s.

[...]

PR #332, opened 2018-10-24
Add "owner" field to zos.network.json file

Destiner:
This PR adds a new field to the zos.network.json file, named `owner`. It tracks an account that pushed the app.

Related to #43.

Open questions:
* If "from" is not specified, I pull owner address from `web3.eth.coinbase`. Is that a correct approach? It's seems a bit low-level, and also await calls to web3 can be quite slow.
* Should we add owner once, or add it to package, app, and provider individually? Currently it's the former.
* Wording


packages/lib/src/utils/Transactions.js
```diff
@@ -79,6 +79,10 @@ export async function sendDataTransaction(contract, txParams) {
   return contract.sendTransaction({ gas, ... txParams });
 }
 
+export async function getCoinbase() {
+  return web3.eth.coinbase;
```
facuspagnuolo: please promisify the web3 call

packages/cli/test/scripts/push.test.js
```diff
@@ -323,6 +323,19 @@ contract('push script', function([_, owner]) {
     });
   };
 
+  const shouldSaveOwner = function () {
+    it('should save owner when not specifying "from"', async function () {
+      const txParams = {};
+      await push({ networkFile: this.networkFile, network, txParams });
+      should.exist(this.networkFile.owner);
```
facuspagnuolo: based on your implementation, this one should check the saved owner is the coinbase address, however, I think the owner is going to be the first address in the list instead of the coinbase

packages/cli/src/models/network/NetworkBaseController.js
```diff
@@ -344,4 +345,11 @@ export default class NetworkBaseController {
     await this.project.freeze()
     this.networkFile.frozen = true
   }
+
+  async _saveOwner() {
+    if (this.networkFile.owner == null) {
+      const owner = this.txParams.from || await getCoinbase();
+      this.networkFile.owner = owner;
+    }
+  }
```
facuspagnuolo: we should save the address of each contract individually

Destiner:
Hi there.

Anything I missed with latest commits?

packages/lib/src/utils/Transactions.js
Destiner: Replaced with web3.eth.accounts[0].

packages/lib/src/utils/Transactions.js
```diff
@@ -79,6 +79,10 @@ export async function sendDataTransaction(contract, txParams) {
   return contract.sendTransaction({ gas, ... txParams });
 }
 
+export async function getAccount(index) {
+  return web3.eth.accounts[index];
```
facuspagnuolo: AFAIK this is not async, I think it would be clearer using a promisified version of `getAccounts`

packages/cli/src/models/network/NetworkBaseController.js
```diff
@@ -344,4 +345,25 @@ export default class NetworkBaseController {
...
+      appPackage.owner = owner;
+      this.networkFile.package = appPackage;
+    }
+    const provider = this.networkFile.provider;
+    if (provider.owner == null) {
+      const owner = this.txParams.from || await getAccount(0);
+      provider.owner = owner;
+      this.networkFile.provider = provider;
+    }
+  }
```
facuspagnuolo: I think it is more correct to save this info when we create these contracts. For example, if you already have deployed these contracts and you run the push command, you will be storing the current sender address as the owner of those contracts you deployed before, which may not be right. 
Additionally, you could look for the owner only once with `const owner = this.txParams.from || await getAccount(0);`

nfurfaro: It would be nice to track the owner when creating with `zos create` as well, for those not publishing.

Destiner:
Moved saving of the owner to the `publish` script and updated tests.

spalladino:
Closing as stale

PR #334, opened 2018-10-24
docs: avoid the use of upgradeability

elopio:
Upgradeability is a very complicated word, that we usually can easily avoid.

[CI skip]


facuspagnuolo:
@elopio please fix conflicts and merge

PR #356, opened 2018-10-25
Throw error if a dependency was unpublished when trying to link it

### LINKED ISSUE #296: Error when linking unpublished evm package
Given a project that has an evm package linked that has not been published, running `zos push` on the project yields the following error:

```
{ Error: Failed to link dependency dependency@1.0.0 with error: Contract or address expected
    at allPromisesOrError (.../node_modules/zos/lib/utils/async.js:36:11)
    at <anonymous>
    at process._tickCallback (internal/process/next_tick.js:182:7) cb: [Function: cb] }
```

Instead, show a friendlier error telling the user that the library needs to be published in order to be linked, and suggesting that they can run `zos push --deploy-dependencies` in the meantime to set up a custom deployment.

Also, verify that `--deploy-dependencies` *will* deploy an unpublished evm package that has a zos.network.json for the current network.

spalladino:
Fixes #296

### DIFF
--- a/packages/cli/package-lock.json
+++ b/packages/cli/package-lock.json
@@ -6583,6 +6583,10 @@
       "version": "file:test/mocks/mock-stdlib-undeployed-2",
       "dev": true
     },
+    "mock-stdlib-unpublished": {
+      "version": "file:test/mocks/mock-stdlib-unpublished",
+      "dev": true
+    },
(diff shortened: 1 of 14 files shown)


facuspagnuolo:
I'm putting this PR on-hold until we tag v2.0.0

PR #357, opened 2018-10-25
Added a bit more details for the push command.

ajsantander:
Fix https://github.com/zeppelinos/zos/issues/19

### DIFF
--- a/packages/docs/docs/docs/deploying.md
+++ b/packages/docs/docs/docs/deploying.md
@@ -115,9 +115,13 @@ And back in the original terminal:
 zos push --network local
 ```
 
-This creates a `zos.dev-<network_id>.json` file with all the information about your
-project in this specific network. You can read more about this file format
+This deploys `MyContract` to the specified network. If your project added other contracts (using the `add` command) they would be deployed as well. An important thing to understand is that the contracts deployed by the `push` command are implementation or logic contracts and are not intended to be used directly, rather to be linked to a proxy as we will see later in the [Upgrading your project](https://docs.zeppelinos.org/docs/upgrading.html) section.
+
+Calling `push` creates a `zos.dev-<network_id>.json` file with all the information about your
+project in this specific network, including the addresses of the deployed contract implementations in `contracts["MyContract"].address`. You can read more about this file format
 in the [configuration files](configuration.md#zos-network-json) section.
+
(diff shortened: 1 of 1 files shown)


packages/docs/docs/docs/deploying.md
```diff
@@ -115,10 +115,14 @@ And back in the original terminal:
 zos push --network local
 ```
 
-This creates a `zos.dev-<network_id>.json` file with all the information about your
-project in this specific network. You can read more about this file format
+This deploys `MyContract` to the specified network. If your project added other contracts (using the `add` command) they would be deployed as well. An important thing to understand is that the contracts deployed by the `push` command are implementation or logic contracts and are not intended to be used directly, rather to be linked to a proxy as we will see later in the [Upgrading your project](https://docs.zeppelinos.org/docs/upgrading.html) section.
+
+Calling `push` creates a `zos.dev-<network_id>.json` file with all the information about your
```
```suggestion
Calling `push` creates a `zos.network.json` file with all the information about your
```

`dev-network_id` is only used when the network is not well-known, such as kovan or mainnet

packages/docs/docs/docs/deploying.md
```diff
@@ -115,10 +115,14 @@ And back in the original terminal:
 zos push --network local
 ```
 
-This creates a `zos.dev-<network_id>.json` file with all the information about your
-project in this specific network. You can read more about this file format
+This deploys `MyContract` to the specified network. If your project added other contracts (using the `add` command) they would be deployed as well. An important thing to understand is that the contracts deployed by the `push` command are implementation or logic contracts and are not intended to be used directly, rather to be linked to a proxy as we will see later in the [Upgrading your project](https://docs.zeppelinos.org/docs/upgrading.html) section.
```
spalladino: Question: are we talking about proxies at this stage, or do we refer to them as "instances"?

ajsantander: @elopio :point_up: ?

packages/docs/docs/docs/deploying.md
ajsantander: I think that's reasonable, because the narrative is using `network --local` at this point. It later mentions a "Deploying to mainnet" guide.

packages/docs/docs/docs/deploying.md
facuspagnuolo: Yes, please let's talk about upgradeable instances and avoid proxy unless we are talking really low-level

PR #359, opened 2018-10-25
Do not check gas price if working on ganache

### LINKED ISSUE #328: Improve the gas price check
Currently, the default workflow of `zos` ends in an error message:
Cowardly refusing to execute transaction with gas price set to Truffle's default of 100 gwei. Consider explicitly setting a different value in your truffle.js file.

This is problematic, our default experience should be the green path with no errors. Also, it's not clear what's going on, and it's not clear how to solve it.

A few things can improve this easily:
* Skip the gas check if we are using a ganache local network.
* Explain that we are being cowards because the gas price is set too high, and you will be probably wasting a lot of money. We shouldn't say "cowardly" at all in the error message :)
* Explain the gas price can be updated setting the gasPrice value on truffle, or link to a guide about estimating gas prices, or something.

For the future, maybe we should make this command interactive with a prompt: are you sure you want to waste so much gas on this? Or maybe, having always a confirmation: This command will spend this much gas (the estimated required gas is x), are you sure you want to continue? And probably, we should have a `--force` flag to skip the check altogether.

spalladino:
Also rephrased error message, and added link to ethgasstation for
obtaining a reasonable value.

Fixes #328

### DIFF
--- a/packages/lib/test/src/utils/Transactions.test.js
+++ b/packages/lib/test/src/utils/Transactions.test.js
@@ -62,13 +62,6 @@ contract('Transactions', function([_account1, account2]) {
       actualValue.toNumber().should.eq(42);
     });
 
-    it('refuses to send tx with truffle default gas price', async function () {
-      // Gas price check is skipped in test env, so we trick the Transactions module into thinking it's actually on a real run
-      process.env.NODE_ENV = 'not-test';
-      await sendTransaction(this.instance.initialize, [42, 'foo', [1,2,3]]).should.be.rejectedWith(/cowardly refusing/i);
-      process.env.NODE_ENV = 'test';
-    });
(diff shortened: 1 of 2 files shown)


packages/lib/test/src/utils/Transactions.test.js
```diff
@@ -13,7 +13,7 @@ import { setInterval } from 'timers';
 
 const DEFAULT_GAS = 6721975;
 
-contract('Transactions', function([_account1, account2]) {
+contract.only('Transactions', function([_account1, account2]) {
```
facuspagnuolo: 👍

spalladino:
@facuspagnuolo removed `.only` and tested on local ganache

PR #362, opened 2018-10-25
Review security warnings messages and capture them in docs

### LINKED ISSUE #248: Review security warnings messages and capture them in docs
This file contains all security warnings we are currently throwing at the user: https://github.com/zeppelinos/zos/blob/master/packages/cli/src/interface/ValidationLogger.js

As discussed, we should:
- review all texts
- ensure there is part of the docs that cover those warns (`writing.md` _should_ be covering most)
- ideally add links to those sections from the texts

spalladino:
Updates all validation messages from ValidationLogger to simplify them
and add links to documentation. Add to writing contracts doc info on any
validations missing.

Fixes #248

### DIFF
--- a/packages/docs/docs/docs/writing_contracts.md
+++ b/packages/docs/docs/docs/writing_contracts.md
@@ -186,6 +186,16 @@ contract MyContract is Initializable {
 }
 ```
 
+## Potentially unsafe operations
+
+When working with upgradeable smart contracts, you will always interact with the contract instance, and never with the underlying logic contract. However, nothing prevents a malicious actor from sending transactions to the logic contract directly. This does not pose a threat, since any changes to the state of the logic contracts do not affect your contract instances, as the storage of the logic contracts is never used in your project.
+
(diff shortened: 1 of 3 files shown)


packages/docs/docs/docs/writing_contracts.md
```diff
@@ -186,6 +186,16 @@ contract MyContract is Initializable {
 }
 ```
 
+## Potentially unsafe operations
+
+When working with upgradeable smart contracts, you will always interact with the contract instance, and never with the underlying logic contract. However, nothing prevents a malicious actor from sending transactions to the logic contract directly. This does not pose a threat, since any changes to the state of the logic contracts do not affect your contract instances, as the storage of the logic contracts is never used in your project.
+
+There is, however, an exception. If the direct call to the logic contract triggers a `selfdestruct` operation, then the logic contract will be destroyed, and all your contract instances will end up delegating all calls to an address without any code. This would effectively break all contract instances in your project.
+
+A similar effect can be achieved is the logic contract contains a `delegatecall` operation. If the contract can be made to `delegatecall` into a malicious contract that contains a `selfdestruct`, then the calling contract will be destroyed.
```
elopio: s/is/if

PR #371, opened 2018-10-26
Add create proxy example with multiple args

jcarpanelli:
Related to https://github.com/zeppelinos/zos/issues/343


spalladino:
Couldn't have thought of a better sample value for the string variable given the context ;-)

PR #383, opened 2018-10-26
docs: polish the deploying guide

elopio:
[CI skip]


packages/docs/docs/docs/deploying.md
```diff
@@ -78,9 +78,9 @@ contract MyContract is Initializable {
 ```
 
 Notice that our sample contract has an `initialize` function instead of the
-standard Solidity constructor. This is required by ZeppelinOS if you want to
-upgrade your contract later, which we'll see in more detail in the next
-guide.
+standard Solidity constructor. This is the way to initialize variables on
+ZeppelinOS in order to allow a proxy to call this function and thus make the
+contract upgradeable, which we'll see in more detail in the next guide.
```
facuspagnuolo: I like the clarification but I'll rewrite this sentence to avoid the word `proxy`

PR #392, opened 2018-10-26
Remove all todos from docs

### LINKED ISSUE #343: Docs polish suggestions
This review is based on commit https://github.com/zeppelinos/zos/commit/542ba321f535608f672453b9418e9362041c6ca7.

### General comments
- [x] Avoid `NODE_EVN=test` statements (#335)
- [x] Note that `zos.local.json` is no longer true since we are using the `dev-#` for local nodes, consider using sth like `zos.dev-*.json` instead (https://github.com/zeppelinos/zos/pull/348)
- [x] Typos, wording and minor fixes proposal (https://github.com/zeppelinos/zos/pull/346)
- [x] Confusing random bold words in code-blocks (#340)
- [x] Improve Solidity syntax highlighting for code-blocks (#341)

### Questions
- [x] Many links point to an `advanced` section that seems to be removed, we should either add it or fix the links that are still pointing to it (#376)
- [ ] Do we want to introduce the `session` command to avoid the `--network local` in every example?
- [ ] I'm still wondering the result of `upgrade` vs `update` discussion (https://github.com/zeppelinos/zos-cli/issues/338)

### Overview section
- [ ] Consider being more explicit about smart contracts upgrades

- [x] Fix `initializers vs constructors` link, it's broken

### Quickstart section
- [x] Rewrite the initializers clarifying sentence to make clear that this is they way constructors should be tackled now in ZeppelinOS to make them upgradeable. Note that this is different from saying that is required if you want to upgrade your contract later (#383)

> `pragma solidity ^0.4.24`
- [x] Should we add a note saying that the current version of ZeppelinOS works with 0.4.24 by default. Otherwise, users should use the --skip-compile option providing their own build files. (#383)

> Notice that our sample contract has an initialize function instead of the standard Solidity constructor. This is required by ZeppelinOS if you want to upgrade your contract later, which we'll see in more detail in the next guide.
- [x] Consider adding at least a small sentence explaining what problem is the `Initializable` contract solving (#383)

> This contract imports another contract from the zos-lib package.
- [x] Consider adding a line to clarify what is `zos-lib`. It is the first time it's mentioned (#383)

- [x] Consider adding a note to clarify that since zOS is using Truffle, there is no need to install it (https://github.com/zeppelinos/zos/pull/370)

> Note: When calling an initializer with many variables, these should be passed as a comma-separated list, with no spaces in between.
- [x] Consider adding an example (https://github.com/zeppelinos/zos/pull/371)

This section ends without testing the created instance
- [x] Consider adding a simple test with the Truffle console testing that `MyContract.at(..).x() === 42` (https://github.com/zeppelinos/zos/pull/391)

- [x] Consider adding a link to the section where that command was run

The upgrade test does not work since we are requesting the user to use the same Truffle console opened before, but given that the contract was changed, it should be loaded again, otherwise, users will face a `TypeError: myContract.increment is not a function` error
> To try the new function we just added, jump to the terminal where the Truffle command is running and execute the following instructions:
- [x] Consider running a `ganache` instance instead of a `truffle develop` console at the beginning of the guide, and then opening a `truffle console` against it every time we want the user to check something. Bear in mind that this will impact the blueprint we are using for the `truffle-config.js` on `init`, we will need either change the default port to be `8545` or make the user run ganache-cli with the `--port 9545` option

- [x] Solve that todo (#378)

### Guides
- [x] Fix the `in the advanced topics section` link, it's broken

#### Using the ZeppelinOS programmatic library

> And now, let's write our upgrading script in index.js:
- [ ] Consider splitting this script into at least 2 or 3 steps 

> As you can see on the code, this script was prepared to be executed with Truffle. So let's start a Truffle local development network and console:
- [x] No need to run a truffle console to run a script with truffle, we can use `truffle exec script.js` instead (https://github.com/zeppelinos/zos/pull/370)

### Advanced Topics
- [x] Fix link, it's broken (#379)

ajsantander:
Fix #343

### DIFF
--- a/packages/docs/docs/docs/zos_lib.md
+++ b/packages/docs/docs/docs/zos_lib.md
@@ -136,5 +136,3 @@ truffle exec index.js --network local
 This is just a very simple example to show the basic functions of the
 ZeppelinOS JavaScript library. You can find more examples in the
 [ZeppelinOS repository](https://github.com/zeppelinos/zos/tree/master/examples).
-
-// TODO add link to the API docs.


facuspagnuolo:
@elopio please create an issue for it

PR #397, opened 2018-10-26
Improve quickstart guide to work with session

facuspagnuolo:


packages/docs/docs/docs/deploying.md
```diff
@@ -113,23 +113,38 @@ project writing it in the `zos.json` configuration file.
...
+Once we have done that, let's go back to the original terminal and 
+run the following command:
 
 ```console
-zos push --network local
+zos session -n local -f 0x1df62f291b2e969fb0849d99d9ce41e2f137006e --expires 3600 
+``` 
+
+The `session` command starts a session to work with a desired network.
+In this case, we are telling to work with the `local` network with the
```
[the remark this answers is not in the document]
spalladino: I think I got this tip from @nventuro: always avoid short forms of flags or commands when writing documentation, and prefer longer forms as they are more descriptive and easy to understand at first glance.

packages/docs/docs/docs/upgrading.md
```diff
@@ -35,11 +35,18 @@ just as with any other contract on the blockchain, we can define governance
 mechanisms to decide when and how to upgrade the contracts, that can be manual,
 automated, or any combination of both that will earn the trust of our users.
 
-Let's create an upgradeable instance of this contract so you can experiment what
-this is all about:
+We also started a session to work with the `local` network, before we continue, 
+let's renew the session to ensure it is not expired:
```
spalladino: I'd move this into a callout, and go with something like:
```
> If the following commands fail with an `A network name must be provided to execute the requested action` error, it means our session has expired. Renew it by running the command `zos session -n local -f 0x1df62f291b2e969fb0849d99d9ce41e2f137006e --expires 3600` again.
```

packages/docs/docs/docs/upgrading.md
```diff
@@ -35,11 +35,18 @@ just as with any other contract on the blockchain, we can define governance
...
 
 ```console
-zos create MyContract --init initialize --args 42,hitchhiker --network local
+zos session -n local -f 0x1df62f291b2e969fb0849d99d9ce41e2f137006e --expires 3600
+```
+
+Great, now let's create an upgradeable instance of this contract so you can
```
spalladino: -1 to `Great` and similar expressions, but it's because I'm a boring guy. If @elopio approved it, then it's ok with me.

packages/docs/docs/docs/deploying.md
```diff
@@ -113,23 +113,38 @@ project writing it in the `zos.json` configuration file.
...
-zos push --network local
+zos session -n local -f 0x1df62f291b2e969fb0849d99d9ce41e2f137006e --expires 3600 
+``` 
+
+The `session` command starts a session to work with a desired network.
+In this case, we are telling to work with the `local` network with the 
+`-n` option, and also setting a default sender address for the transactions 
+we will run with the `-f` option. Additionally, the `expires` flag allows 
+us to indicate the session expiration time in seconds. 
+
```
spalladino: I'd add here another callout explaining at this very point that they should not use the default address in `--from` because link-to-transparent proxies.

facuspagnuolo:
@elopio @spalladino comments addressed, can you please review?

packages/docs/docs/docs/deploying.md
```diff
@@ -113,23 +113,45 @@ project writing it in the `zos.json` configuration file.
...
+``` 
+
+The `session` command starts a session to work with a desired network.
+In this case, we are telling it to work with the `local` network with the 
+`--network` option, and also setting a default sender address for the 
+transactions we will run with the `--from` option. Additionally, the 
+`expires` flag allows us to indicate the session expiration time in seconds. 
+
+> Note that we are using an specific address for the `--from` option which 
+is different to the default address that `ganache-cli` would use by default.
```
elopio: you are using 'default' twice in a row here. I suggest to remove the second.

facuspagnuolo:
@elopio please review

packages/docs/docs/docs/deploying.md
```diff
@@ -113,23 +113,45 @@ project writing it in the `zos.json` configuration file.
...
+zos session --network local --from 0x1df62f291b2e969fb0849d99d9ce41e2f137006e --expires 3600 
+``` 
+
+The `session` command starts a session to work with a desired network.
+In this case, we are telling it to work with the `local` network with the 
+`--network` option, and also setting a default sender address for the 
+transactions we will run with the `--from` option. Additionally, the 
+`expires` flag allows us to indicate the session expiration time in seconds. 
+
+> Note that we are using an specific address for the `--from` option which
```
```suggestion
> Note that we are using a specific address for the `--from` option which 
```

PR #413, opened 2018-11-05
Add Etherscan as remote verification application for 'verify' command

ProtonGustave:
Related to https://github.com/zeppelinos/zos/issues/251

Etherscan works with apikey, so this update also add --api-key option to verify command. Example:
`zos verify --remote etherscan --api-key <api key> -n rinkeby <contract name>`

### DIFF
--- a/packages/cli/src/models/Verifier.js
+++ b/packages/cli/src/models/Verifier.js
@@ -9,13 +9,18 @@ const Verifier = {
   async verifyAndPublish(remote, params) {
     if (remote === 'etherchain') {
       await publishToEtherchain(params)
+    } else if (remote === 'etherscan') {
+      await publishToEtherscan(params)
     } else {
-      throw new Error('Invalid remote. Currently, ZeppelinOS contract verifier only supports etherchain as remote verification application.')
+      throw new Error('Invalid remote. Currently, ZeppelinOS contract verifier supports etherchain(mainnet only) and etherscan as remote verification applications.')
     }
   }
 }
 
 async function publishToEtherchain(params) {
+  if(params.network !== 'mainnet')
+    throw new Error('Invalid network. Currently, etherchain supports only mainnet')
+  
   const etherchainVerificationUrl = 'https://www.etherchain.org/tools/verifyContract'
   const etherchainContractUrl = 'https://www.etherchain.org/account'
   const { compilerVersion, optimizer, contractAddress } = params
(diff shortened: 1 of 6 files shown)


packages/cli/src/models/Verifier.js
```diff
@@ -9,13 +9,18 @@ const Verifier = {
   async verifyAndPublish(remote, params) {
     if (remote === 'etherchain') {
       await publishToEtherchain(params)
+    } else if (remote === 'etherscan') {
+      await publishToEtherscan(params)
     } else {
-      throw new Error('Invalid remote. Currently, ZeppelinOS contract verifier only supports etherchain as remote verification application.')
+      throw new Error('Invalid remote. Currently, ZeppelinOS contract verifier supports etherchain(mainnet only) and etherscan as remote verification applications.')
```
jcarpanelli: nit: add a space between *etherchain* and *(mainnet only)*.

packages/cli/src/models/Verifier.js
```diff
@@ -47,4 +52,103 @@ async function publishToEtherchain(params) {
   }
 }
 
+async function publishToEtherscan(params) {
+  if (params.apiKey === undefined) {
+    throw new Error('API key not specified.')
```
jcarpanelli: Let's be a bit more explicit here, and rephrase the error to: "*Etherscan API key not specified. To get one, follow this link: https://etherscancom.freshdesk.com/support/solutions/articles/35000022163-i-need-an-api-key*", or something similar.

packages/cli/src/models/Verifier.js
```diff
@@ -47,4 +52,103 @@ async function publishToEtherchain(params) {
...
+      })
+
+      if (response.status === 200) {
+        if (response.data.status === '1') {
+          return true
+        } else {
+          return false
+        }
+      } else {
+        throw new Error(`Error while trying to verify contract; ${response.data.status}; ${response.data.message}; ${response.data.result}`)
```
jcarpanelli: Showing all the etherscan's messages could be a bit annoying. 
```console
foo@bar:~$ Error while trying to verify contract; 0; NOTOK; Missing or invalid ApiKey
```
Let's remove the `response.data.status` and the `response.data.message`, and only keep the `response.data.result`.

packages/cli/src/models/Verifier.js
```diff
@@ -47,4 +52,103 @@ async function publishToEtherchain(params) {
...
+  } else if (network === 'kovan') {
+    apiSubdomain = 'api-kovan'
+  } else {
+    throw new Error('Invalid network. Currently, etherscan supports mainnet, rinkeby, ropsten and kovan')
+  }
+
+  const etherscanApiUrl = `https://${apiSubdomain}.etherscan.io/api`
+  const etherscanContractUrl = `https://${network}.etherscan.io/address`
+
+  const checkVerificationStatus = async (guid) => {
```
jcarpanelli: Could we move this function to another place? So we can keep the `publishToEtherscan` function smaller and cleaner.

packages/cli/src/models/Verifier.js
```diff
@@ -47,4 +52,103 @@ async function publishToEtherchain(params) {
...
+        compilerversion: compiler,
+        optimizationUsed: optimizerStatus,
+        runs: params.optimizerRuns,
+      }),
+      headers: {
+        'Content-type': 'application/x-www-form-urlencoded'
+      }
+    })
+    if (response.status === 200 && response.data.status === '1') {
+      log.info(`Contract verification in process(this is usually under 30 seconds)...\nGUID: ${response.data.result}`)
```
jcarpanelli: nit: add a space between *process* and *(..)* and rephrase it a bit to something like: *(this usually takes under 30 seconds)*.

Also, a question here: Is it necessary to print the `GUID`?

packages/cli/src/models/Verifier.js
```diff
@@ -47,4 +52,103 @@ async function publishToEtherchain(params) {
...
+
+      await new Promise((resolve, reject) => {
+        const checkIntervalId = setInterval(async () => {
+          const verificationStatus = await checkVerificationStatus(response.data.result)
+
+          if (verificationStatus === false) {
```
jcarpanelli: nit: let's change this to `if (!verificationStatus)`

packages/cli/src/models/Verifier.js
```diff
@@ -47,4 +52,103 @@ async function publishToEtherchain(params) {
...
+
+          log.info(`Contract verified successfully. You can check it here: ${etherscanContractUrl}/${contractAddress}#code`)
+
+          clearInterval(checkIntervalId)
+
+          resolve()
+        }, 3000)
+      })
+    } else {
+      throw new Error(`Error while trying to verify contract; ${response.data.status}; ${response.data.message}; ${response.data.result}`)
```
jcarpanelli: Same here, let's only show the `response.data.result`.

ProtonGustave:
@jcarpanelli done(without solidity libs option).

packages/cli/test/scripts/verify.test.js
```diff
@@ -77,7 +77,11 @@ contract('verify script', function () {
       await assertVerify(contractAlias, { network, networkFile: this.networkFile, remote: 'invalid-remote' }, /Invalid remote/)
     })
 
-    it('throws error if contract could not be verified', async function () {
+    it('throws error if specifying anything but mainnet as a network for etherchain', async function() {
+      await assertVerify(contractAlias, { network, networkFile: this.networkFile, remote: 'etherchain' }, /Invalid network/)
+    })
+    
+    /*it('throws error if contract could not be verified on etherchain', async function () {
```
jcarpanelli: We want these test cases to be working, and we should also add tests covering the new etherscan implementation. It would be a good idea to restructure the `contract verification` tests to something like:

```javascript
describe('validations', function() {
  // already existent tests 
})

describe('contract verification', function() {
  describe('against etherscan', function() {
    // new tests**
  })
  describre('against etherchain', function() {
   // here we should move the old tests we have for etherchain. 
  })
})
```
** We should at least cover the same cases we are covering for etherchain.

Last but not least: you can fix these commented tests by adding the `network: 'mainnet'` option in the object parameter :wink:.

ProtonGustave:
@jcarpanelli here we go

jcarpanelli:
Awesome, @ProtonGustave! Thank you very much :smile:!

PR #422, opened 2018-11-20
Added encodeCall import

PaulRBerg:
Small addition, but it'd be really helpful to let devs know how they can import the `encodeCall` helper function in their tests.


packages/docs/docs/website/versioned_docs/version-2.0.0/testing.md
```diff
@@ -111,6 +111,8 @@ This means that calls to contracts with more than one function named `initialize
 The current solution to this issue is to `npm install zos-lib` and use the same helper function `zos create` uses: `encodeCall`. `encodeCall` receives the signature of your `initialize` function, as well as its arguments and their types. It then crafts the calldata which you can send in a raw call. Here's an example:
 
 ```js
+const encodeCall = require('zos-lib/lib/helpers/encodeCall').default;
```
facuspagnuolo: shouldn't this be `const { encodeCall } = require('zos-lib')` ?

@jcarpanelli can you confirm if this change should be placed in the versioned dir?

jcarpanelli: This should be added in both `docs/docs/website/versioned_docs/version-2.0.0/testing.md` and `docs/docs/docs/testing.md` so we don't lose track of it in future versions of this file.

PaulRBerg: I've just added it to `docs/docs/docs/testing.md` too!

PR #443, opened 2018-11-28
Implemented TsMigration, Lib Phase 0: Bootstrapping

### LINKED ISSUE #432: TsMigration, Lib Phase 0: Bootstrapping

ajsantander:
Fix #432 
**READY TO MERGE** :tada:

## Description

This PR is a proof of concept of how Typescript could be integrated into the project. The implementation plays along with Lerna just fine, and uses Typescript as a replacement of Babel to produce ES5 Javascript output. Other approaches involve keeping Babel and just using Typescript for type checking, independently of the build pipeline, but we are not choosing that approach here. This PR aims to completely remove Babel out of the build pipeline.

## Current status
Right now, tsc is set globally and "activated" in the lib, cli and vouching packages. This means that the activated packages are using the tsc compiler instead of Babel to output ES5 Javascript. Lerna prepare was run globally and the zos cli was verified to work, with at least one ported to .ts in the activated packages.

**Tests are currently not passing. (See below)**

## Tests
At the moment, Babel is still included as a dependency on each of the packages because it is used by Truffle tests. After removing Truffle from the testing pipeline, we should be able to completely remove all Babel dependencies in favor of Typescript.

## Tasklist
- [x] Implement tsc as a global dependency.
- [x] Implement tsc in packages/lib.
- [x] Verify that the current setup does not break truffle tests.
- [x] Port something in packages/lib. Ported packages/cli/src/utils/FileSystem.ts 
- [x] Implement tsc in packages/cli.
- [x] Port something in packages/cli. Ported packages/cli/src/utils/sleep.ts.
- [x] Implement tsc in packages/vouching.
- [x] Port something in packages/vouching. Ported packages/vouching/src/helpers/validateAddress.ts.
- [x] Hook up a linter.
- [x] Make sure all tests pass. **TECHNICAL PROBLEM** Truffle needs to be removed from the tests because it uses Babel, our tests use code from the packages that may now be written in Typescript, and Babel doesn't understand Typescript. Once we remove Truffle from the tests, we will be able to transpile the tests from Typescript-ish to Javascript, and then run them.


spalladino:
@ajsantander this is awesome!! I'll review the files now. As for the items in the list: I would definitely **not** add TS to the docs package, so feel free to cross them out. And if we remove Babel, which I agree if we're going full TS, maybe we can get @frangio to join us in this project :-)

Also, when you mean "installing tsc as a global dependency", I take it you mean installing it in the global scope. Can we instead install it at the root package.json, where we are keeping lerna?

packages/cli/src/utils/sleep.ts
```diff
@@ -1,6 +1,3 @@
-'use strict'
-
-export default function sleep(ms) {
+export default function sleep(ms:number):Promise<void> {
```
spalladino: I'm getting excited already!

packages/lib/src/utils/FileSystem.ts
```diff
@@ -0,0 +1,111 @@
+import * as fs from 'fs'
+import * as path from 'path'
+
+// TS-TODO: Analyze which of these functions could be encapsulated.
+// TS-TODO: Make this a class with static methods?
```
spalladino: Please don't! It's perfectly ok for plain functions to be _just functions_, no need to wrap them in a class, object, or anything.

tsconfig.json
```diff
@@ -0,0 +1,17 @@
+{
+  "compilerOptions": {
+    "module": "commonjs",
+    "target": "es5",
+    "allowJs": true,
+    "checkJs": false,
```
spalladino: What happens if we enable this flag?

ajsantander:
> Also, when you mean "installing tsc as a global dependency", I take it you mean installing it in the global scope. Can we instead install it at the root package.json, where we are keeping lerna?
Yes, that's what I meant. The Typescript dependency is installed in `./package.json` (root).

```json
{
  "name": "root",
  "private": true,
  "devDependencies": {
    "@types/node": "^10.12.10",
    "lerna": "~3.3.2",
    "tslint": "^5.11.0",
    "typescript": "^3.1.6"
  }
}
```

tsconfig.json
ajsantander: From http://www.typescriptlang.org/docs/handbook/compiler-options.html:
"Report errors in .js files. Use in conjunction with --allowJs."

ajsantander:
> Looks great @ajsantander!! Do you think it'd be worthwhile to give https://github.com/Microsoft/dts-gen a shot, in case we want to start typing CLI without having typed lib entirely? I don't have any strong opinions, just stumbled upon it, and wanted to make sure it was in your radar.

IMHO it's good to know that this tool is out there, but I wouldn't use it to type our stuff. I think this kind of tool is useful if you happen to use a small dependency and want to use it with types, but you couldn't find any types for it, anywhere. It would be a bummer to have to write the types yourself, so you'd use this tool to quickly get the problem out of your way and keep coding in happy Ts-land.

In our case, since this is our lib, we'll be more than happy to write our own `.d.ts` files. Right?

ajsantander:
> Also: can we check if a typed module exported from zos-lib is properly recognised in zos-cli? What I mean is: given function f(a:string) exported from zos-lib, if we attempt to do require('zos-lib').f(4) in CLI, does the ts compiler load the typing from zos-lib and properly understand it?

Looks like types from lib don't make it to cli unless the `declaration: true` setting is set in `tsconfig.json`. Unfortunately, we can't use this setting until we use the setting `allowJs: false`, which means that we may need to completely port lib before we port cli. I'm guessing that we could just write `.d.ts` files manually on the meantime, or even use https://github.com/Microsoft/dts-gen, but I'd go for the simpler path of just porting lib first, to the point of which we can let the Typescript compiler auto generate the declaration files.

spalladino:
Sounds good! Something I discussed with Jota offline: let's focus on typing
only the exported functions/classes from each module (ie file) first, and
not the internal or private functions. We can then start typing everything
internal to each module as we move forward. Also, don't hesitate to
prioritise not just based on topographical order (ie the modules with less
dependencies first), consider also the importance or relevance of the
module within the project.

packages/cli/package.json
```diff
@@ -6,7 +6,8 @@
   "private": false,
   "license": "MIT",
   "scripts": {
-    "prepare": "rm -rf lib && babel src --out-dir lib",
+    "compile-babel": "rm -rf lib && babel src --out-dir lib",
+    "prepare": "npm run compile-babel",
```
facuspagnuolo: is it necessary to create a separate script?

packages/lib/package.json
```diff
@@ -47,14 +49,7 @@
...
   "devDependencies": {
-    "babel-cli": "^6.26.0",
-    "babel-plugin-transform-object-rest-spread": "^6.26.0",
-    "babel-polyfill": "^6.26.0",
-    "babel-preset-env": "^1.7.0",
-    "babel-preset-es2015": "^6.24.1",
-    "babel-preset-stage-2": "^6.24.1",
-    "babel-preset-stage-3": "^6.24.1",
-    "babel-register": "^6.26.0",
+    "@types/node": "^10.12.12",
```
facuspagnuolo: 😍

packages/lib/src/utils/Contracts.ts
```diff
@@ -0,0 +1,73 @@
...
+// TODO: rename to Artifacts and move to /artifacts
+
+export default class Contracts {
+
+  private static DEFAULT_SYNC_TIMEOUT:number = 240000;
+  private static DEFAULT_BUILD_DIR:string = `${process.cwd()}/build/contracts`;
+
+  private static timeout:number = Contracts.DEFAULT_SYNC_TIMEOUT;
+  private static buildDir:string = Contracts.DEFAULT_BUILD_DIR;
+  private static artifactDefaults:any = {};
```
facuspagnuolo: these variables are still initializable by constructor right?

packages/lib/src/artifacts/ZWeb3.ts
```diff
@@ -0,0 +1,177 @@
...
+  1:  'mainnet',
+  2:  'morden',
+  3:  'ropsten',
+  4:  'rinkeby',
+  42: 'kovan'
+};
+
+// TS-TODO: Type Web3.
+// TS-TODO: Review what could be private in this class.
+export default class ZWeb3 {
```
jcarpanelli: I don't remember if we discussed this, but personally, I'm not comfortable with having classes full of static methods when we could have objects instead. I know you proposed to discuss this on https://github.com/zeppelinos/zos/issues/434 's phase 6, but I also consider this is not an implementation (javascript/typescript) concern but rather a design issue.

packages/lib/src/artifacts/ContractFactory.ts
```diff
@@ -0,0 +1,213 @@
+import BN from 'bignumber.js';
+import ZWeb3 from './ZWeb3';
+import decodeLogs from '../helpers/decodeLogs';
+import { getSolidityLibNames, hasUnlinkedVariables } from '../utils/Bytecode';
+
+interface ContractSchema {
+  contractName:string;
```
jcarpanelli: Just an OCD here, but I really prefer if we add a space before types from the beginning of this beautiful and grateful process of typscripting the repo:
```suggestion
  contractName: string;
```

packages/lib/src/test/behaviors/Ownable.ts
```diff
@@ -0,0 +1,59 @@
...
+
+  describe('transferOwnership', function() {
+    describe('when the proposed owner is not the zero address', function() {
+      const newOwner = anotherAccount;
+
+      describe('when the sender is the owner', function() {
+        const from = owner;
+
+        it('transfers the ownership', async function() {
+
```
jcarpanelli: nit: Let's avoid adding unnecessary empty lines. I know we don't have rules for this, but I also believe we've been pretty consistent so far :smile:

packages/lib/src/test/helpers/assertRevert.ts
```diff
@@ -1,10 +1,12 @@
-export default async function assertRevert(promise, invariants = () => {}) {
+import assert from 'assert';
+
+export default async function assertRevert(promise:Promise<any>, invariants = ():void => {return;}) {
```
jcarpanelli: Can we avoid the `return` statement in these cases?

packages/lib/src/utils/Addresses.ts
```diff
@@ -0,0 +1,23 @@
+import _ from 'lodash';
+
+export const ZERO_ADDRESS:string = '0x0000000000000000000000000000000000000000';
+
+// TS-TODO: Web3 typings? => contract.
+export function toAddress(contractOrAddress:string | any):string {
```
jcarpanelli: :heart_eyes:

packages/lib/src/utils/errors/DeployError.ts
```diff
@@ -0,0 +1,8 @@
+export class DeployError extends Error {
+  constructor(error:Error, props:any) {
+    super(error.message);
+    this.stack = error.stack;
+    this.name = 'DeployError';
+    Object.keys(props).forEach(( prop:string ) => this[prop] = props[prop]);
```
jcarpanelli: nit: I don't see any reason for adding spaces between the arrow function's params, as we are not adding them in regular functions.
```suggestion
    Object.keys(props).forEach((prop: string) => this[prop] = props[prop]);
```

packages/lib/src/utils/Contracts.ts
ajsantander: I don't think so... There is no constructor implemented in this class atm, and the variables are meant to be used statically, as in:
```
Contracts.timeout = 20;
Contracts.buildDir = 'some/path';
Contracts.artifactDefaults = {...};
```

packages/lib/src/artifacts/ZWeb3.ts
ajsantander: Yeah, I also like to stay away from "static classes" or singleton-ish patterns, but it wasn't my task to change the existing design. I found an object being used "statically" and I figured that a "static class" would serve as a first off translation. It works. I would re-address this discussion at a later phase when we focus on design.

packages/lib/src/artifacts/ContractFactory.ts
ajsantander: Sure. If you want to change the linter setting now, could you please modify the existing .ts files to use the convention, so that we don't get an explosion of linting errors all around.
Otherwise, you can add a TS-TODO note on the linter configuration (root/tsling.json) to change it later when we revisit the compiler and linter settings in Phase 5 of the plan.

packages/lib/src/test/behaviors/Ownable.ts
ajsantander: Sure! It's something I do to navigate paragraphs with Vim. Sorry if I am not aware of current conventions since I am not that familiar with the code right now. My suggestion is that you configure TSLint so that we start enforcing this rule with it. Again, you could do it now or just add a note to the linter setttings.

packages/lib/src/test/helpers/assertRevert.ts
ajsantander: I'm pretty sure we can :laughing: 
That definitely looks weird... let me see what I can do.

packages/lib/src/utils/errors/DeployError.ts
ajsantander: Agreed! Can you treat is as the above comments about TSLint? There is currently a recommended rule being enforced that forces () always on arrow functions which is kind of a pain.

ajsantander:
Rebased to master to include fixes to integration tests #473

packages/lib/src/test/helpers/assertRevert.ts
ajsantander: Done.

packages/lib/src/artifacts/ZWeb3.ts
facuspagnuolo: IMO, if we turn plain objects into classes we are changing the existing design, actually, I've been having many conflicts due to this. Can we stick to the kind of object that was decided during the migration please?

PR #446, opened 2018-12-03
This process does not work with ZOS

crazyrabbitLTC:
Attempting to deploy (and EVM package) using zos following the mainnet recomendations as listed here fails with the error: 

Cowardly refusing to execute transaction with excessively high default gas price of 100 gwei. Consider explicitly setting a different gasPrice value in your truffle.js config file. You can check reasonable values for gas price in https://ethgasstation.info/.

We need to set the gas value and gas price. I added an example above, however this is not always enough either and can fail with either: 

ProviderError: insufficient funds for gas * price + value

or

Error: Contract transaction couldn't be found after 50 blocks

or

Error: ProviderError: gas required exceeds allowance or always failing transaction

They contract may still be deployed however in the case of the final error example.


facuspagnuolo:
Thanks @crazyrabbitLTC

PR #462, opened 2018-12-06
Updated OpenZeppelin to v2.0.

nventuro:
Fixes https://github.com/zeppelinos/zos/issues/565

OpenZeppelin 2.0 has been audited and has a stable API, so migrating early to the 2.x branch will prevent future upgrade issues.

### DIFF
--- a/packages/lib/contracts/application/ImplementationDirectory.sol
+++ b/packages/lib/contracts/application/ImplementationDirectory.sol
@@ -2,13 +2,15 @@ pragma solidity ^0.4.24;
 
 import "./ImplementationProvider.sol";
 import "openzeppelin-solidity/contracts/ownership/Ownable.sol";
-import 'openzeppelin-solidity/contracts/AddressUtils.sol';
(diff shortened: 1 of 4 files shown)


nventuro:
This is failing because the App tests use OpenZeppelin's `expectEvent` test helper, which changed behavior from 1.10 to 2.0: the 2.0 version actually checks the values emitted in the event, whereas in 1.10 only the event name was compared.

The test suit has a mistake in that it expects the `version` field in the `PackageChanged` event to be a string (it is an array). Sadly, since `expectEvent` doesn't support checking arrays of `BigNumber` instances, the new test helper cannot be used to test this. I've [opened an issue](https://github.com/OpenZeppelin/openzeppelin-test-helpers/issues/1) in the newly created [`openzeppelin-test-hel[ers`](https://github.com/OpenZeppelin/openzeppelin-test-helpers) package to track this - once implemented, I'll update this PR and fix the test.

spalladino:
Closing in favour of #739

PR #463, opened 2018-12-06
[WIP] Add custom compiler using solc

facuspagnuolo:


packages/lib/src/utils/FileSystem.ts
```diff
@@ -1,14 +1,14 @@
-import * as fs from 'fs';
-import * as path from 'path';
```
ajsantander: This may cause an incompatibility with the Ts port. No action needed tho, we'll see...

packages/cli/test/models/compiler/solidity/SolidityProjectCompiler.test.js
```diff
@@ -0,0 +1,39 @@
+require('../../../setup')
```
ajsantander: It would be interesting to use the native solc compiler (not solcjs) and compare bytecode output. This could be done easily: https://github.com/ajsantander/solz/blob/master/src/compile/Compiler.js#L33

packages/cli/src/models/initializer/ConfigVariablesInitializer.js
```diff
@@ -1,11 +1,14 @@
...
 import Session from '../network/Session'
+import Compiler from '../compiler/Compiler'
 import TruffleConfig from './truffle/TruffleConfig'
 
 const ConfigVariablesInitializer = {
   initStaticConfiguration() {
     const buildDir = TruffleConfig.buildDir()
     Contracts.setLocalBuildDir(buildDir)
+    const solcSettings = TruffleConfig.solcSettings()
+    Compiler.setSettings(solcSettings)
```
ajsantander: If truffle is not used, how does the user specify the compiler settings?

facuspagnuolo: That's an issue we have to solve somehow, already discuss that with @spalladino and we decided to tackle it later (see https://github.com/zeppelinos/zos/issues/502)

packages/cli/test/models/compiler/solidity/SolidityProjectCompiler.test.js
facuspagnuolo: Niceeee, we should! Already added an issue for that (https://github.com/zeppelinos/zos/issues/506)

packages/cli/src/commands/add.js
```diff
@@ -20,7 +20,7 @@ const register = program => program
 
 async function action(contractNames, options) {
   ConfigVariablesInitializer.initStaticConfiguration()
-  if(!options.skipCompile) await compile()
+  if(!options.skipCompile) await Compiler.call()
```
jcarpanelli: Have you thought about a way of centralizing this `Compiler.call()` call so we don't call it in each command `action` function?

packages/cli/src/models/compiler/Compiler.js
```diff
@@ -0,0 +1,46 @@
...
+import { FileSystem, Contracts, Logger } from 'zos-lib'
+import TruffleConfig from '../initializer/truffle/TruffleConfig'
+import SolidityProjectCompiler from './solidity/SolidityProjectCompiler'
+
+const log = new Logger('Compiler')
+
+export default {
+  async call() {
+    return this._isTruffleProject()
+      ? this.compileWithTruffle()
```
jcarpanelli: If we don't call this method outside the class, let's add the `_` at the beginning of the function name for consistency.
```suggestion
      ? this._compileWithTruffle()
```

packages/cli/src/models/compiler/Compiler.js
```diff
@@ -0,0 +1,46 @@
...
+      : this.compileWithSolc()
```
jcarpanelli:
```suggestion
      : this._compileWithSolc()
```

packages/cli/src/models/compiler/solidity/SolidityContractsCompiler.js
```diff
@@ -0,0 +1,181 @@
+import _ from 'lodash'
+import solc from 'solc'
+import axios from 'axios'
+import { Logger, FileSystem } from 'zos-lib'
+import SolidityDependenciesFinder from './SolidityDependenciesFinder'
+
+const log = new Logger('SolidityContractsCompiler')
+
+const DEFAULT_OPTIMIZER = { runs: 200, enabled: false }
```
jcarpanelli: Isn't `runs` a property used only when the optimiser is enabled?

packages/cli/test/commands/share.js
```diff
@@ -64,6 +63,7 @@ exports.stubCommands = function () {
     this.update = sinon.stub(update, 'default')
     this.verify = sinon.stub(verify, 'default')
     this.setAdmin = sinon.stub(setAdmin, 'default')
+    this.compiler = sinon.stub(Compiler, 'call').callsFake(() => null)
```
jcarpanelli: :+1:

packages/cli/test/models/compiler/solidity/SolidityContractsCompiler.test.js
```diff
@@ -0,0 +1,344 @@
+import CaptureLogs from "../../../helpers/captureLogs";
+
+require('../../../setup')
+
+import { FileSystem } from 'zos-lib'
+import SolidityContractsCompiler from '../../../../src/models/compiler/solidity/SolidityContractsCompiler'
+
+describe('SolidityContractsCompiler', function () {
+
+  const contract_Solc04 = {
```
jcarpanelli: I don't feel comfortable with having so many constant objects declared here. We should move them to another file, or at least move them outside the `describe` scope if we are not saving them in `this`.

packages/cli/src/models/compiler/Compiler.js
facuspagnuolo: I left those as is in case someone would like to compile with a particular compiler besides using Truffle or not

packages/cli/src/models/compiler/Compiler.js
facuspagnuolo: same above

packages/cli/src/commands/add.js
facuspagnuolo: Added an issue for that :) https://github.com/zeppelinos/zos/issues/508

PR #464, opened 2018-12-07
Token mechanics challenges + public vouching proposal

### LINKED ISSUE #433: Token mechanics design with public vouching
The current design of token mechanics assumes that only EVM package developers vouch for their own packages. Extend the design to support vouching by anyone.

facuspagnuolo:
Fixes #433

### DIFF
--- a/design/token-mechanics-v2.md
+++ b/design/token-mechanics-v2.md
@@ -1,0 +1,324 @@
+# Token Mechanics V2
+
+### Assumptions
+- Users will register/deprecate versions of a package. 
+- Users will vouch/unvouch for an specific version of a package. 
+- Users will be able to move their vouch through different versions of the same package. 
+- Challengers will challenge a specific version of a package.
+- The minimum stake is validated against the summation of the amount vouched by the package owner for each of its version.
+- There is no penalization for unvouching/moving tokens
+- Packages reputation could be calculated based on total vouch amount weighted by the registered versions
+
+### Issues
+1- How we can avoid users unvouching their tokens given a new critical challenge yet unsolved?
+> We can either use cyclic windows to state when users can challenge, when users can vouch/unvouch, and when challenges 
+  will be solved. Or, we can lock a fraction of the vouched tokens of each user to ensure we can pay successful 
+  challenges. Both options have their downsides: locking may not be possible if there is a large list of challenges 
+  if the payout formula depends on the total stake. OTOH, windows may no be the ideal solution for critical challenges
+
+2- Challenges front-running?
+> We won't tackle this issue for this version.
+
+3- Challenges payout formula
+  3.1- Severity
+  > Each challenge can have a severity label given by the challenger. 
+
+  3.2- How we can make sure challenges payout cannot be manipulated based on their resolution order? 
+  > An obvious approach would be to solve challenges in a FIFO order.
+
+  3.3- How we can make sure there is enough tokens to pay successful challenges?
+  > We should decide whether we want to accept defaults or not.
+    If not, we should pay a proportional amount of tokens based on the severity, the challenge amount and the total stake. 
+    
+  3.4- How we will handle a huge list of vouchers during a payout?
+  > We can have a rate per version in ZEP that will be adjust each time a challenge is solved.  
+
+4- What would happen if package registrars' vouch reaches an amount lower than the minimum stake due to challenges payout?
+> Users will have to provide a certain amount of tokens the next time they register a version to reach the minimum stake.
+  We should make sure this is an edge scenario since we will be forcing package registrars to buy more tokens.
+
+### Proof of concept
+
+#### Types
+```
+[packages]   = string (name) => { package, owner (address), ownerTotalVouch (uint) }
+[package]    = string (version ID) => { vouch, totalVouch (uint), rate (fraction), deprecated (bool) } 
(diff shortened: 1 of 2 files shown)


design/token-mechanics-v2.md
```diff
@@ -0,0 +1,324 @@
...
+
+3- Challenges payout formula
+  3.1- Severity
+  > Each challenge can have a severity label given by the challenger. 
+
+  3.2- How we can make sure challenges payout cannot be manipulated based on their resolution order? 
+  > An obvious approach would be to solve challenges in a FIFO order.
+
+  3.3- How we can make sure there is enough tokens to pay successful challenges?
+  > We should decide whether we want to accept defaults or not.
```
spalladino: _Default_ may not be the best word in this context to explain this. Maybe "we want to accept Argentina-2001s"?

design/token-mechanics-v2.md
```diff
@@ -0,0 +1,324 @@
...
+    If not, we should pay a proportional amount of tokens based on the severity, the challenge amount and the total stake.
```
spalladino: Allowing the user to get a payout based on the challenge amount is risky. I'm not sure we want to scale the payout based on how much the reporter did "bet" on the bug.

design/token-mechanics-v2.md
```diff
@@ -0,0 +1,324 @@
...
+
+  3.2- How we can make sure challenges payout cannot be manipulated based on their resolution order? 
+  > An obvious approach would be to solve challenges in a FIFO order.
+
+  3.3- How we can make sure there is enough tokens to pay successful challenges?
+  > We should decide whether we want to accept defaults or not.
+    If not, we should pay a proportional amount of tokens based on the severity, the challenge amount and the total stake. 
+    
+  3.4- How we will handle a huge list of vouchers during a payout?
+  > We can have a rate per version in ZEP that will be adjust each time a challenge is solved.
```
spalladino: To clarify: when a user stakes N ZEP on a OpenZeppelin 2.2 (for instance), he should get N "OpenZeppelin 2.2" tokens (probably not tradeable). Same as any other voucher. If they want to unvouch, they redeem their OZ 2.2 tokens for ZEP at a 1:1 rate. However, if OZ 2.2 gets slashed, the way to reduce the stake of each voucher is to simply change the rate. For instance, if the version gets slashed at a 10%, instead of reducing the number of ZEP staked by each of them, we just adjust the ZEP:OZ2.2 rate to be 9:10 instead of 1:1.

design/token-mechanics-v2.md
```diff
@@ -0,0 +1,324 @@
...
+
+  3.3- How we can make sure there is enough tokens to pay successful challenges?
+  > We should decide whether we want to accept defaults or not.
+    If not, we should pay a proportional amount of tokens based on the severity, the challenge amount and the total stake. 
+    
+  3.4- How we will handle a huge list of vouchers during a payout?
+  > We can have a rate per version in ZEP that will be adjust each time a challenge is solved.  
+
+4- What would happen if package registrars' vouch reaches an amount lower than the minimum stake due to challenges payout?
+> Users will have to provide a certain amount of tokens the next time they register a version to reach the minimum stake.
```
spalladino: Keep in mind that we cannot necessarily enforce this, depending on how packages are "registered".

design/token-mechanics-v2.md
```diff
@@ -0,0 +1,324 @@
...
+  We should make sure this is an edge scenario since we will be forcing package registrars to buy more tokens.
+
+### Proof of concept
+
+#### Types
+```
+[packages]   = string (name) => { package, owner (address), ownerTotalVouch (uint) }
+[package]    = string (version ID) => { vouch, totalVouch (uint), rate (fraction), deprecated (bool) }
```
spalladino: A package needs an address as well

design/token-mechanics-v2.md
```diff
@@ -0,0 +1,324 @@
...
+
+4- What would happen if package registrars' vouch reaches an amount lower than the minimum stake due to challenges payout?
+> Users will have to provide a certain amount of tokens the next time they register a version to reach the minimum stake.
+  We should make sure this is an edge scenario since we will be forcing package registrars to buy more tokens.
+
+### Proof of concept
+
+#### Types
+```
+[packages]   = string (name) => { package, owner (address), ownerTotalVouch (uint) }
```
spalladino: Why are we keeping the owner and totalvouch out of the package structure?

design/token-mechanics-v2.md
```diff
@@ -0,0 +1,324 @@
+# Token Mechanics V2
```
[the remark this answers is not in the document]
facuspagnuolo: nice catch, given this will imply a breaking change I'd call it v3

design/token-mechanics-v2.md
facuspagnuolo: my bad, it was supposed to be part of each `package`

design/token-mechanics-v2.md
facuspagnuolo: I think that the way I proposed the registration process to work with public vouching, we can. LMKYT

design/token-mechanics-v2.md
facuspagnuolo: sounds good

PR #491, opened 2018-12-11
Port App, Project, Proxy and copyContract

jcarpanelli:
Related to https://github.com/zeppelinos/zos/issues/475
@ajsantander would you mind taking a look at it? :slightly_smiling_face:


packages/lib/src/helpers/copyContract.ts
```diff
@@ -0,0 +1,38 @@
+import { promisify } from 'util';
```
ajsantander: Are you using prepare_ts.sh or otherwise git mv instead of mv? That way we get more atomic diffs like this one: https://github.com/zeppelinos/zos/pull/443/commits/56bc171da4ac19136169d3b105b3e08e9a092219

ajsantander: Otherwise LGTM!

jcarpanelli: > Are you using prepare_ts.sh or otherwise git mv instead of mv? That way we get more atomic diffs like this one: [56bc171](https://github.com/zeppelinos/zos/commit/56bc171da4ac19136169d3b105b3e08e9a092219)

I will use it from now on :slightly_smiling_face:

packages/lib/src/application/App.ts
```diff
@@ -0,0 +1,158 @@
...
+import { isZeroAddress, toAddress } from '../utils/Addresses';
+import { buildCallData, callDescription, CalldataInfo } from '../utils/ABIs';
+import ContractFactory, { ContractWrapper } from '../artifacts/ContractFactory';
+import { toSemanticVersion, semanticVersionEqual } from '../utils/Semver';
+import { deploy as deployContract, sendTransaction, sendDataTransaction } from '../utils/Transactions';
+
+const log: any = new Logger('App');
+
+export default class App {
+  private appContract: any;
```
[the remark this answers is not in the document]
ajsantander: Ah, but I see that if you do so, you will not have members like `getPackage(name)`...
What I did in application/ImplementationDirectory was add an interface that extended ContractWrapper with these additional members.

packages/lib/src/application/App.ts
```diff
@@ -0,0 +1,158 @@
...
+  public static getContractClass(): ContractFactory {
+    return Contracts.getFromLib('App');
+  }
+
+  constructor(appContract: ContractWrapper, txParams: any = {}) {
+    this.appContract = appContract;
+    this.txParams = txParams;
+  }
+
+  public async getPackage(name): Promise<{ package: Package, version: any }> {
```
ajsantander: version is probably a string right?

packages/lib/src/application/Package.ts
```diff
@@ -0,0 +1,86 @@
...
+import { toSemanticVersion, SemanticVersion } from '../utils/Semver';
+import { toAddress, isZeroAddress } from '../utils/Addresses';
+import ContractFactory, { ContractWrapper } from '../artifacts/ContractFactory';
+import { deploy as deployContract, sendTransaction } from '../utils/Transactions';
+
+const log: any = new Logger('Package');
+
+interface PackageInterface extends ContractWrapper {
+  address: string;
+  hasVersion?;
```
ajsantander: Probably an boolean: `hasVersion?:boolean;`

packages/lib/src/application/Package.ts
```diff
@@ -0,0 +1,86 @@
...
+  public async hasVersion(version: string): Promise<boolean> {
+    return this.packageContract.hasVersion(toSemanticVersion(version));
+  }
+
+  public async isFrozen(version: string): Promise<boolean> {
+    const directory = await this.getDirectory(version);
+    return directory.isFrozen();
+  }
+
+  public async freeze(version: string): Promise<any> {
```
ajsantander: return `Promise<any> | never`

packages/lib/src/utils/Semver.ts
```diff
@@ -5,34 +5,23 @@ import util from 'util';
 // TS-TODO: use typed SemVer dependency, some of these methods may actually
 // return a custom object.
 
-type SemanticVersion = [number, number, number];
+export type SemanticVersion = [number, number, number];
 type RawSemanticVersion = [any, any, any];
 
 export function toSemanticVersion(version: string | RawSemanticVersion): SemanticVersion | never {
-
```
ajsantander: Sorry for the lines =S

packages/lib/src/application/App.ts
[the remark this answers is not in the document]
jcarpanelli: Yeah, that's the same I did in `Package` and `Proxy`, but I decided to leave it an `any`. Should I extend `ContractWrapper` here as well, or I can leave it as it is?

packages/lib/src/application/Package.ts
jcarpanelli: `hasVersion` is a function. I prefer leaving this as it is, as we are probably adding `TypeChain` or whatsoever to type the ABI.

packages/lib/src/application/App.ts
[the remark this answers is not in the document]
ajsantander: As you prefer. At this point of the port either approach is ok.

Issue #475, opened 2018-12-10
TsMigration, Lib Phase 1: Syntax port of the src folder


ajsantander:
Following the migration plan from #434

jcarpanelli:
@facuspagnuolo, @ajsantander let's be sure we don't close this issue until we do a general review of all file ports.

PR #492, opened 2018-12-11
Ported lib/src/utils/Transactions.ts

ajsantander:
Part of solving #475


packages/lib/src/utils/ABIs.ts
```diff
@@ -83,18 +83,18 @@ function tryGetFunctionNodeFromMostDerivedContract(contractClass:ContractFactory
...
     return ast.getLinearizedBaseContracts(true);
   } catch (err) {
     // This lookup may fail on contracts loaded from libraries, so we just silently fail and fall back to other methods
     return null;
   }
 }
 
-export function callDescription(method:any, args:any):string {
-  const argsDescriptions:any = method.inputs.map((input:any, index:number) => ` - ${input.name} (${input.type}): ${JSON.stringify(args[index])}`);
+export function callDescription(method: any, args: any): string {
```
jcarpanelli: I think we can make these function definitions a bit less generic. I believe that method will always be a `string`, and args will always be a `string[]` (please check it out by yourself)

packages/lib/src/utils/Transactions.ts
```diff
@@ -109,61 +116,60 @@ async function _sendTransaction(contractFn, args = [], txParams = {}) {
...
-    state.nodeInfo = await ZWeb3.getNode()
+    state.nodeInfo = await ZWeb3.getNode();
   }
   return state.nodeInfo;
 }
 
-async function getETHGasStationPrice() {
-  if (state.gasPrice) return state.gasPrice;
+async function getETHGasStationPrice(): Promise<any> | never {
+  if (state.gasPrice) { return state.gasPrice; }
```
jcarpanelli: We should definitely review some tslint rules. Adding braces in a one-line statement is somewhat annoying.

packages/lib/src/artifacts/ZWeb3.ts
```diff
@@ -10,6 +11,34 @@ const NETWORKS = {
   42: 'kovan'
 };
 
+export interface Web3Transaction {
```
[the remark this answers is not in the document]
ajsantander: Yup, totally agreed! For now, Whenever I create an interface, I throw it in the same class that it's used. When we discuss architecture, it'd be great to consider your suggestion.
Also, if we do find typings for Web3 that work, this particular interface definition may not be needed at all.

packages/lib/src/utils/ABIs.ts
ajsantander: Ok on `args: string[]` but `method: string` doesn't work because `string` doesn't have the property `.inputs`.

packages/lib/src/utils/Transactions.ts
ajsantander: Agreed! Implemented:
`"curly": [true, "ignore-same-line"]`
"curly" also has the "as-needed" option, but I decided to not use it, because it would allow something like this, which I don't like:
```
if(blah)
  thenDoMeh();
```

packages/lib/src/utils/Transactions.ts
```diff
@@ -35,14 +42,14 @@ const TRUFFLE_DEFAULT_GAS_PRICE = BN(100000000000);
  * @param txParams other transaction parameters (from, gasPrice, etc)
  * @param retries number of transaction retries
  */
-export async function sendTransaction(contractFn, args = [], txParams = {}, retries = RETRY_COUNT) {
-  await fixGasPrice(txParams)
+export async function sendTransaction(contractFn: GenericFunction, args: any[] = [], txParams: any = {}, retries: number = RETRY_COUNT) {
```
jcarpanelli: I'm pretty sure `args` will always be an array of strings. Could you please check it out?
Also, the function's return type is missing.

packages/lib/src/artifacts/ZWeb3.ts
```diff
@@ -10,6 +11,34 @@ const NETWORKS = {
...
+  transactionIndex: number;
+  from: string;
+  to: string;
+  value: BN;
+  gasPrice: BN;
+  gas: number;
+  input: string;
+}
+
+export interface Web3TransactionReceipt {
```
jcarpanelli: I'm concerned about these interfaces, as they depend on external libraries that could change their structure in the future (just assuming things here, but I'm pretty sure `web3` has changed a lot from 0.x to 1.y). I would suggest to only define interfaces related to our classes, functions, etc. (note that `Web3TransactionReceipt` and `Web3Transaction` have a strong dependence with `web3`).

packages/lib/src/utils/Transactions.ts
```diff
@@ -53,14 +60,14 @@ export async function sendTransaction(contractFn, args = [], txParams = {}, retr
  * @param txParams other transaction parameters (from, gasPrice, etc)
  * @param retries number of deploy retries
  */
-export async function deploy(contract, args = [], txParams = {}, retries = RETRY_COUNT) {
-  await fixGasPrice(txParams)
+export async function deploy(contract: ContractFactory, args: any[] = [], txParams: any = {}, retries: number = RETRY_COUNT) {
```
jcarpanelli: nit: Return type is missing.

packages/lib/src/utils/Transactions.ts
ajsantander: Fixed `args: string[]`, but the return function is unknown. Setting it as `any` explicitly tho.

packages/lib/src/artifacts/ZWeb3.ts
ajsantander: Agreed, removing the interfaces.

PR #493, opened 2018-12-11
Ported some small files in lib to ts.

ajsantander:
Part of #475


packages/lib/src/helpers/advanceBlock.ts
```diff
@@ -1,6 +1,6 @@
-import ZWeb3 from '../artifacts/ZWeb3'
+import ZWeb3 from '../artifacts/ZWeb3';
 
-export default function advanceBlock () {
+export default function advanceBlock():Promise<any> {
```
jcarpanelli: Don't forget to run the linter before pushing :slightly_smiling_face:. Don't worry about fixing this though, I'll create a PR in a few minutes running the linter with the latest rules before starting the cli's migration.

Issue #482, opened 2018-12-10
TsMigration, Cli Phase 0: Bootstrapping 


ajsantander:
Following the migration plan from #434

ajsantander:
Solved by #494

PR #494, opened 2018-12-12
Bootstraps Typescript into the cli package

ajsantander:
Solves #482


ajsantander:
Not sure why the integration tests are failing at "Local geth node":
```
module.js:550
    throw err;
    ^
Error: Cannot find module 'ethereumjs-abi'
    at Function.Module._resolveFilename (module.js:548:15)
    at Function.Module._load (module.js:475:25)
    at Module.require (module.js:597:17)
    at require (internal/module.js:11:18)
    at Object.<anonymous> (/home/travis/build/zeppelinos/zos/tests/cli-app/node_modules/eth-sig-util/index.js:2:16)
```

jcarpanelli:
I'm blocking this PR until we fix and merge https://github.com/zeppelinos/zos/pull/491 https://github.com/zeppelinos/zos/pull/493, https://github.com/zeppelinos/zos/pull/492, https://github.com/zeppelinos/zos/pull/500 and review what we have done so far.

ajsantander:
CI Integrations tests failing, not sure why :(
https://travis-ci.org/zeppelinos/zos/builds/469003179?utm_source=github_status&utm_medium=notification

packages/cli/src/models/files/ZosNetworkFile.ts
```diff
@@ -0,0 +1,365 @@
...
+
+  get version(): string {
+    return this.data.version;
+  }
+
+  get frozen(): boolean {
+    return this.data.frozen;
+  }
+
+  get contracts(): any[] {
```
jcarpanelli: nit: This will always return an object, not an array.
Also, this port if one of those where it would be extremely useful to define our own interfaces. For example, we could define a basic `ContractInterface` listing all its attributes without going very deep:
```ts
interface ContractsInterface {
  [contractAlias: string]: ContractInterface
}

interface ContractInterface {
  address: string;
  constructorCode: string;
  localBytecodeHash: string;
  ...
  types: any,
  storage: any[],
  ...
}

File: packages/cli/src/models/files/ZosNetworkFile.ts [hunk 2]
```diff
@@ -0,0 +1,365 @@
...
+
+  get contractAliases(): string[] {
+    return Object.keys(this.contracts);
+  }
+
+  get isLightweight(): boolean {
+    return this.packageFile.isLightweight;
+  }
+
+  get solidityLibs(): any[] {
```
jcarpanelli: Same as above.

packages/cli/src/models/files/ZosNetworkFile.ts
```diff
@@ -0,0 +1,365 @@
...
+
+  public hasSolidityLib(libName: string): boolean {
+    return !_.isEmpty(this.solidityLib(libName));
+  }
+
+  public solidityLibsMissing(libs: any): string[] {
+    return _.difference(Object.keys(this.solidityLibs), libs);
+  }
+
+  public getSolidityLibOrContract(aliasOrName: string): any {
```
jcarpanelli: In this case, we could return a `Contract` (see comments above) or a `SolidityLib`

packages/cli/src/models/files/ZosNetworkFile.ts
```diff
@@ -0,0 +1,365 @@
...
+    return this.hasSolidityLib(aliasOrName) || this.hasContract(aliasOrName);
+  }
+
+  public updateImplementation(aliasOrName: string, fn: (...args: any[]) => any): void {
+    if (this.hasContract(aliasOrName)) this.data.contracts[aliasOrName] = fn(this.data.contracts[aliasOrName]);
+    else if (this.hasSolidityLib(aliasOrName)) this.data.solidityLibs[aliasOrName] = fn(this.data.solidityLibs[aliasOrName]);
+    else return;
+  }
+
+  get dependencies(): any[] {
```
jcarpanelli: nit: This always returns an object, not an array.

packages/cli/src/models/files/ZosNetworkFile.ts
```diff
@@ -0,0 +1,365 @@
...
+    return this.data.dependencies || {};
+  }
+
+  get dependenciesNames(): string[] {
+    return Object.keys(this.dependencies);
+  }
+
+  public getDependency(name: string): any | null {
```
jcarpanelli: Another useful interfaces we can define here are `Dependencies` and `Dependency`.

packages/cli/src/models/files/ZosNetworkFile.ts
```diff
@@ -0,0 +1,365 @@
...
+      }))
+    ));
+    return _.filter(allProxies, (proxy) => (
+      (!packageName || proxy.package === packageName) &&
+      (!contract || proxy.contract === contract) &&
+      (!address || proxy.address === address)
+    ));
+  }
+
+  // TS-TODO: return type can probably be typed to Proxy.ts from lib?
```
jcarpanelli: `ZosNetworkFile` edits/inserts/deletes `zos.network.json` content, and doesn't manage any other class instance such as `Proxy` from lib. 
Here comes the last pair of interfaces I would define: `ProxyInterface` and `ProxiesInterface`.

packages/cli/src/models/files/ZosNetworkFile.ts
```diff
@@ -0,0 +1,365 @@
...
+
+  public hasDependency(name: string): boolean {
+    return !_.isEmpty(this.getDependency(name));
+  }
+
+  public hasDependencies(): boolean {
+    return !_.isEmpty(this.dependencies);
+  }
+
+  public getProxies({ package: packageName, contract, address }: {package?: string, contract?: any, address?: string} = {}): any[] {
```
jcarpanelli:
```suggestion
   public getProxies({ package: packageName, contract, address }: { package?: string, contract?: any, address?: string } = {}): any[] {
```

packages/cli/src/models/files/ZosNetworkFile.ts
```diff
@@ -0,0 +1,365 @@
...
+  public hasSameBytecode(alias: string, klass: any): boolean {
+    const contract = this.contract(alias) || this.solidityLib(alias);
+    if (contract) {
+      const localBytecode = contract.localBytecodeHash;
+      const currentBytecode = bytecodeDigest(klass.bytecode);
+      return currentBytecode === localBytecode;
+    }
+  }
+
+  set version(version: string) {
```
jcarpanelli: If we can, let's try to define all return types (there are also some other methods below without them)

packages/cli/src/models/files/ZosVersion.ts
```diff
@@ -2,12 +2,10 @@ import _ from 'lodash';
...
-  if (version == ZOS_VERSION) {
-    return true;
-  } else if (_.isUndefined(version)) {
-    throw Error(`zos version identifier not found in ${where}. This means the project was built with an older version of zos (1.x), and needs to be upgraded. Please refer to the documentation at https://docs.zeppelinos.org for more info.`)
-  } else {
-    throw Error(`Unrecognized zos version identifier ${version} found in ${where}. This means the project was built with an unknown version of zos. Please refer to the documentation at https://docs.zeppelinos.org for more info.`)
-  }
-}
\ No newline at end of file
+export function checkVersion(version: string, where: any): boolean {
```
jcarpanelli: This function could also return `never`.

packages/cli/src/models/files/ZosVersion.ts
```diff
@@ -2,12 +2,10 @@ import _ from 'lodash';
...
+  if (version === ZOS_VERSION) return true;
+  else if (_.isUndefined(version))
```
jcarpanelli: Are we using the same linter rules that we are using in `lib`? If we are, please re-run the linter  :slightly_smiling_face:

packages/cli/src/models/initializer/ConfigVariablesInitializer.ts
```diff
@@ -0,0 +1,38 @@
...
+import Compiler from '../compiler/Compiler';
+import TruffleConfig from './truffle/TruffleConfig';
+
+interface NetworkOptions {
+  network: string;
+  from: string;
+  timeout: number;
+}
+
+const ConfigVariablesInitializer = {
```
jcarpanelli: :heart_eyes:

packages/cli/src/models/initializer/truffle/TruffleProjectInitializer.ts
```diff
@@ -0,0 +1,38 @@
+import path from 'path';
+import { FileSystem as fs } from 'zos-lib';
+import TruffleConfig from './TruffleConfig';
+
+const TruffleProjectInitializer = {
+
+  call(root = process.cwd()): void {
```
jcarpanelli: Pretty sure `root` is a string (or maybe `typescript` has its own `path` type? :stuck_out_tongue:)

packages/cli/src/models/status/StatusFetcher.ts
```diff
@@ -1,117 +1,120 @@
-import { ZWeb3, Contracts, Logger, FileSystem as fs, bytecodeDigest, bodyCode, constructorCode } from 'zos-lib'
+import { ZWeb3, Contracts, Logger, FileSystem as fs, bytecodeDigest, bodyCode, constructorCode } from 'zos-lib';
```
jcarpanelli: I don't see this file completely ported. Maybe you forgot to push a commit? If you are planning to finish this port in another PR, please uncheck it in https://github.com/zeppelinos/zos/issues/434 until you tackle it again so we don't forget to to port it in the future :slightly_smiling_face:

ajsantander: Yep, I'm currently working on it, as well as on all the files marked on the plan. Don't worry, if this is merged, I take notes of what I'm doing, so I'd know exactly where and how to continue with the marked files.

packages/cli/src/models/files/ZosNetworkFile.ts
ajsantander: Awesome!

packages/cli/src/models/files/ZosVersion.ts
ajsantander: We were using "similar" settings. I've updated them here so that they are as similar as possible to the master branch settings.

ajsantander:
@jcarpanelli Thx for the review! Great catch there on seeing that we could type a lot of interfaces in ZosNetworkFile.ts. Notice that I created interfaces for ContractInterface, SolidityLibInterface, ProxyInterface, etc, but I didn't create interfaces for their mappings. Instead, the code directly exposes mappings, like this:

```
  get contracts(): { [contractAlias: string]: ContractInterface } {
    return this.data.contracts || {};
  }
```

I think that's super readable, and better than using an interface that just wraps the mapping. You like?

packages/cli/src/models/status/StatusReport.ts
```diff
@@ -0,0 +1,20 @@
+import { Logger } from 'zos-lib';
+
+export default class StatusReport {
+
+  public expected: string;
```
jcarpanelli: We should start taking advantage of encapsulation. I don't see a reason to use getters and setters if not necessary (i.e., if an attribute is accessed from the outside, I think it's ok to define it as public) but if it's not accessed from the outside, let's define them as `private` or `protected` as appropiate.

jcarpanelli:
@facuspagnuolo we are still struggling with the integration tests here, and I don't have sudo privileges to merge this PR. Could you please give us a :hand: ?

packages/cli/src/models/status/StatusReport.ts
ajsantander: Yes agreed. I am considering encapsulation something to be addressed formally after the first phase of the migration. For now, I'm making everything public unless I see an underscore at the beginning of the member name. I believe that that is a convention in the code. 
Why do I want to do it later? Once the whole codebase is ported to ts, tools can very easily find the usages of a function in Typescript, while it's quite difficult in Javascript.

packages/cli/src/models/status/StatusComparator.ts
```diff
@@ -0,0 +1,102 @@
...
+
+  public onMismatchingDependencyVersion(expected: string, observed: string, { name, version }: ComparedObject): void {
+    this._addReport(expected, observed, `Package version of ${name} does not match`);
+  }
+
+  public onUnregisteredDependency(expected: string, observed: string, { name, package: packageAddress }: ComparedObject): void {
+    this._addReport(expected, observed, `Dependency with name ${name} at address ${packageAddress} is not registered`);
+  }
+
+  public _addReport(expected: string, observed: string, description: string): void {
```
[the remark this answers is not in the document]
ajsantander: Will fix on an upcoming cluster port.

packages/cli/scripts/ts_prepare.sh
```diff
@@ -0,0 +1 @@
+/home/ajs/zos/packages/lib/scripts/ts_prepare.sh
```
facuspagnuolo: was this file committed intentionally?

packages/cli/src/models/files/ZosNetworkFile.ts
```diff
@@ -0,0 +1,418 @@
...
+  customDeploy?: boolean;
+}
+
+interface PackageInterface {
+  address?: string;
+}
+
+interface ProviderInterface {
+  address?: string;
+}
```
facuspagnuolo: is it ok declaring all these interfaces here? shouldn't they be somewhere else?

packages/cli/src/models/files/ZosNetworkFile.ts
```diff
@@ -0,0 +1,418 @@
...
+    {
+      package: proxyPackageName,
+      contract: proxyContractName,
+      address: proxyAddress
+    }: {
+        package: string,
+        contract: string,
+        address: string
+      },
+    fn: (...args: any[]) => any
```
facuspagnuolo: is there a better way to do this to avoid multi-lines here? maybe defining an interface?

packages/cli/src/models/files/index.ts
```diff
@@ -0,0 +1,9 @@
+import * as ZosVersion from './ZosVersion';
```
facuspagnuolo: `import ZosVersion from './...'`

packages/cli/scripts/ts_prepare.sh
ajsantander: Yes, it is a symlink to a script residing in lib that is very useful for porting.

packages/cli/src/models/files/ZosNetworkFile.ts
ajsantander: We are not worrying about this sort of stuff (design architecture decisions) in this phase of the port. The idea is that we port all the primitive types, introduce simple interfaces, and at a latter stage, we decide what to do with them.

packages/cli/src/models/files/ZosNetworkFile.ts
ajsantander: Definitely. I'm taking note and making this change in a future PR of the port.

packages/cli/src/models/files/index.ts
ajsantander: Sorry, what do you mean?

packages/cli/scripts/ts_prepare.sh
facuspagnuolo: and what about the absolute path?

ajsantander: The path is wrong tho... pushing a fix that uses a real file instead.

packages/cli/src/models/files/index.ts
ajsantander: Yes, you're right. The `* as` is not necessary here.

PR #500, opened 2018-12-12
Port project files

### LINKED ISSUE #475: TsMigration, Lib Phase 1: Syntax port of the src folder

jcarpanelli:
Fixes #475 :ok_man:


packages/lib/src/application/App.ts
```diff
@@ -0,0 +1,158 @@
...
+    log.info(`Creating ${packageName} ${contractName} proxy and calling ${callDescription(initMethod, initArgs)}`);
+    return sendTransaction(this.appContract.create, [packageName, contractName, callData], this.txParams);
+  }
+
+  private async _upgradeProxy(proxyAddress: string, packageName: string, contractName: string): Promise<any> {
+    log.info(`Upgrading ${packageName} ${contractName} proxy without running migrations...`);
+    return sendTransaction(this.appContract.upgrade, [proxyAddress, packageName, contractName], this.txParams);
+  }
+
+  private async _upgradeProxyAndCall(proxyAddress: string, contractClass: ContractFactory, packageName: string, contractName: string, initMethodName: string, initArgs: any): Promise<any> {
```
ajsantander: We don't know what type sendTransaction returns, but considering that we know the Solidity code of the contract being called in these funcitons,  we could specify the return type of sendTransaction calls in them. However, it would be time consuming to do now and it'd really just be a nice to have at this point. If you want to do it, great, otherwise, take this as something to consider for later.

packages/lib/src/project/AppProject.ts
```diff
@@ -0,0 +1,154 @@
...
+    return this.version;
+  }
+
+  // TODO: Testme
+  public async getImplementation({ packageName, contractName }: { contractName: string, packageName?: string }): Promise<string> {
+    return this.app.getImplementation(packageName || this.name, contractName);
+  }
+
+  // TODO: Testme
+  public async createContract(contractClass: ContractFactory, { packageName, contractName, initMethod, initArgs }: { packageName?: string, contractName?: string, initMethod?: string, initArgs?: string[] } = {}): Promise<ContractWrapper> {
```
ajsantander: The same return type is defined multiple times. Maybe define an interface?

packages/lib/src/project/AppProject.ts
```diff
@@ -0,0 +1,154 @@
...
+
+const DEFAULT_NAME: string = 'main';
+const DEFAULT_VERSION: string = '0.1.0';
+
+export default class AppProject extends BasePackageProject {
+  private name: string;
+  private app: App;
+
+  // REFACTOR: Evaluate merging this logic with CLI's ProjectDeployer classes
+  public static async fetchOrDeploy(name = DEFAULT_NAME, version = DEFAULT_VERSION, txParams = {}, { appAddress, packageAddress }: { appAddress?: string, packageAddress?: string } = {}): Promise<AppProject> | never {
```
ajsantander: Why not type name and version? Type very nice.

packages/lib/src/project/AppProject.ts
```diff
@@ -0,0 +1,154 @@
...
+  }
+
+  // TODO: Testme
+  public async createContract(contractClass: ContractFactory, { packageName, contractName, initMethod, initArgs }: { packageName?: string, contractName?: string, initMethod?: string, initArgs?: string[] } = {}): Promise<ContractWrapper> {
+    if (!contractName) contractName = contractClass.contractName;
+    if (!packageName) packageName = this.name;
+    return this.app.createContract(contractClass, packageName, contractName, initMethod, initArgs);
+  }
+
+  public async createProxy(contractClass: ContractFactory, { packageName, contractName, initMethod, initArgs }: { packageName?: string, contractName?: string, initMethod?: string, initArgs?: string[] } = {}): Promise<ContractWrapper> {
```
ajsantander: As comment above, but for the object with the properties: packageName, contractName, etc.

packages/lib/src/project/BasePackageProject.ts
```diff
@@ -0,0 +1,70 @@
...
+    await directory.unsetImplementation(contractName);
+  }
+
+  public async registerImplementation(contractName: string, { address }: { address: string }): Promise<void> {
+    log.info(`Registering implementation of ${contractName} at ${address} in directory...`);
+    const directory: ImplementationDirectory = await this.getCurrentDirectory();
+    await directory.setImplementation(contractName, address);
+  }
+
+  public async getCurrentDirectory(): Promise<any> {
```
ajsantander: These should return never. I suppose that you want to keep the Promise function signature because the methods are intended to be overriden... Perhaps you could consider using the `abstract` keyword of typescript? Anyway, may be overkill to do now, but I'd at least leave a comment to do this later.

packages/lib/src/project/AppProject.ts
jcarpanelli: They are defined. But the definition of variable types inside an object is a bit annoying and confusing
```javascript
{ packageName, contractName, initMethod, initArgs }: { packageName?: string, contractName?: string, initMethod?: string, initArgs?: string[] }
```

packages/lib/src/project/AppProject.ts
ajsantander: Sorry, I meant the composite object params.

PR #507, opened 2018-12-13
Polish truffle utils

facuspagnuolo:


packages/cli/src/models/compiler/Compiler.js
```diff
@@ -1,13 +1,13 @@
 import { exec } from 'child_process'
 import { FileSystem, Contracts, Logger } from 'zos-lib'
-import TruffleConfig from '../initializer/truffle/TruffleConfig'
+import Truffle from '../initializer/truffle/Truffle'
```
jcarpanelli: I think that using an external library name (Truffle) for a zos local class/object could be somewhat confusing. I don't have any other name in mind, though.

packages/lib/src/artifacts/Contracts.ts
```diff
@@ -69,7 +65,7 @@ export default class Contracts {
   }
 
   public static setLocalContractsDir(dir: string): void {
-    Contracts.buildDir = dir;
+    Contracts.contractsDir = dir;
```
jcarpanelli: Nice catch!

packages/lib/test/src/artifacts/Contracts.test.js
```diff
@@ -0,0 +1,50 @@
+'use strict'
+require('../../setup')
+
+import Contracts from '../../../src/artifacts/Contracts'
+
+contract('Contracts', () => {
```
jcarpanelli: We've been using regular functions over arrow functions in tests blocks. If there isn't a scope-related reason to move from one to another, let's try to keep consistency.
```suggestion
contract('Contracts', function() {
```

Issue #483, opened 2018-12-10
TsMigration, Cli Phase 1: Syntax port of src folder


ajsantander:
Following the migration plan from #434

PR #516, opened 2018-12-18
Ports the src/models/errors folder of cli to Ts

ajsantander:
Part of solving #483


packages/cli/src/models/errors/ErrorHandler.ts
```diff
@@ -0,0 +1,27 @@
+import { Logger } from 'zos-lib';
+import ScriptError from './ScriptError';
+
+const log = new Logger('Error');
```
jcarpanelli: I'm still typing constant variables that are not inside functions, as our rule was only not to type function bodies :stuck_out_tongue:

packages/cli/src/models/errors/ScriptError.ts
```diff
@@ -0,0 +1,11 @@
+export default class ScriptError extends Error {
+
+  public stack: any;
+  public cb: () => void;
+
+  constructor({ message, stack }: { message: string, stack: any }, cb: () => void) {
```
jcarpanelli: nit: `stack` will always be a `string`.

packages/cli/src/models/files/ZosNetworkFile.ts
```diff
@@ -374,15 +374,7 @@ export default class ZosNetworkFile {
...
-    {
-      package: proxyPackageName,
-      contract: proxyContractName,
-      address: proxyAddress
-    }: {
-        package: string,
-        contract: string,
-        address: string
-      },
+    { package: proxyPackageName, contract: proxyContractName, address: proxyAddress }: ProxyInterface,
```
jcarpanelli: :heart:

packages/cli/src/models/errors/ErrorHandler.ts
ajsantander: Yeah, I saw this on your PRs. You are typing this stuff, I'm not. I believe that the code looks prettier with inference in these cases. Your call...

packages/cli/src/models/errors/ScriptError.ts
ajsantander: Fixed.

PR #517, opened 2018-12-18
Port Verifier, Session and ProjectDeployer

jcarpanelli:
Related to #483 
Ports:
- [x] Verifier
- [x] ProjectDeployer
- [x] Session


packages/cli/src/models/network/ProjectDeployer.ts
```diff
@@ -0,0 +1,137 @@
...
+  constructor(controller: NetworkController, requestedVersion: string) {
+    this.controller = controller;
+    this.packageFile = controller.packageFile;
+    this.networkFile = controller.networkFile;
+    this.txParams = controller.txParams;
+    this.requestedVersion = requestedVersion;
+  }
+}
+
+class BasePackageProjectDeployer extends BaseProjectDeployer {
```
ajsantander: :heart_eyes:

packages/cli/src/models/network/ProjectDeployer.ts
```diff
@@ -0,0 +1,137 @@
...
+  get packageAddress(): string {
+    return this.controller.packageAddress;
+  }
+
+  protected _tryRegisterPartialDeploy({ thepackage, directory }: PartialDeploy): void {
+    if (thepackage) this._registerPackage(thepackage);
+    if (directory) this._registerVersion(this.requestedVersion, directory);
+  }
+
+  protected _registerPackage({ address }: { address: string }): void {
```
ajsantander: Probably not worth doing anything atm, but note that `{ address }` is the de-structuring of an incoming object that has a property "address" (and could have any other properties). OTOH, the typing `{ address: string }` is a type that defines an object that _only_ has the property "address". 

I'm not a 100% sure about this, but let's just keep this in the back of our heads for now.

packages/cli/src/models/network/ProjectDeployer.ts
```diff
@@ -0,0 +1,137 @@
...
+    return this._run((existingAddresses: ExistingAddresses) => (
+      AppProject.fromSimpleProject(simpleProject, this.requestedVersion, existingAddresses)
+    ));
+  }
+
+  get appAddress(): string {
+    return this.controller.appAddress;
+  }
+
+  public async _run(createProjectFn: CreateProjectFn): Promise<AppProject | never> {
```
ajsantander: private?

jcarpanelli: nice catch :slightly_smiling_face:

packages/cli/src/models/network/ProjectDeployer.ts
jcarpanelli: Already discussed it offline.

PR #519, opened 2018-12-19
Make `lib/bin/zos-cli.js` executable after transpiling to ts

jcarpanelli:
Closes https://github.com/zeppelinos/zos/issues/518.
Just running a `chmod 755 ./lib/bin/zos-cli.js` after running `npm run compile-ts`. It seems that tsc doesn't keep file system permissions, as mentioned here: https://github.com/Microsoft/TypeScript/issues/16257#issuecomment-319604800
@facuspagnuolo, I noticed that `src/bin/errors.js` and `src/bin/program.js` have also execute permissions, but I don't see why. Any hints?

### DIFF
--- a/packages/cli/package.json
+++ b/packages/cli/package.json
@@ -7,7 +7,7 @@
   "license": "MIT",
   "scripts": {
     "compile-ts": "rm -rf lib && tsc",
-    "prepare": "npm run compile-ts",
+    "prepare": "npm run compile-ts && chmod 755 ./lib/bin/zos-cli.js",
     "test": "./scripts/test.sh",
     "gen-docs": "babel-node docs/bin/docs.js",
     "watch": "tsc -w",


packages/cli/package.json
```diff
@@ -7,7 +7,7 @@
   "license": "MIT",
   "scripts": {
     "compile-ts": "rm -rf lib && tsc",
-    "prepare": "npm run compile-ts",
+    "prepare": "npm run compile-ts && chmod 755 ./lib/bin/zos-cli.js",
```
facuspagnuolo: shouldn't this be part of the `compile-ts` script? I guess every time we compile TS we want that to happen, not only during installation right?

jcarpanelli: Strictly speaking, `compile-ts` should only **compile** (transpile :confused:) ts code to js code. IMO, `prepare` should manage both compilation plus any other task that leaves the  `cli` **prepared** to run.

PR #520, opened 2018-12-19
Ports cli's compiler folder to Ts

ajsantander:
Part of solving #483


packages/cli/src/models/compiler/Compiler.ts
```diff
@@ -0,0 +1,42 @@
...
+import { FileSystem, Contracts, Logger } from 'zos-lib';
+import Truffle from '../initializer/truffle/Truffle';
+import SolidityProjectCompiler from './solidity/SolidityProjectCompiler';
+import { CompilerOptions } from './solidity/SolidityContractsCompiler';
+
+const log = new Logger('Compiler');
+
+export default {
+  async call(): Promise<void> {
+    return Truffle.isTruffleProject()
```
facuspagnuolo: I guess the return statement is useless here

ajsantander: Will remove on master.

PR #521, opened 2018-12-19
Ts cli/port bin folder

jcarpanelli:
Related to #483 

Ports:
- [x] zos-cli.ts
- [x] program.ts
- [x] options.ts
- [x] errors.ts


packages/cli/src/bin/errors.ts
```diff
@@ -0,0 +1,12 @@
+import ErrorHandler from '../models/errors/ErrorHandler';
+
+export default function registerErrorHandler(program): void {
```
ajsantander: Why program paramitir no type =(

packages/cli/src/bin/options.ts
```diff
@@ -1,22 +1,22 @@
-import program from 'commander'
-import { DEFAULT_TX_TIMEOUT } from '../models/network/Session'
+import program, { Command } from 'commander';
```
ajsantander: :heart_eyes:

packages/cli/src/bin/program.ts
```diff
@@ -0,0 +1,40 @@
...
+import chalk from 'chalk';
+import program from 'commander';
+import { Logger } from 'zos-lib';
+import commands from '../commands';
+import registerErrorHandler from './errors';
+
+require('./options');
+
+// TS-TODO: Decide whether this should be an interface or a type
+interface Command {
```
ajsantander: Hmmm... We have a `Command` interface for our commands and a `Command` interface for commander. Maybe, we should 1) Rename this interface to something like `ZosCommand` and when importing commander's command, do it in a way that the type needs to be referenced with the namespace, like this: `commander.Command`

jcarpanelli: I would prefer to import change our custom `Command` interface name rather to use `commander.Command`. Thx!

PR #523, opened 2018-12-19
Port Dependency, TestHelper and ValidationLogger in the cli package to Ts

ajsantander:
Part of solving #483 

This port was easy for Dependency and TestHelper. OTOH, it was very difficult for ValidationLogger, because of the following reasons:
 
1) ValidationLogger uses interfaces that where previously internal to the lib package, such as Operation, StorageLayoutInfo, and some others. These had to be exposed in the lib package's main index.ts file, so that the cli package could use them. This hasn't happened before in my experience of porting Js to Ts in this project.

2) After introducing @types/solc.d.ts in #520, it is evident that some of the interfaces defined in the AST related files of the lib package are redundant, and could be replaced by interfaces from solc.d.ts if the declaration file was moved to the lib package.

3) It was very interesting to see that AST related code in the lib package was not entirely ported correctly. This became evident AFTER porting ValidationLogger in either of 2 ways: Static code analysis showing interface incompatibilities or ValidationLogger's tests not passing in cli.

4) Code needing to be altered in the lib package because tests don't pass in the cli package means that something is not quite right with the tests. Either some tests need to be moved to the lib package, or the lib test coverage is not very good _by itself_.


packages/cli/src/models/dependency/Dependency.ts
```diff
@@ -1,58 +1,67 @@
...
+  private _networkFiles: { [network: string]: ZosNetworkFile };
+  private _packageFile: ZosPackageFile;
+
+  public static fromNameWithVersion(nameAndVersion: string): Dependency {
+    const [name, version] = nameAndVersion.split('@');
+    return new this(name, version);
   }
 
-  static satisfiesVersion(version, requirement) {
+  public static satisfiesVersion(version: string | semver.SemVer, requirement: string | semver.Range): boolean {
```
jcarpanelli: Can we import `SemVer` and `Range` in `L3` like this?
```ts
import semver, { SemVer, Range } from 'semver';
```

ajsantander: Yes, we can do that, although I think typing 3rd party stuff with namespaces is clearer. Will add a note to my list of things to do after the merge. I'll ping you for discussing this.

PR #524, opened 2018-12-20
Integration tests fix

jcarpanelli:
This PR attempts to fix TravisCI integration tests' builds. Long story short, running `npm install` just after running `lerna bootstrap` did the trick. 
This issue may be related to how lerna manages `npm i` runs in a continuous integration environment (see: https://github.com/lerna/lerna/pull/1360#discussion_r179851072). Basically, it runs a `npm ci` instead of a `npm i`.

I also implemented a small fix related to a file import in `packages/cli` that was breaking `packages/vouching` tests.

### DIFF
--- a/packages/cli/src/models/files/index.ts
+++ b/packages/cli/src/models/files/index.ts
@@ -1,9 +1,10 @@
-import ZosVersion from './ZosVersion';
+import { ZOS_VERSION, checkVersion } from './ZosVersion';
 import ZosNetworkFile from './ZosNetworkFile';
 import ZosPackageFile from './ZosPackageFile';
 
 export default {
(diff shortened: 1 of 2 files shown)


packages/cli/src/models/files/index.ts
```diff
@@ -1,9 +1,10 @@
-import ZosVersion from './ZosVersion';
+import { ZOS_VERSION, checkVersion } from './ZosVersion';
```
ajsantander: This is a more elegant fix to what I did in https://github.com/zeppelinos/zos/pull/525 to fix this problem. So when it comes to merge that PR, we should resolve the conflict favoring this solution.

PR #526, opened 2018-12-20
Ports some files in the local and network folders of cli to Ts

ajsantander:
Part of solving #482


packages/lib/src/proxy/Proxy.ts
```diff
@@ -15,9 +15,9 @@ export default class Proxy {
   private txParams: any;
   public address: string;
 
-  public static at(address: string, txParams: any = {}): Proxy {
+  public static at(contractOrAddress: string | any, txParams: any = {}): Proxy {
```
jcarpanelli: nit: both contract name and address are strings

packages/cli/src/models/network/NetworkController.ts
```diff
@@ -105,179 +115,176 @@ export default class NetworkController {
...
-      this.networkFile.contractAliasesMissingFromPackage().map(contractAlias => this.unsetContract(contractAlias))
-    )
+      this.networkFile.contractAliasesMissingFromPackage().map((contractAlias) => this.unsetContract(contractAlias))
+    );
   }
 
-  //Contract model
-  async unsetContract(contractAlias) {
+  // Contract model
+  public async unsetContract(contractAlias: string): Promise<void | never> {
```
jcarpanelli: nit: private.

packages/cli/src/models/network/NetworkController.ts
```diff
@@ -105,179 +115,176 @@ export default class NetworkController {
...
+      error.message = `Removal of ${contractAlias} failed with error: ${error.message}`;
+      throw error;
     }
   }
 
-  //DeployerController || Contract model
-  validateContracts(contracts, buildArtifacts) {
-    return _.every(contracts.map(([contractAlias, contractClass]) => 
+  // DeployerController || Contract model
+  public validateContracts(contracts: Array<[string, ContractFactory]>, buildArtifacts: BuildArtifacts): boolean {
```
jcarpanelli: nit: private.

packages/cli/src/models/network/NetworkController.ts
```diff
@@ -105,179 +115,176 @@ export default class NetworkController {
...
+    return _.every(contracts.map(([contractAlias, contractClass]) =>
       this.validateContract(contractAlias, contractClass, buildArtifacts))
-    )
+    );
   }
 
-  //DeployerController || Contract model
-  validateContract(contractAlias, contractClass, buildArtifacts) {
+  // DeployerController || Contract model
+  public validateContract(contractAlias: string, contractClass: ContractFactory, buildArtifacts: BuildArtifacts): boolean {
```
jcarpanelli: nit: private.

packages/cli/src/models/network/NetworkController.ts
```diff
@@ -307,101 +314,102 @@ export default class NetworkController {
...
     }
-    return !this.networkFile.hasSameBytecode(contractAlias, contractClass)
+    return !this.networkFile.hasSameBytecode(contractAlias, contractClass);
   }
 
-  //Contract model
-  isLocalContract(contractAlias) {
-    return this.packageFile.hasContract(contractAlias)
+  // Contract model
+  public isLocalContract(contractAlias: string): boolean {
```
jcarpanelli: nit: private.

packages/cli/src/models/network/NetworkController.ts
```diff
@@ -307,101 +314,102 @@ export default class NetworkController {
...
-  get isLightweight() {
+  // DeployerController
+  get isLightweight(): boolean {
     return this.packageFile.isLightweight && !this.appAddress;
   }
 
-  //DeployerController
-  getDeployer(requestedVersion) {
+  // DeployerController
+  public getDeployer(requestedVersion: string): SimpleProjectDeployer | AppProjectDeployer {
```
jcarpanelli: nit: private

packages/cli/src/models/network/NetworkController.ts
```diff
@@ -416,240 +424,240 @@ export default class NetworkController {
...
       implementation: implementationAddress
-    })
+    });
     return proxyInstance;
   }
 
-  //Proxy model
-  checkInitialization(contractClass, calledInitMethod, calledInitArgs) {
+  // Proxy model
+  public checkInitialization(contractClass: ContractFactory, calledInitMethod: string, calledInitArgs: string[]): void {
```
jcarpanelli: nit: private.

packages/cli/src/models/network/NetworkController.ts
```diff
@@ -416,240 +424,240 @@ export default class NetworkController {
...
     await allPromisesOrError(
       _.map(this.packageFile.dependencies, (version, dep) => this.deployDependencyIfNeeded(dep, version))
-    )
+    );
   }
 
-  //DependencyController
-  async deployDependencyIfNeeded(depName, depVersion) {
+  // DependencyController
+  public async deployDependencyIfNeeded(depName: string, depVersion: string): Promise<void | never> {
```
jcarpanelli: nit: private.

packages/cli/src/models/network/NetworkController.ts
```diff
@@ -416,240 +424,240 @@ export default class NetworkController {
...
-      throw error
+      error.message = `Failed deployment of dependency ${depName} with error: ${error.message}`;
+      throw error;
     }
   }
 
-  //DependencyController
-  async handleDependenciesLink() {
+  // DependencyController
+  public async handleDependenciesLink(): Promise<void> {
```
jcarpanelli: nit: private.

packages/cli/src/models/network/NetworkController.ts
```diff
@@ -416,240 +424,240 @@ export default class NetworkController {
...
-      _.map(this.networkFile.dependenciesNamesMissingFromPackage(), dep => this.unlinkDependency(dep))
-    ))
+      _.map(this.networkFile.dependenciesNamesMissingFromPackage(), (dep) => this.unlinkDependency(dep))
+    ));
   }
 
-  //DependencyController
-  async unlinkDependency(depName) {
+  // DependencyController
+  public async unlinkDependency(depName: string): Promise<void | never> {
```
jcarpanelli: nit: private

packages/cli/src/models/network/NetworkController.ts
```diff
@@ -416,240 +424,240 @@ export default class NetworkController {
...
     } catch (error) {
-      throw Error(`Failed to unlink dependency ${depName} with error: ${error.message}`)
+      throw Error(`Failed to unlink dependency ${depName} with error: ${error.message}`);
     }
   }
 
-  //DependencyController
-  async linkDependency(depName, depVersion) {
+  // DependencyController
+  public async linkDependency(depName: string, depVersion: string): Promise<boolean | never> {
```
jcarpanelli: nit: private.

packages/cli/src/models/network/NetworkController.ts
```diff
@@ -416,240 +424,240 @@ export default class NetworkController {
...
     try {
       if (this.networkFile.dependencyHasMatchingCustomDeploy(depName)) {
         log.info(`Using custom deployment of ${depName}`);
         const depInfo = this.networkFile.getDependency(depName);
-        return await this.project.setDependency(depName, depInfo.package, depInfo.version);
+        return await (<AppProject>this.project).setDependency(depName, depInfo.package, depInfo.version);
```
jcarpanelli: Mmmh. Having dependencies linked is possible in a `SimpleProject`. I'm not sure if this cast is accurate.

packages/lib/src/proxy/Proxy.ts
ajsantander: Actually not. This was quite tricky and something pointed out by @facuspagnuolo 

Proxy uses the following function from utils/Addresses.ts:

```
export function toAddress(contractOrAddress: string | any): string {
  if (_.isEmpty(contractOrAddress)) throw Error(`Contract or address expected`);
  else if (_.isString(contractOrAddress)) return contractOrAddress;
  else return contractOrAddress.address;
}
```

packages/cli/src/models/network/NetworkController.ts
ajsantander: Is it private because you know it should be used that way or because of something else? So far I've been using private only for Js function that start with an underscore.

packages/cli/src/models/network/NetworkController.ts
ajsantander: Same as above.

packages/cli/src/models/network/NetworkController.ts
ajsantander: Wow! Good catch =D

packages/cli/src/models/network/NetworkController.ts
ajsantander: That's great. However, I think we should worry about encapsulation at a later stage of the migration.

PR #537, opened 2018-12-21
Port src/scripts to TS.

jcarpanelli:
Related to #483

Ports:
- [x] init
- [x] add
- [x] session
- [x] bump
- [x] remove
- [x] push
- [x] update
- [x] publish
- [x] create
- [x] set-admin
- [x] freeze
- [x] check
- [x] link
- [x] status
- [x] unlink
- [x] verify

Others:
- [x] Refactor `src/scripts/interfaces.ts` taking advantage of interfaces inheritance


packages/cli/src/commands/session.ts
```diff
@@ -16,7 +16,7 @@ const register: (program: Command) => Command = (program) => program
 
 function action(options: Command): void {
   const { network, from, timeout, close, expires } = options;
-  session({ network, from, timeout, close, expires });
+  session({ network, from, timeout: <number> timeout, close, expires: <number> expires });
```
ajsantander: This cast is interesting. 
A small curiosity (outside of the scope of this port): What happens if the user inputs an invalid number for timeout?

packages/cli/src/commands/verify.ts
```diff
@@ -8,7 +8,6 @@ const name: string = 'verify';
 const signature: string = `${name} <contract-alias>`;
 const description: string = 'verify a contract with etherchain. Provide a contract name.';
 
-
```
ajsantander: Linter <3

packages/cli/src/scripts/add.ts
```diff
@@ -0,0 +1,15 @@
+import ControllerFor from '../models/local/ControllerFor';
+import ZosPackageFile from '../models/files/ZosPackageFile';
+import { AddParams } from './interfaces';
+
+export default function add({ contractsData, packageFile }: AddParams): void | never {
```
ajsantander: :heart_eyes: :heart_eyes: :heart_eyes: :heart_eyes:

packages/cli/src/scripts/add.ts
```diff
@@ -0,0 +1,15 @@
...
+  if (contractsData.length === 0) throw new Error('At least one contract name must be provided to add.');
+  contractsData = contractsData.map((data) => typeof(data) === 'string' ? { name: data, alias: data } : data);
+
+  const controller = ControllerFor(packageFile);
```
ajsantander: `controller` would be a LocalController here right? It may be nice to add the explcity type for easier reading of the code.

packages/cli/src/scripts/bump.ts
```diff
@@ -0,0 +1,11 @@
+import stdout from '../utils/stdout';
+import ControllerFor from '../models/local/ControllerFor';
+import { BumpParams } from './interfaces';
+
+export default async function bumpVersion({ version, packageFile }: BumpParams): Promise<void | never> {
```
ajsantander: Question: do the commands use these interfaces like BumpParams?

packages/cli/src/scripts/interfaces.ts
```diff
@@ -0,0 +1,39 @@
+import ZosPackageFile from '../models/files/ZosPackageFile';
+
+export interface InitParams {
```
ajsantander: Could some of these be non optional?

packages/cli/src/utils/stdout.ts
```diff
@@ -2,9 +2,9 @@ const state: { silent: boolean } = {
   silent: false
 };
 
-export default function log(): void {
+export default function log(...args: any[]): void {
   if (!state.silent && process.env.NODE_ENV !== 'test') {
-    console.log(...arguments);
```
ajsantander: Nice! `arguments` always seems too mysterious to me.

packages/cli/test/scripts/bump.test.js
```diff
@@ -1,9 +1,9 @@
 'use strict'
 require('../setup')
 
-import add from '../../src/scripts/add.js';
```
ajsantander: Bummer you had to change these too... great!

packages/cli/src/scripts/bump.ts
jcarpanelli: Nope. There might be some shared attributes, though, there are not exact matches.

packages/cli/src/utils/stdout.ts
jcarpanelli: Yeap! Typescript won't let me call `stdout` with params when the function definition doesn't have parameters defined :slightly_smiling_face:

packages/cli/src/scripts/interfaces.ts
jcarpanelli: Yes, most of them. I was planning to refactor these interfaces at the end of the port :smile:

packages/cli/src/scripts/add.ts
jcarpanelli: We thought the same thing! I was planning to type all `ControllerFor`s, but as we talked offline, I'll remove both local and network `ControllerFor` ASAP.

packages/cli/src/commands/session.ts
jcarpanelli: I'm not sure what would happen, but I think that typescript *tries* to cast the variable if possible. If not, it leaves it as it is. OTOH, speaking about what's the current zos behaviour, I'm pretty sure that the option will be ignored.

jcarpanelli:
@ajsantander Would you mind reviewing this again? :slightly_smiling_face:

PR #541, opened 2018-12-26
Fixes a bug on encodeCall

### LINKED ISSUE #421: Initialization argument encoding issues.

```shell
1) npm i  
2) zos session -n local  
3) zos push -d  

```

### Software versions:
`macOS 10.14.1`  

`npm version  
{ test: '1.0.0',  
  npm: '6.4.1',  
  ares: '1.14.0',  
  cldr: '33.1',  
  http_parser: '2.8.0',  
  icu: '62.1',  
  modules: '67',  
  napi: '3',  
  nghttp2: '1.34.0',  
  node: '11.1.0',  
  openssl: '1.1.0i',  
  tz: '2018e',  
  unicode: '11.0',  
  uv: '1.23.2',  
  v8: '7.0.276.32-node.7',  
  zlib: '1.2.11' }`  

`Solidity v0.4.24 (solc-js)`  
`Ganache GUI version v1.2.2`  
`ZeppelinOS version 2.0.1`

ajsantander:
Fix #421 

Lib's encodeCall helper was failing for any hexadecimal strings that contained the character 'e'. It was interpreting these as an exponential number (e.g. `1.5e9`) and interpreting the hexadecimal value as a decimal.

Note that this PR also includes some tests for the helper, and some suggestions to (1) further improve it and (2) increase its test coverage.

### DIFF
--- a/packages/lib/src/helpers/encodeCall.ts
+++ b/packages/lib/src/helpers/encodeCall.ts
@@ -1,15 +1,35 @@
 import abi from 'ethereumjs-abi';
 import BN from 'bignumber.js';
 
-function formatValue(value: any): string {
-  if (typeof(value) === 'number' || BN.isBigNumber(value)) return value.toString();
-  else if (typeof(value) === 'string' && value.match(/\d+(\.\d+)?e(\+)?\d+/)) return (new BN(value)).toString(10);
-  else return value;
+export function formatValue(value: any): string | void {
+
+  // Numbers.
+  if(typeof(value) === 'number') {
+    if(value % 1 === 0) return value.toString(10);
+    else throw new Error('Floating point numbers are not supported on parameter encoding.');
+  }
+
+  // Big numbers.
+  if(BN.isBigNumber(value)) return value.toString();
+
+  // Strings.
+  if(typeof(value) === 'string') {
+    const hasHexRadix = value.indexOf('0x') !== -1 || value.indexOf('0X') !== -1;
+
+    // Numeric strings with exponents, e.g. '1.5e9'.
+    if(!hasHexRadix && value.match(/\d+(\.\d+)?e(\+)?\d+/)) {
+      return (new BN(value)).toString(10);
+    }
+  }
+
+  // Something else.
+  return value;
 }
 
 export default function encodeCall(name: string, args: string[] = [], rawValues: any[] = []): string {
-  const values: string[] = rawValues.map(formatValue);
+  const values: string[] = <string[]>rawValues.map(formatValue);
   const methodId: string = abi.methodID(name, args).toString('hex');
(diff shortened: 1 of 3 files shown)


packages/lib/src/helpers/encodeCall.ts
```diff
@@ -1,15 +1,35 @@
...
 
-function formatValue(value: any): string {
-  if (typeof(value) === 'number' || BN.isBigNumber(value)) return value.toString();
-  else if (typeof(value) === 'string' && value.match(/\d+(\.\d+)?e(\+)?\d+/)) return (new BN(value)).toString(10);
-  else return value;
+export function formatValue(value: any): string | void {
+
+  // Numbers.
+  if(typeof(value) === 'number') {
+    if(value % 1 === 0) return value.toString(10);
```
spalladino: How about when `value` is a very large number? It seems that toString outputs scientific notation:
```
> (2**255).toString(10)
'5.78960446186581e+76'
```

Will the etherumjs abi encoder handle the number in scientific notation?

packages/lib/src/helpers/encodeCall.ts
```diff
@@ -1,15 +1,35 @@
...
+    else throw new Error('Floating point numbers are not supported on parameter encoding.');
+  }
+
+  // Big numbers.
+  if(BN.isBigNumber(value)) return value.toString();
+
+  // Strings.
+  if(typeof(value) === 'string') {
+    const hasHexRadix = value.indexOf('0x') !== -1 || value.indexOf('0X') !== -1;
```
spalladino: Shouldn't we check that the 0x radix is at the beginning of the string (maybe ignoring whitespace)? Otherwise "hell0x" whould test positive for having a 0x radix.

packages/lib/src/helpers/encodeCall.ts
```diff
@@ -1,15 +1,35 @@
...
+
+    // Numeric strings with exponents, e.g. '1.5e9'.
+    if(!hasHexRadix && value.match(/\d+(\.\d+)?e(\+)?\d+/)) {
```
spalladino: Aren't we doing the same mistake as before here, in trying to match just a part of the string, instead of all of it? Shouldn't replace the regex by something like `/^\s*\d+(\.\d+)?e(\+)?\d+\s*$/`? (note the start and end of string anchors, and the optional whitespace)

packages/lib/src/helpers/encodeCall.ts
spalladino: Can we add a test for this, to make sure that "hell0xa" is handled properly as a string?

ajsantander:
Closing this in favour of #543

PR #543, opened 2018-12-27
Refactor encodeCall

### LINKED ISSUE #421: Initialization argument encoding issues.

ajsantander:
Fix #421 

Lib's encodeCall helper was failing for any hexadecimal strings that contained the character 'e'. It was interpreting these as scientific notation numbers (e.g. 1.5e9) and seeing their hexadecimal value parsed a decimal instead =S.

This PR goes quite far beyond simply fixing the bug; It rewrites the helper from the ground up, with type/value pair validation and adds tests for this particular helper. The validation involves comparing the expected type with its actual value and making sure that they are compatible. I.e. 'uint256' is not compatible with the value 'hello world', and 'address' is not compatible with the value '0x123'.

### DIFF
--- a/packages/lib/src/helpers/encodeCall.ts
+++ b/packages/lib/src/helpers/encodeCall.ts
@@ -11,12 +11,15 @@ import _ from 'lodash';
 const ERROR_MESSAGE_PREFIX = 'Encoding error';
 const ERROR_MESSAGE = (type: string, value: any) => `${ERROR_MESSAGE_PREFIX} for type ${type} and value ${value.toString()}`;
 
-export default function encodeCall(name: string, types: string[] = [], rawValues: any[] = []): string {
-  if(types.length !== rawValues.length) throw new Error(ERROR_MESSAGE_PREFIX + '. Supplied number of types and values do not match.');
-  const values = _.zipWith(types, rawValues, parseTypeValuePair);
+export default function encodeCall(name: string, types: string[] = [], values: any[] = []): string {
   const methodId = abi.methodID(name, types).toString('hex');
   const params = abi.rawEncode(types, values).toString('hex');
   return '0x' + methodId + params;
+}
+
+export function parseCallValues(types: string[] = [], rawValues: any[] = []): any[] | never {
+  if(types.length !== rawValues.length) throw new Error(ERROR_MESSAGE_PREFIX + '. Supplied number of types and values do not match.');
+  return _.zipWith(types, rawValues, parseTypeValuePair);
 }
 
 export function parseTypeValuePair(type: string, rawValue: any): any | never {
--- a/packages/lib/src/helpers/encodeCall.ts
+++ b/packages/lib/src/helpers/encodeCall.ts
@@ -30,18 +30,18 @@ export function parseTypeValuePair(type: string, rawValue: any): string | never 
   else if(type === 'string') {
     return rawValue; // Validated by ethereumjs-abi.
   }
-  else if(type.startsWith('bytes')) {
+  else if(type.startsWith('bytes') && !type.includes('[]')) {
     return parseBytes(rawValue);
   }
-  else if(type.startsWith('uint')) {
+  else if(type.startsWith('uint') && !type.includes('[]')) {
     return parseNumber(rawValue, true);
   }
-  else if(type.startsWith('int')) {
+  else if(type.startsWith('int') && !type.includes('[]')) {
     return parseNumber(rawValue, false);
   }
   else {
     // TODO: parse remianing types: fixed, arrays, etc.
-    return rawValue.toString();
+    return rawValue;
   }
 }
 
--- a/packages/lib/src/helpers/encodeCall.ts
+++ b/packages/lib/src/helpers/encodeCall.ts
@@ -3,62 +3,68 @@ import * as util from 'ethereumjs-util';
 import BN from 'bignumber.js';
 import _ from 'lodash';
 
-const ERROR_MESSAGE_PREFIX = 'Invalid parameter';
-const ERROR_MESSAGE_POSITIVE_NUMBER = (value) => `${ERROR_MESSAGE_PREFIX} number "${value}" must be positive.`;
-const ERROR_MESSAGE_INTEGER_NUMBER = (value) => `${ERROR_MESSAGE_PREFIX} number "${value}" must be an integer.`;
-const ERROR_MESSAGE_NUMBER = (value) => `${ERROR_MESSAGE_PREFIX} number "${value}".`;
-const ERROR_MESSAGE_ADDRESS = (value) => `${ERROR_MESSAGE_PREFIX} address "${value}".`;
-const ERROR_MESSAGE_CHECKSUM_ADDRESS = (value) => `${ERROR_MESSAGE_PREFIX} address "${value}" checksum fails.`;
-const ERROR_MESSAGE_BYTES = (value) => `${ERROR_MESSAGE_PREFIX} bytes "${value}".`;
-const ERROR_MESSAGE_STRING = (value) => `${ERROR_MESSAGE_PREFIX} string "${value}".`;
+const ERROR_MESSAGE_PREFIX = 'Encoding error';
+const ERROR_MESSAGE = (type: string, value: any) => `${ERROR_MESSAGE_PREFIX} for type ${type} and value ${value.toString()}`;
 
 export default function encodeCall(name: string, types: string[] = [], rawValues: any[] = []): string {
-  if(types.length !== rawValues.length) throw new Error('Supplied number of types and values do not match.');
+  if(types.length !== rawValues.length) throw new Error(ERROR_MESSAGE_PREFIX + '. Supplied number of types and values do not match.');
   const values =_.zipWith(types, rawValues, (type: string, value: any) => parseTypeValuePair(type, value));
   const methodId = abi.methodID(name, types).toString('hex');
   const params = abi.rawEncode(types, values).toString('hex');
   return '0x' + methodId + params;
 }
 
-export function parseTypeValuePair(type: string, rawValue: any): string | never {
-  if(type === 'address') return parseAddress(rawValue);
+export function parseTypeValuePair(type: string, rawValue: any): string | string[] | never {
+  // Deal with arrays recursively.
+  if(/^[^\[]+\[.*\]$/.test(type)) {
+    if(typeof(rawValue) === 'string') rawValue = rawValue.split(',');
+    if(rawValue.length === 0) return [];
+    const baseType = type.slice(0, type.lastIndexOf('[')); // Remove array part.
+    // TODO: validate length on fixed size arrays.
+    return _.map(rawValue, (rawValueElement) => parseTypeValuePair(baseType, rawValueElement));
+  }
+  // Singles.
+  if(type === 'address') return parseAddress(type, rawValue);
   else if(type === 'string') return rawValue; // Validated by ethereumjs-abi.
-  else if(type.startsWith('bytes') && !type.includes('[]')) return parseBytes(rawValue);
-  else if(type.startsWith('uint') && !type.includes('[]')) return parseNumber(rawValue, true);
-  else if(type.startsWith('int') && !type.includes('[]')) return parseNumber(rawValue, false);
-  else return rawValue; // TODO: parse remianing types: fixed, arrays, etc.
+  else if(type.startsWith('bytes')) return parseBytes(type, rawValue);
+  else if(type.startsWith('uint')) return parseNumber(type, rawValue, true, true);
+  else if(type.startsWith('int')) return parseNumber(type, rawValue, false, true);
+  else if(type.startsWith('ufixed')) return parseNumber(type, rawValue, true, false);
+  else if(type.startsWith('fixed')) return parseNumber(type, rawValue, false, false);
+  else if(type.startsWith('bool')) return rawValue; // TODO
+  else throw new Error(ERROR_MESSAGE(type, rawValue) + '. Unsupported or invalid type.');
 }
 
-function parseBytes(rawValue: string | Buffer): string | never {
+function parseBytes(type: string, rawValue: string | Buffer): string | never {
   if(Buffer.isBuffer(rawValue)) return rawValue.toString();
   else if(typeof(rawValue) === 'string') {
     if(rawValue.toString().length === 0) return rawValue;
     // Require buffers expressed as strings to be valid hexadecimals.
-    if(!/^(0x)[0-9a-f]+$/i.test(rawValue)) throw new Error(ERROR_MESSAGE_BYTES(rawValue));
+    if(!/^(0x)[0-9a-f]+$/i.test(rawValue)) throw new Error(ERROR_MESSAGE(type, rawValue));
     // Test the validity of the buffer.
-    if(Buffer.from(rawValue.substring(2), 'hex').length === 0) throw new Error(ERROR_MESSAGE_BYTES(rawValue));
+    if(Buffer.from(rawValue.substring(2), 'hex').length === 0) throw new Error(ERROR_MESSAGE(type, rawValue));
     return rawValue;
   }
 }
 
-function parseAddress(rawValue: string | Buffer): string | never {
+function parseAddress(type: string, rawValue: string | Buffer): string | never {
   const strAddress = rawValue.toString();
-  if(!util.isValidAddress(strAddress)) throw new Error(ERROR_MESSAGE_ADDRESS(rawValue));
+  if(!util.isValidAddress(strAddress)) throw new Error(ERROR_MESSAGE(type, rawValue));
   // If the address' characters are not all uppercase or lowercase, assume that there is checksum to validate.
   if(/[a-f]/.test(strAddress.substring(2)) && /[A-F]/.test(strAddress.substring(2))) {
-    if(!util.isValidChecksumAddress(strAddress)) throw new Error(ERROR_MESSAGE_CHECKSUM_ADDRESS(rawValue));
+    if(!util.isValidChecksumAddress(strAddress)) throw new Error(ERROR_MESSAGE(type, rawValue) + '. Checksum error.');
   }
   return strAddress;
 }
 
-function parseNumber(rawValue: number | string | BN, mustBePositive: boolean): string | never {
-  if(isNaN(<any>rawValue)) throw new Error(ERROR_MESSAGE_NUMBER(rawValue));
-  if(typeof(rawValue === 'number') && !isFinite(<any>rawValue)) throw new Error(ERROR_MESSAGE_NUMBER(rawValue));
+function parseNumber(type: string, rawValue: number | string | BN, mustBePositive: boolean, mustBeInteger: boolean): string | never {
+  if(isNaN(<any>rawValue)) throw new Error(ERROR_MESSAGE(type, rawValue));
+  if(typeof(rawValue === 'number') && !isFinite(<any>rawValue)) throw new Error(ERROR_MESSAGE(type, rawValue));
   if(typeof(rawValue) === 'string') rawValue = (<string>rawValue).toLowerCase();
   // Funnel everything through bignumber.js.
   const bn = BN.isBigNumber(rawValue) ? <BN>rawValue : new BN(rawValue);
-  if(bn.isNaN()) throw new Error(ERROR_MESSAGE_NUMBER(rawValue));
-  if(mustBePositive && bn.isNegative()) throw new Error(ERROR_MESSAGE_POSITIVE_NUMBER(rawValue));
-  if(!bn.isInteger()) throw new Error(ERROR_MESSAGE_INTEGER_NUMBER(rawValue));
+  if(bn.isNaN()) throw new Error(ERROR_MESSAGE(type, rawValue));
+  if(mustBePositive && bn.isNegative()) throw new Error(ERROR_MESSAGE(type, rawValue));
+  if(mustBeInteger && !bn.isInteger()) throw new Error(ERROR_MESSAGE(type, rawValue));
   return bn.toString(10);
 }
--- a/packages/lib/src/helpers/encodeCall.ts
+++ b/packages/lib/src/helpers/encodeCall.ts
@@ -13,7 +13,7 @@ const ERROR_MESSAGE = (type: string, value: any) => `${ERROR_MESSAGE_PREFIX} for
 
 export default function encodeCall(name: string, types: string[] = [], rawValues: any[] = []): string {
   if(types.length !== rawValues.length) throw new Error(ERROR_MESSAGE_PREFIX + '. Supplied number of types and values do not match.');
-  const values =_.zipWith(types, rawValues, (type: string, value: any) => parseTypeValuePair(type, value));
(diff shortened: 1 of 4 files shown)


packages/lib/src/helpers/encodeCall.ts
```diff
@@ -1,15 +1,99 @@
...
-export default function encodeCall(name: string, args: string[] = [], rawValues: any[] = []): string {
-  const values: string[] = rawValues.map(formatValue);
-  const methodId: string = abi.methodID(name, args).toString('hex');
-  const params: Buffer = abi.rawEncode(args, values).toString('hex');
+export default function encodeCall(name: string, types: string[] = [], rawValues: any[] = []): string {
+  if(types.length !== rawValues.length) throw new Error('Supplied number of types and values do not match.');
+  const values = [];
+  _.zipWith(types, rawValues, function(type: string, value: any) {
+    values.push(parseTypeValuePair(type, value));
+  });
```
spalladino: This can be rewritten as:
```js
const values = _.zipWith(types, rawValues, (type: string, value: any) => parseTypeValuePair(type, value))
```

packages/lib/src/helpers/encodeCall.ts
```diff
@@ -1,15 +1,99 @@
...
+  if(!/^(0x)?[0-9a-f]{40}$/i.test(value)) throw new Error(ERROR_MESSAGE_ADDRESS(value));
+  // TODO: It'd be nice to perform a checksum validation here.
+  return value;
+}
+
+function parseNumber(value: any, mustBePositive: boolean): string | never {
+  if(typeof(value) === 'number') return parseNumberFromLiteral(value, mustBePositive);
+  else if(typeof(value) === 'string') return parseNumberFromString(value, mustBePositive);
+  else if(BN.isBigNumber(value)) return parseNumberFromBigNumber(new BN(value), mustBePositive);
+  throw new Error(ERROR_MESSAGE_NUMBER(value));
```
spalladino: Question about ts: is it possible to dispatch or pattern match based on the type of `value` in runtime? I mean, is it possible to have the functions `parseNumberFromLiteral`, `parseNumberFromString` and `parseNumberFromBigNumber` changed to just `parse(value: BN)`, `parse(value: string)`, and `parse(value: number)`, and then ts chooses which function to call?

packages/lib/src/helpers/encodeCall.ts
```diff
@@ -1,15 +1,99 @@
...
+}
+
+function parseNumberFromBigNumber(value: BN, mustBePositive: boolean): string | never {
+  if(mustBePositive && value.isNegative()) throw new Error(ERROR_MESSAGE_POSITIVE_NUMBER(value));
+  if(!value.isInteger()) throw new Error(ERROR_MESSAGE_INTEGER_NUMBER(value));
+  return value.toString();
+}
+
+function parseNumberFromString(value: string, mustBePositive: boolean): string | never {
```
spalladino: Can't we delegate all this logic to BigNumber? It would seem like `new BigNumber("0x40")`, `new BigNumber("4e10")`, etc all work properly. We should only check whether the result is NaN or not, since `new BigNumber("foo")` yields `NaN`.

packages/lib/src/helpers/encodeCall.ts
```diff
@@ -1,15 +1,99 @@
...
+    return parseNumber(rawValue, false);
+  }
+  else {
+    // TODO: parse remianing types: fixed, arrays, etc.
+    return rawValue;
+  }
+}
+
+function parseBytes(value: any): string | never {
+  if(typeof(value) !== 'string') throw new Error(ERROR_MESSAGE_BYTES(value));
```
spalladino: Not sure I like to enforce this. What if someone calls encodeCall programatically, using not a string but an actual array of  bytes? Can we skip the parseBytes case altogether?

packages/lib/src/helpers/encodeCall.ts
```diff
@@ -1,15 +1,99 @@
...
+  if(value.toString().length === 0) return value;
+  if(!/^(0x)?[0-9a-f]$/i.test(value)) throw new Error(ERROR_MESSAGE_BYTES(value));
+  return value;
+}
+
+function parseAddress(value: any): string | never {
+  if(typeof(value) !== 'string') throw new Error(ERROR_MESSAGE_ADDRESS(value));
```
spalladino: Can an address be stored as a Buffer or array of bytes?

packages/lib/src/helpers/encodeCall.ts
```diff
@@ -1,15 +1,99 @@
...
+  if(typeof(value) === 'number') return parseNumberFromLiteral(value, mustBePositive);
+  else if(typeof(value) === 'string') return parseNumberFromString(value, mustBePositive);
+  else if(BN.isBigNumber(value)) return parseNumberFromBigNumber(new BN(value), mustBePositive);
+  throw new Error(ERROR_MESSAGE_NUMBER(value));
+}
+
+function parseNumberFromBigNumber(value: BN, mustBePositive: boolean): string | never {
+  if(mustBePositive && value.isNegative()) throw new Error(ERROR_MESSAGE_POSITIVE_NUMBER(value));
+  if(!value.isInteger()) throw new Error(ERROR_MESSAGE_INTEGER_NUMBER(value));
+  return value.toString();
```
spalladino: Can we add a check to ensure that value is not NaN, which was the original issue?

packages/lib/src/helpers/encodeCall.ts
ajsantander: I'm afraid not. Or at least I can't think of any...
The closes thing I can think of is:
```function parseNumber(rawValue: number, mustBePositive: boolean): string | never;
function parseNumber(rawValue: string, mustBePositive: boolean): string | never;
function parseNumber(rawValue: BN, mustBePositive: boolean): string | never;
function parseNumber(rawValue: number | string | BN, mustBePositive: boolean): string | never {
  // Code...
  return 'a';
}
```
In Ts, you have to write method declarations for each of the overloads and then one method that takes a union of all the options that contains the actual code. Simpler overloading is supported when the number of parameters is different. I guess this is because if everything get's translated to Js, you have to perform the type checks anyway.
More info here: https://stackoverflow.com/questions/12688275/is-there-a-way-to-do-method-overloading-in-typescript/12689054#12689054

packages/lib/src/helpers/encodeCall.ts
ajsantander: Nice! You just annihilated `parseNumberFromString`, which was the most complex function in the file.

spalladino: Good to hear! The only thing I love more than writing code is deleting code.

ajsantander: Actually, the devastation reached parseNumberFromLiteral and parseNumberFromBN, which left us with just the one parseNumber function. Great feedback!! :tada:

packages/lib/src/helpers/encodeCall.ts
ajsantander: Yes, but I'm not understanding what you said about "which was the original issue". The original issue was caused by our scientific notation regex identifying an address with 'e' characters as a scientific notation number.

spalladino: Sort-of. The original issue was caused by that regex, but also because of the fact that we happily sent `NaN`s as valid integer values without checking them. I'd like to fix both issues.

packages/lib/src/helpers/encodeCall.ts
ajsantander: Yup. Agreed.
You will see an update that allows Buffers to be used for type bytes.

packages/lib/src/helpers/encodeCall.ts
ajsantander: Agreed. It can now also receive a Buffer.

packages/lib/src/helpers/encodeCall.ts
ajsantander: In that case, yes, we are now doing isNaN checks with regular Js as well as with BN.isNaN.

packages/lib/src/helpers/encodeCall.ts
```diff
@@ -1,15 +1,97 @@
...
+  // Single type.
+  if(type === 'address') return parseAddress(type, rawValue);
+  else if(type === 'bool') return parseBool(type, rawValue);
+  else if(type === 'string') return parseString(type, rawValue);
+  else if(type.startsWith('bytes')) return parseBytes(type, rawValue);
+  else if(type.startsWith('uint')) return parseNumber(type, rawValue, true, true);
+  else if(type.startsWith('int')) return parseNumber(type, rawValue, false, true);
+  else if(type.startsWith('ufixed')) return parseNumber(type, rawValue, true, false);
+  else if(type.startsWith('fixed')) return parseNumber(type, rawValue, false, false);
+  else throw new Error(ERROR_MESSAGE(type, rawValue) + '. Unsupported or invalid type.');
```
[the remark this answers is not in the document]
ajsantander: True! Given that it's a pretty rare usage of the ABI, do you think it would be ok to add a couple of TODOs here for now, and simply forward these types without erroring "Unsupported or invalid type"? Implementing a validation for these may be a bit complex. If you think that would be ok, I would create a new issue to deal with these more complex types in a separate PR.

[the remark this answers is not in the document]
spalladino: Absolutely, and we can delay that check until much later.

packages/lib/src/helpers/encodeCall.ts
```diff
@@ -1,15 +1,97 @@
...
+  if(types.length !== rawValues.length) throw new Error(ERROR_MESSAGE_PREFIX + '. Supplied number of types and values do not match.');
+  const values =_.zipWith(types, rawValues, (type: string, value: any) => parseTypeValuePair(type, value));
+  const methodId = abi.methodID(name, types).toString('hex');
+  const params = abi.rawEncode(types, values).toString('hex');
   return '0x' + methodId + params;
 }
+
+export function parseTypeValuePair(type: string, rawValue: any): any | never {
+  // Array type (recurse by calling this function with the individual elements).
+  if(/^[^\[]+\[.*\]$/.test(type)) { // Test for '[]' in type.
```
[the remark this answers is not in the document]
ajsantander: Are you sure? Check this out: https://regex101.com/r/pnS8Lq/1

packages/lib/src/helpers/encodeCall.ts
```diff
@@ -1,15 +1,97 @@
...
-  else return value;
-}
+const ERROR_MESSAGE_PREFIX = 'Encoding error';
+const ERROR_MESSAGE = (type: string, value: any) => `${ERROR_MESSAGE_PREFIX} for type ${type} and value ${value.toString()}`;
 
-export default function encodeCall(name: string, args: string[] = [], rawValues: any[] = []): string {
-  const values: string[] = rawValues.map(formatValue);
-  const methodId: string = abi.methodID(name, args).toString('hex');
-  const params: Buffer = abi.rawEncode(args, values).toString('hex');
+export default function encodeCall(name: string, types: string[] = [], rawValues: any[] = []): string {
```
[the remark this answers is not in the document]
ajsantander: We could definitely split the function to:
1) A simpler encodeCall function that performs point 1, i.e. with no validation.
2) An encodeAndParseCall function that runs the validations.
We would then have to identify the points in the code (outside of this file of course) that involve CLI interaction. The tests would have to be modified too, to account for the two different usages.

However, I'm afraid I don't see the point of not validating encodeCall when it's being used programmatically. If the code is being called via Javascript, the user can screw up as well. In fact, although it is less likely, the user can also screw up when using typescript.

Regarding your comment about using ethereumjs-abi's simpleEncode function, I also don't see the advantage of that. We would have to manually build the strings for type and data, instead of just passing it arrays of types and values.

packages/lib/src/helpers/encodeCall.ts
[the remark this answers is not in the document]
spalladino: Damn, I missed the end-of-string anchor. Forget I said anything!

packages/lib/src/helpers/encodeCall.ts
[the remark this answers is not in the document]
spalladino: > However, I'm afraid I don't see the point of not validating encodeCall when it's being used programmatically. If the code is being called via Javascript, the user can screw up as well. In fact, although it is less likely, the user can also screw up when using typescript.

The point is providing a consistent API with ABI encode. For instance, if the user supplies `"false"` (the string) for a boolean argument in a regular encodeCall, it gets parsed as true, since a non-empty string is truthy in javascript. With the current behaviour, a user would get different results if initializing a contract manually vs via ZeppelinOS, assuming they are using the programmatic interface.

packages/lib/src/helpers/encodeCall.ts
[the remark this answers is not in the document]
ajsantander: Actually, I couldn't help myself and I implemented them :grimacing: 
Implementing the function type was trivial, and implementing tuples is almost identical to arrays.

packages/lib/src/helpers/encodeCall.ts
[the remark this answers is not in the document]
ajsantander: Ok great. I think I understand you fully now :)

I changed the code so that encodeCall now performs no validation whatsoever. It simply forwards the incoming data to ethereumjs-abi. And so, when using encodeCall programmatically, (like in tests for example, or from a program using zos-lib as a dependency), there will be no parsing nor validation on the incoming values.

However, when encode call is used from the zos CLI, which does so via the single entry point of ABIs.ts, instead of calling `encodeCall(method.name, argTypes, args)`, it now does `encodeCall(method.name, argTypes, parseCallValues(argTypes, args))`. Note that parseCallValues is a new function in the encodeCall file, that exclusively does the rawValues to values transformation, with type/value validation.

packages/lib/src/helpers/encodeCall.ts
```diff
@@ -1,15 +1,116 @@
...
+  else if(type.startsWith('int')) return parseNumber(type, rawValue, false, true);
+  else if(type.startsWith('ufixed')) return parseNumber(type, rawValue, true, false);
+  else if(type.startsWith('fixed')) return parseNumber(type, rawValue, false, false);
+  else throw new Error(ERROR_MESSAGE(type, rawValue) + '. Unsupported or invalid type.');
+}
+
+function parseTuple(type: string, rawValue: any): any | never {
+  if(typeof rawValue === 'string') rawValue = rawValue.split(',');
+  if(rawValue.length === 0) return [];
+  const types = type.replace(/[{()}]/g, '').split(','); // Remove the parenthesis and split the tuple into types.
```
spalladino: Shouldn't we remove only the most outer parentheses? Otherwise, we are losing any kind of tuple nesting, which may lead to errors. For instance:
```
> "(uint,(uint,uint)[])".replace(/[{()}]/g, '')
'uint,uint,uint[]'
```

packages/lib/src/helpers/encodeCall.ts
```diff
@@ -1,15 +1,116 @@
...
+  if(rawValue.length === 0) return [];
+  const size = type.match(/(.*)\[(.*?)\]$/)[2]; // Find number between '[*]'.
+  if(size !== '' && parseInt(size, 10) !== rawValue.length) throw new Error(ERROR_MESSAGE(type, rawValue) + '. Invalid array length.');
+  const baseType = type.slice(0, type.lastIndexOf('[')); // Remove array part '[*]'.
+  return _.map(rawValue, (rawValueElement) => parseTypeValuePair(baseType, rawValueElement));
+}
+
+function parseFunction(type: string, rawValue: string): string | never {
+  if(typeof rawValue !== 'string') throw new Error(ERROR_MESSAGE(type, rawValue)); // Runtime type check.
+  if(!/^(0x)[0-9a-f]{46}$/i.test(rawValue)) throw new Error(ERROR_MESSAGE(type, rawValue));
```
spalladino: Why not fall back to `parseNumber` here, instead of forcing a hex value?

packages/lib/src/helpers/encodeCall.ts
```diff
@@ -1,15 +1,116 @@
...
+function parseString(type: string, rawValue: string): string | never {
+  if(typeof rawValue !== 'string') throw new Error(ERROR_MESSAGE(type, rawValue)); // Runtime type check.
+  return rawValue;
+}
+
+function parseBool(type: string, rawValue: string | boolean): boolean | never {
+  if(typeof rawValue !== 'string' && typeof rawValue !== 'boolean') throw new Error(ERROR_MESSAGE(type, rawValue)); // Runtime type check.
+  if(typeof rawValue === 'boolean') return rawValue;
+  else if(typeof rawValue === 'string') {
+    if(rawValue === 'true') return true;
```
spalladino: Can we make this case insensitive?

packages/lib/src/helpers/encodeCall.ts
```diff
@@ -1,15 +1,116 @@
...
-  const values: string[] = rawValues.map(formatValue);
-  const methodId: string = abi.methodID(name, args).toString('hex');
-  const params: Buffer = abi.rawEncode(args, values).toString('hex');
+export default function encodeCall(name: string, types: string[] = [], values: any[] = []): string {
+  const methodId = abi.methodID(name, types).toString('hex');
+  const params = abi.rawEncode(types, values).toString('hex');
   return '0x' + methodId + params;
 }
+
+export function parseCallValues(types: string[] = [], rawValues: any[] = []): any[] | never {
```
spalladino: If we are now using parse just to handle input strings, shouldn't rawValues be a string[]? Which use case do we want to cover by accepting a `bool` value for `parseBool`, for instance?

packages/lib/src/utils/ABIs.ts
```diff
@@ -20,7 +20,7 @@ interface FunctionInfo {
 export function buildCallData(contractClass: ContractFactory, methodName: string, args: any[]): CalldataInfo {
   const method = getABIFunction(contractClass, methodName, args);
   const argTypes = method.inputs.map((input) => input.type);
-  const callData = encodeCall(method.name, argTypes, args);
+  const callData = encodeCall(method.name, argTypes, parseCallValues(argTypes, args));
```
spalladino: I'm not sure if this is the best place to add `parseCallValues`. This function gets called from lib/App and lib/SimpleProject, which can be called directly from the js interface, where the user specifies actual values and we do not want to go through the parsing process. I'd consider moving this all the way up to the CLI, potentially cli/scripts/create and cli/scripts/update.

packages/lib/src/helpers/encodeCall.ts
ajsantander: Yes, you're right. I was just removing all parentheses and then splitting by commas, which resulted in a complete flattening of the tuple, which I guess could be ok.
I changed it so that the outer parentheses are removed, and then the string is split by types again (which may include other tuples). Unfortunately, I couldn't come up with a reliable regex for this, so I had to add a quite wordy function `splitTupleTypes(type: string): string[];` :disappointed: 
However, it's consuming super complicated nesting, as you can see in the tests.

packages/lib/src/helpers/encodeCall.ts
ajsantander: Because I want to make sure that the value has exactly 24 bytes (an address + a function selector hash).

packages/lib/src/helpers/encodeCall.ts
ajsantander: Actually, encodeCall should also have values typed to any[] and not string[]. That's because ethereumjs-abi is not expecting values to be just strings. That makes me hesitate a bit about removing validation from programmatic usage...
In the case of parseCallValues, most of the parsing functions are prepared to receive non string values, such as parseNumber, which can receive number types as well as big number types, or the case of parseBytes, that can receive Buffers.

packages/lib/src/utils/ABIs.ts
ajsantander: I see what you mean, but we have the type data at ABI.ts via analysis of the ABI for the function being called: https://github.com/zeppelinos/zos/blob/master/packages/lib/src/utils/ABIs.ts#L21

In the CLI's scripts this information is not yet available.

spalladino: I see, but we do need a way to keep a consistent interface when the user interacts programatically. I'm not comfortable in having different behaviours when initializing a contract via SimpleProject versus via web3 or truffle.

Can we make the ABI types available higher up in the chain? Or add an optional flag `parseInitArgs` that controls whether they are parsed or not, and is set to true only from the CLI?

packages/lib/src/helpers/encodeCall.ts
spalladino: > Actually, encodeCall should also have values typed to any[] and not string[]. That's because ethereumjs-abi is not expecting values to be just strings. 

Yes, I'd expect encodeCall to work with any[], not string[], since it should receive the correct type (or something that can be easily coercerd to) for each value.

> In the case of parseCallValues, most of the parsing functions are prepared to receive non string values, such as parseNumber, which can receive number types as well as big number types, or the case of parseBytes, that can receive Buffers.

I see, but my point is: if we are *only* using parseCallValues to parse strings, why support scenarios that we never use? I'd rather clear that code out if it's never used.

packages/lib/src/helpers/encodeCall.ts
```diff
@@ -1,15 +1,139 @@
...
+  else if(type === 'function') return parseFunction(type, rawValue);
+  else if(type.startsWith('bytes')) return parseBytes(type, rawValue);
+  else if(type.startsWith('uint')) return parseNumber(type, rawValue, true, true);
+  else if(type.startsWith('int')) return parseNumber(type, rawValue, false, true);
+  else if(type.startsWith('ufixed')) return parseNumber(type, rawValue, true, false);
+  else if(type.startsWith('fixed')) return parseNumber(type, rawValue, false, false);
+  else throw new Error(ERROR_MESSAGE(type, rawValue) + '. Unsupported or invalid type.');
+}
+
+function parseTuple(type: string, rawValue: any): any | never {
```
spalladino: I'd remove tuple parsing altogether if ethereumjs-abi cannot support them. I insist in that there is no point in parsing something that we cannot encode. We could keep this code somewhere when we do need to add that support, but for now, I'd remove that case.

ajsantander:
After discussion with @spalladino, we are closing this PR in favor of a new one that uses ethers.js/abi-coder instead of ethereumjs-abi in encodeCall, since it performs a lot of the validations that were being developed in this PR internally, and because it is what web3 is currently using.

PR #550, opened 2019-01-02
Remove lightweight terminology

jcarpanelli:
This PR removes some deprecated terminology related to the state of a zos project. Basically, all references to a `full project` and `lightweight project` were renamed in favor of  `published project` and `unpublished project`.

### DIFF
--- a/packages/cli/src/models/files/ZosNetworkFile.ts
+++ b/packages/cli/src/models/files/ZosNetworkFile.ts
@@ -121,8 +121,8 @@ export default class ZosNetworkFile {
     return Object.keys(this.contracts);
   }
 
-  get isLightweight(): boolean {
-    return this.packageFile.isLightweight;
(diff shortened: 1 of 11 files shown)


spalladino:
@jcarpanelli I'm unsure about the change `isLightweight => !isPublished` in both ZosPackageFile and ZosNetworkFile. 

The flag `publish` determines if a project should be published by default when running a `zos push` to a network, which is not the same as whether the project is published. For instance, a project may not have the `publish` flag, but could be published to a network, and in that case `isPublished` would return false.

I'm not very imaginative with names today, but maybe something like `publishOnPush`, `alwaysPush`, or `publishByDefault` may be more accurate?

Also, I'm fine with just deferring this issue for later. Internal name changes are not a priority right now, and I'd rather have you invest your time in more interesting stuff :-)

PR #555, opened 2019-01-03
Add lodash functions separately

slorenzo:
I have removed `lodash` to add the same functions separately. With that, you will take more control about what you are using of `lodash` and specify this on the `package.json`. In addition, it reduces the bundle size and avoid to include unnecessary functions to the project.

If you need to use other functions, type: `npm install --save lodash.yourfunction` to include it to the project.

Hope you like it!


facuspagnuolo:
Thanks @slorenzo for your contribution! The CI is failing tho, it seems that `isEmpty` is missing, can you please take a look at it?

slorenzo:
Hi @facuspagnuolo. I have fixed it [here](https://github.com/zeppelinos/zos/pull/555/commits/a5e8e9ccb62a2a214232727f534226a23ab139db). Everything should works now. Let me know if it's OK.

P.S. ZepellinOS is an awesome project and I made this PR to introduce myself to the team. If the search is still open, I hope you can consider me for this [position](https://zeppelin.solutions/jobs/frontend-engineer/).

facuspagnuolo:
Thanks @slorenzo! Would you like to do the same on lib? Since the CLI is depending on lib, it would be nice to have it fixed in both sides so we don't depend on the whole lodash lib in the entire project anymore.

slorenzo:
Sure @facuspagnuolo , I will look at it.

facuspagnuolo:
@slorenzo sure! we already have it fixed on master, would you mind rebasing?

PR #556, opened 2019-01-05
Replaces ethereumjs-abi with ethers.js/abi-coder

### LINKED ISSUE #421: Initialization argument encoding issues.

ajsantander:
Fix #421
(3rd attempt at fixing it the right way :+1: )

### The bug:
Lib's encodeCall helper was failing for any hexadecimal strings that contained the character 'e'. It was interpreting these as scientific notation numbers (e.g. 1.5e9) and seeing their hexadecimal value parsed a decimal instead =S.

### Beyond the bug:
This PR doesn't only fix the bug, but replaces the dependency used for encoding in favor of ethers/abi-coder, which performs a lot of type/value pair validations internally. Hopefully, this would avoid many errors like #421 in the future. It also adds a few tests for the encodeCall helper.

@spalladino please note a few things:
1) ethers.js/abi-coder is typed :tada: 
2) No distinction was made for encodeCall regarding programmatic and cli usage. This is because, given that erhers is what web3 is using, both could be treated equally; unless we want our cli usage to be more permissive of course (or even safer). In that case we could add a flag to the helper function so that it parses some of the incoming values in order to be more flexible in terms of what encodeCall accepts from the command line, and/or perform additional validations.
3) Despite ethers/abi-coder being very robust in terms of the validations it performs on it's incoming values during encoding, it does not support some types listed in the Solidity documentation, such as ufixed, fixed, and functions. I've opened an issue in ethers.js, just to be sure: https://github.com/ethers-io/ethers.js/issues/389


packages/cli/test/scripts/create.test.js
```diff
@@ -117,15 +117,10 @@ contract('create script', function([_, owner]) {
     });
 
     it('should initialize a proxy using scientific notation', async function() {
-      const proxy = await createProxy({ contractAlias, network, txParams, initMethod: 'initialize', initArgs: ["20e10"], networkFile: this.networkFile });
+      const proxy = await createProxy({ contractAlias, network, txParams, initMethod: 'initialize', initArgs: [Number("20e10")], networkFile: this.networkFile });
       (await proxy.value()).toString().should.be.eq('200000000000');
     });
 
-    it('should initialize a proxy using large number on scientific notation', async function() {
```
ajsantander: This test was removed because ethers/abi-coder checks that the incoming number is within uint256 bounds.

packages/lib/src/helpers/encodeCall.ts
```diff
@@ -1,15 +1,8 @@
...
-}
-
-export default function encodeCall(name: string, args: string[] = [], rawValues: any[] = []): string {
-  const values: string[] = rawValues.map(formatValue);
-  const methodId: string = abi.methodID(name, args).toString('hex');
-  const params: Buffer = abi.rawEncode(args, values).toString('hex');
-  return '0x' + methodId + params;
+export default function encodeCall(name: string, types: Array<string | ParamType> = [], rawValues: any[] = []): string {
+  const encodedParameters = defaultAbiCoder.encode(types, rawValues).substring(2);
+  const signatureHash = sha3(`${name}(${types.join(',')})`).substring(2, 10);
```
spalladino: Can we add a TODO here to switch to something native of web3 once we migrate to 1.0, such as https://github.com/ethereum/web3.js/blob/1.0/packages/web3-eth-abi/src/index.js#L52?

spalladino: Moreover, when we migrate to web3 1.0, can we directly use the `web3-eth-abi`, and remove the explicit dependency on ethers.js?

ajsantander: Sure! TODO note added.
Web3 1.x is actually using exactly the same dependencies we are: https://github.com/ethereum/web3.js/blob/1.0/packages/web3-eth-abi/src/index.js

ajsantander:
@spalladino, regarding:

> Do we have any tests that check that, for every argument type, we can successfully supply them via the command line? In other words, if the create command (or script) is invoked with initArgs ["42", "0x12345","false"] (note that these are all strings, since they were received via the command line) for types [uint256,address,bool], is encodeCall going to work properly?

Most existing tests have an entry that takes in values in string format, except a couple. I'll add to these, and there shouldn't be a problem. The only case in which I see an issue is with booleans, since passing `"false"` (the string) will result in the abi-coder encoding something like `0x9476a7100000000000000000000000000000000000000000000000000000000000000001`. Perhaps for these cases we could re-introduce a simple parseParameters function that, for now, only pre-processes booleans.

spalladino:
> The only case in which I see an issue is with booleans, since passing "false" (the string) will result in the abi-coder encoding something like 0x9476a7100000000000000000000000000000000000000000000000000000000000000001. Perhaps for these cases we could re-introduce a simple parseParameters function that, for now, only pre-processes booleans?

+1, let's try to do that. Also, please make sure that there is at least a test to double check that numbers are handled correctly if passed in as strings.

ajsantander:
As discussed with @facuspagnuolo, encodeCall will not be parsing any values (like booleans). Pls see: https://github.com/zeppelinos/zos/issues/558

packages/lib/test/src/helpers/encodeCall.test.js
```diff
@@ -0,0 +1,192 @@
...
+    
+    it('should throw when the passed tuple types do not match', () => {
+      assertBadEncoding(['tuple(uint256,string)'], [['hello', 42]], null);
+      assertBadEncoding(['tuple(uint256,string)'], [['42']], null);
+    });
+    
+    it('supports nested tuples', function() {
+      assertGoodEncoding(['tuple(uint256,string[])'], [[42, ['one', 'two', 'three']]]);
+      assertGoodEncoding(['tuple(uint256,tuple(uint256,string))'], [[42, [42, 'hello']]]);
+      //
```
facuspagnuolo:
```suggestion
```

packages/lib/test/src/helpers/encodeCall.test.js
```diff
@@ -0,0 +1,192 @@
...
+      assertBadEncoding(['address'], [2], /invalid address/);
+      assertBadEncoding(['address'], [{}], /invalid address/);
+    });
+  });
+  
+  // TODO: ethers.js/abi-coder does not support ufixed and fixed types?
+  // describe('when the specified type is a function', () => {});
+
+  // TODO: ethers.js/abi-coder does not support function types?
+  // describe('when the specified type is a function', () => {});
```
facuspagnuolo: move these TODOs to issues if necessary please

packages/lib/test/src/helpers/encodeCall.test.js
```diff
@@ -0,0 +1,192 @@
+'use strict';
+
+require('../../setup');
+
+import encodeCall, { decodeCall } from '../../../src/helpers/encodeCall';
+import BN from 'bignumber.js';
+import _ from 'lodash';
+
+const NAME = 'myFunction';
```
facuspagnuolo:
```suggestion
const FUNCTION_NAME = 'myFunction';
```

packages/lib/test/src/helpers/encodeCall.test.js
```diff
@@ -0,0 +1,192 @@
...
+
+function assertGoodEncoding(types, values) {
+  const encoded = encodeCall(NAME, types, values).substring(10); // Remove signature hash.
+  const decoded = decodeCall(types, `0x${encoded}`);
+  if(values.length !== decoded.length) throw new Error('Invalid ecoding/decoding: Mismatch in number of encoded and decoded values.');
```
facuspagnuolo:
```suggestion
  if(values.length !== decoded.length) throw new Error('Invalid encoding/decoding: Mismatch in number of encoded and decoded values.');
```

packages/lib/test/src/helpers/encodeCall.test.js
```diff
@@ -0,0 +1,192 @@
+'use strict';
+
```
facuspagnuolo:
```suggestion
```

packages/lib/test/src/helpers/encodeCall.test.js
```diff
@@ -0,0 +1,192 @@
...
+  if(values.length !== decoded.length) throw new Error('Invalid ecoding/decoding: Mismatch in number of encoded and decoded values.');
+  _.zipWith(values, decoded, (value, decodedValue) => {
+    if(Buffer.isBuffer(value)) value = `0x${value.toString('hex')}`;
+    if(value.toString() != decodedValue.toString()) throw new Error(`Invalid encoding/decoding. Encoded: ${value}, Decoded: ${decodedValue}`);
+  });
+}
+
+function assertBadEncoding(types, values, errorRegex) {
+  expect(() => encodeCall(NAME, types, values)).to.throw(errorRegex);
+}
```
facuspagnuolo: I'd move the assert functions to the end of the file to ensure that the first thing you see when opening a test file is the `describe` statement

packages/lib/test/src/helpers/encodeCall.test.js
ajsantander: I already did. I'll just delete the TODO comments. Thx!

PR #567, opened 2019-01-09
Parse scientific notation numbers in input.ts.

### LINKED ISSUE #559: Restore ability to pass scientific notation numbers in the command line
If PR https://github.com/zeppelinos/zos/pull/556 is merged, it would have removed the ability to pass scientific notation numbers like `1e5` via the cli. `lib/helpers/encodeCall` does not try to parse these strings to numbers anymore.

This feature needs to be restored, at the level of `cli/src/utils/input.ts`.

ajsantander:
Fix #559 

This PR adds a remix-like regex at input.ts in order to parse values like `"1e2"` to `"100"` or `"1.5e20"` to `"150000000000000000000"`, and tests for it at input.test.js. It also modifies a couple of the other tests in this file to make them a bit more accurate.

Note that the introduced `MATCH_SCIENTIFIC` regex is slightly more complicated than the others, because it involves performing a Big Number operation in the replacement string. It should be looked at carefully to make sure that we don't introduce new bugs with this change.

Also note that this PR is compatible with the spirit of PR #556, that intends to do no parsing whatsoever at the programatic level in encodeCall and only do it when parsing CLI input, in order to maintain full compatibility with encodeCall and web3 contract interaction.


packages/cli/src/utils/input.ts
```diff
@@ -1,14 +1,27 @@
...
   // The following is inspired from ethereum/remix (Copyright (c) 2016)
   // https://github.com/ethereum/remix/blob/89f34fc4f35eca0c386379550c300205d35bfae0/remix-lib/src/execution/txFormat.js#L51-L53
   try {
+     // replace non quoted hex string by quoted hex string
     const MATCH_HEX = /(^|([,[]\s*))(0[xX][0-9a-fA-F]+)/g;
-    args = args.replace(MATCH_HEX, '$2"$3"'); // replace non quoted hex string by quoted hex string
+    args = args.replace(MATCH_HEX, '$2"$3"');
+
+    // replace scientific notation numbers by regular numbers
+    const MATCH_SCIENTIFIC = /(^|([,[]\s*))(\s*[-]?\d+(\.\d+)?e(\+)?\d+\s*)/g;
```
spalladino: Aren't we missing an end-of-string (or `($|([,\]])`) anchor in this regex?

ajsantander: I'm not sure I'd call `(^|([,[]\s*))` start of string anchors, they're more like capturing groups. Remember that in this case we are parsing the whole input, not just a single value.

Despite that, if we are missing an "end of string capturing group", then we would need to add them to the other regex-es, since none of them use them.

spalladino: > Despite that, if we are missing an "end of string capturing group", then we would need to add them to the other regex-es, since none of them use them.
Yep, exactly what I was thinking about.

ajsantander: Interesting... if we do add the end of string capturing groups in all the regex-es, it would capture ending commas and `]`'s. Take a look: https://regex101.com/r/oES6aY/1/

packages/cli/test/utils/input.test.js
```diff
@@ -18,10 +21,12 @@ describe('input', function () {
...
-    it('should parse multiple addresses', testFn('0x1234,0x1235', "0x1234", "0x1235"));
+    it('should parse a hex string', testFn("0x1234", "0x1234"));
+    it('should parse an address', testFn("0x39af68cF04Abb0eF8f9d8191E1bD9c041E80e18e", "0x39af68cF04Abb0eF8f9d8191E1bD9c041E80e18e"));
+    it('should parse a quoted address', testFn('"0x39af68cF04Abb0eF8f9d8191E1bD9c041E80e18e"', "0x39af68cF04Abb0eF8f9d8191E1bD9c041E80e18e"));
+    it('should parse multiple addresses', testFn('0x39af68cF04Abb0eF8f9d8191E1bD9c041E80e18e,0x39af68cF04Abb0eF8f9d8191E1bD9c041E80e18e', "0x39af68cF04Abb0eF8f9d8191E1bD9c041E80e18e", "0x39af68cF04Abb0eF8f9d8191E1bD9c041E80e18e"));
+    it('should parse an array of non quoted addresses', testFn('[0x39af68cF04Abb0eF8f9d8191E1bD9c041E80e18e,0x39af68cF04Abb0eF8f9d8191E1bD9c041E80e18e]', ["0x39af68cF04Abb0eF8f9d8191E1bD9c041E80e18e", "0x39af68cF04Abb0eF8f9d8191E1bD9c041E80e18e"]));
+    it('should parse multiple arguments', testFn('1e2,42,0x39af68cF04Abb0eF8f9d8191E1bD9c041E80e18e,43,1.5e20,"foo",[1e2,1,2,"bar",1e3]', '100', '42', '0x39af68cF04Abb0eF8f9d8191E1bD9c041E80e18e', '43', '150000000000000000000', 'foo', ['100', '1', '2', 'bar', '1000']));
```
facuspagnuolo: can we add a test for negative scientific notation numbers?

ajsantander:
@spalladino 

> 1. Be much more strict, and require everything to be properly enclosed in quotes (ie be valid JSON). This is more painful when accepting user input, but is a lot safer.
> 2. Generating a parser from a full grammar to handle the input, which should be safer and accommodate for more cases, but is also a lot more of work.

The third option would be to merge this now (since it does restore the scientific notation feature) and open a new issue to revisit the topic with a safer choice 1 or 2. I would go with the second option; Almost 90% of the cosed from one of my closed encodeCall PRs could be reused.

packages/lib/package.json
```diff
@@ -69,7 +69,7 @@
     "openzeppelin-solidity": "~1.10.0",
     "semver": "^5.5.1",
     "truffle-flattener": "^1.2.8",
-    "web3": "^1.0.0-beta.37"
+    "web3": "1.0.0-beta.37"
```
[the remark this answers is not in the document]
ajsantander: Yes, apparently web3 1.x beta > 37 and < 41 breaks CI tests with the "websocket" sub-dependency of "web3-providers-ws".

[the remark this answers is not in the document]
ajsantander: It's already pinned in master from one of my other PRs: https://github.com/zeppelinos/zos/blob/master/packages/lib/package.json#L72

spalladino:
Agree Ale, merging now! And please add a new issue to review the parsing of arguments, including both options. I think Remix is currently going with (1), but it'd be interesting to have (2).

PR #569, opened 2019-01-09
Fix encodeCall address handling

facuspagnuolo:
Fixes https://github.com/zeppelinos/zos/issues/501

### DIFF
--- a/packages/lib/src/helpers/encodeCall.js
+++ b/packages/lib/src/helpers/encodeCall.js
@@ -1,19 +1,30 @@
 import abi from 'ethereumjs-abi'
 import BN from 'bignumber.js'
 
-function formatValue(value) {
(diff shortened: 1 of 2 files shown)


packages/lib/src/helpers/encodeCall.js
```diff
@@ -1,19 +1,30 @@
...
 
 export default function encodeCall(name, args = [], rawValues = []) {
   const values = rawValues.map(formatValue)
   const methodId = abi.methodID(name, args).toString('hex');
   const params = abi.rawEncode(args, values).toString('hex');
   return '0x' + methodId + params;
 }
+
+export function decodeCall(types, data) {
+  if (typeof data === 'string') data = new Buffer(data, 'hex')
```
ajsantander: nit: "new Buffer()" is deprecated in favor of "Buffer.from(...)"

PR #570, opened 2019-01-09
Remove custom compiling process with solc

facuspagnuolo:
Given that ZeppelinOS currently only works on top of Truffle projects, having all this code supporting a custom compiling process with solc does not make much sense. I'm removing it from master so it is not included in the next release. We can bring it back later when ZeppelinOS does not depend only on Truffle.


packages/cli/src/models/compiler/Compiler.ts
```diff
@@ -1,42 +1,18 @@
...
 import { FileSystem, Contracts, Logger } from 'zos-lib';
-import Truffle from '../initializer/truffle/Truffle';
-import SolidityProjectCompiler from './solidity/SolidityProjectCompiler';
-import { CompilerOptions } from './solidity/SolidityContractsCompiler';
 
 const log = new Logger('Compiler');
 
-export default {
+const Compiler = {
   async call(): Promise<{ stdout: string, stderr: string } | void> {
```
jcarpanelli: nit: `call()` shouldn't have the `void` return type. It was added for the particular case in which we compile with `solc` in `compileWithSolc`, but this function has been removed in this same PR.

PR #571, opened 2019-01-09
Fix doc guides to support Truffle 5 environment

facuspagnuolo:
Fixes https://github.com/zeppelinos/zos/issues/562

We are telling the user to run `npx truffle console` to test their upgradeable contract. Therefore, if the user is running a Truffle 5 project, the steps we are telling them to reproduce won't work: we are using `at` which is now async.

### DIFF
--- a/packages/docs/docs/docs/erc20_onboarding.md
+++ b/packages/docs/docs/docs/erc20_onboarding.md
@@ -220,8 +220,8 @@ or open a new one against the network where your legacy token is deployed. Then,
 commands of the previous step:
 
 ```console
-truffle(local)> erc20Migrator = ERC20Migrator.at('ERC20_MIGRATOR_ADDRESS')
-truffle(local)> upgradeableToken = MyUpgradeableToken.at('UPGRADEABLE_TOKEN_ADDRESS')
+truffle(local)> ERC20Migrator.at('ERC20_MIGRATOR_ADDRESS').then(i => erc20Migrator = i)
+truffle(local)> MyUpgradeableToken.at('UPGRADEABLE_TOKEN_ADDRESS').then(i => upgradeableToken = i)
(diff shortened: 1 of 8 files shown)


packages/docs/docs/docs/erc20_onboarding.md
```diff
@@ -220,8 +220,8 @@ or open a new one against the network where your legacy token is deployed. Then,
 commands of the previous step:
 
 ```console
-truffle(local)> erc20Migrator = ERC20Migrator.at('ERC20_MIGRATOR_ADDRESS')
-truffle(local)> upgradeableToken = MyUpgradeableToken.at('UPGRADEABLE_TOKEN_ADDRESS')
+truffle(local)> ERC20Migrator.at('ERC20_MIGRATOR_ADDRESS').then(i => erc20Migrator = i)
```
jcarpanelli: I know this will do the trick, but isn't it a bit dirty? I mean, defining a global variable inside the arrow function could be a bit confusing imo.
Is there any other way of accomplishing this, working on a node console?

facuspagnuolo: It's the simpler version that came to my mind, are you thinking about any other? Using a plain node console would require more setup making these examples longer. What we can discuss that afterward if we think it's better for the doc examples

facuspagnuolo:
@jcarpanelli I tested this using truffle 4 and 5 and it works ok

packages/docs/docs/docs/erc20_onboarding.md
jcarpanelli: > are you thinking about any other?
Not really :disappointed:. Handling promises inside the console is annoying.

jcarpanelli:
> @jcarpanelli I tested this using truffle 4 and 5 and it works ok

Ok, let's merge then :slightly_smiling_face:

PR #572, opened 2019-01-09
Updates Web3 to v1.x

### LINKED ISSUE #535: Use Web3 1.x types
Web3 v1.x is pretty stable now and should be non-beta soon. We should consider upgrading our Web3 from v0.x to v1.x.

**Related to the Typescript migration**

During the port to Typescript of the lib package, and small experiment was conducted to type web3: #510 
The effort was abandoned because (1) of the proximity of Web3 v1.x, (2) that Web3 v0.x doesn't have official types in the repo nor in DefinitelyTyped, and because (3) looks like Web3 v1.x will offer official types.

During the refactor, @types/web3 should be added ASAP, which should be super helpful for refactoring the code.

### LINKED ISSUE #563: Update web3 to version 1.x
The idea is to take advantage of the improvements coming with the new version of web3, while dropping most part of the custom code around contract wrappers, web3 interfaces, etc. 

Some things you will need to take care are:
- [ ] `ZWeb3` shouldn't be needed anymore
- [ ] `ContractFactory` shouldn't be needed
- [ ] Some tests will need to be updated to the way web3 1.x handles events
- [ ] Check if some of the custom gas/gas price estimation logic could be replaced using some of the new features provided by web3 1.x
- [ ] Review lib initialization process to make sure it works fine within a truffle environment (remember that web3 instances are injected from CLI - consider https://github.com/zeppelinos/zos/issues/508)

ajsantander:
Fix #563 
Fix #535 

## Goals of this PR

This PR intends to update the Web3 dependency in ZeppelinOS from v0.x to v1.x. It does not intend to upgrade Truffle v4.x to v5.x tho, so you will see, especially in tests, that Web3 v0.x is coexisting with Web3 v1.x. If this sounds weird, it's because it is, but it's not that bad. Another PR that updates Truffle to v5 should remove this bizarre intermediate state. Also note that this PR simply attempts to "make the code run" with Web3 v1 and does not focus on design changes, i.e. it does not yet take advantage of several design optimizations that could be implemented because of the way Web3 v1 works, and how it's typed. Another PR should focus on this specifically and hopefully remove a lot of code that will no longer be needed.

## Changes that you will see in the code:

*This should be helpful for reviewing, and also should be very informative to understand what actually changed and how to use Web3 v1 in the code in the future.*

1) Despite this PR not intending to make design changes, there is ONE exception: ContractFactory no longer injects methods at the top level of a contract, as Truffle does. This is meant as yet another step towards interacting with Web3 directly and not via Truffle. So, you will see things like:
```await myContract.methods.initialize(...PARAMS).send()```
or
```await myContract.methods.owner().call()```
instead of the old
```await myContract.initialize(...PARAMS)```
or
```await myContract.owner()```
This is laborious, but more explicit, because whenever we see a `.methods` in the future, we will know that we are interacting with a contract (as opposed to other custom objects we may have) directly via Web3.

2) Something to be aware of, is that Web3 v1 unfortunately contaminates txParams in some of its methods. That is, if you pass a txParams object to a transaction's send method, this method may alter txParams. So in the code you will see things like:
`await myContract.methods.someMethod(...PARAMS).send({ ...txParams })`
Notice how a clone of txParams is passed via deconstruction. This cloning protects txParams from such unwanted modifications. We should raise an issue in Web3 for this.

3) So far, 2 tests (tests, not files) that use Sinon stubs where not ported. This is due to the problem not being trivial, and should probably be handled in a separate PR.  More info in a note in the code: https://github.com/zeppelinos/zos/blob/5c0fbcd4eea91554ef9a1b84e336fd4636544152/packages/lib/test/src/utils/Transactions.test.js#L137

4) Web3 v1 uses the same ABI encoder we are using in encodeCall (ethers/abi-coder) which is super strict in terms of how it handles parameters. For example, empty bytes objects can no longer be represented as an empty string `''`. So you will see in the code things like `Buffer.from('')` instead.

5) Web3 v1 uses checksummed addresses, so at the start of most tests, you will see a mapping that looks like:
```accounts = accounts.map(utils.toChecksumAddress)```
Which uses Web3 v1's sub-package 'web3-utils'. This mapping needs to be performed because the tests still use Web3 v0 via Truffle 4, and it does not produce checksummed addresses. `utils.toChecksumAddress` is not only used to map accounts in tests, but any address that comes from a contract obtained via Truffle artifacts, so you should see it in other parts of the code as well.

6) Changes in event parsing. Mainly, a tx's `logs` array was renamed to `events`, and it is now an object. Also, a single event's `args` property was renamed to `returnValues`.

7) Numbers returned as string instead of BigNumbers. So,
```
const value = myContract.getValue();
value.toNumber();
```
becomes
```
const value = myContract.methods.getValue().call();
parseInt(value, 10);
```

8) Contracts.ts default artifacts use to be set synchronously in tests (setup.js), but since getting accounts is only async now in Web3 v1, the from address from getDefaultTxParams() is loaded lazily now, and no longer obtained before starting tests.

9) General minor changes in the data structures of the objects returned by Web3, such as Transaction and TransactionReceipt objects, and in the way the API is called.

10) `contractClass.new(...PARAMS)` previously used to deploy contracts is replaced by `contractClass.deploy(...PARAMS)`. Now, this and all method calls return a TransactionObject, and it is this object that contains methods like `.send()` or `.call()` or `.estimateGas()`.

11) Notice that Web3 v1 uses promises (PromiEvents actually (?)) instead of callbacks, so there is no longer a need to promisify Web3 calls.

12) During the migration, I found that the helper decodeLogs.ts was not really used, so I removed it. Let me know if this was expected for external use and I will put it back (and adapt it to Web3 v1).

## Tasks

- [x] Migrate lib to Web3 v1.
- [x] Migrate cli to Web3 v1.

### CODE REVIEW

**Comment by ajsantander** [comment]
By this point, all of lib's tests are passing with Web3 v1.

**Comment by spalladino** [comment]
RE Change 5: should we use this chance to move from truffle 4 to 5? I'm open to doing so if it makes things easier.

**Comment by ajsantander** [comment]
@spalladino re:
>RE Change 5: should we use this chance to move from truffle 4 to 5? I'm open to doing so if it makes things easier.

In this PR, it might complicate things. I'd say we do it in a different PR, or at least after all tests pass in this PR.

File: package.json [hunk 1]
```diff
@@ -3,6 +3,7 @@
   "private": true,
   "devDependencies": {
     "lerna": "~3.3.2",
-    "tslint": "^5.11.0"
+    "tslint": "^5.11.0",
+    "typescript": "^3.2.2"
```
ajsantander: This is necessary for tslint to work properly on the project.

packages/cli/package.json
```diff
@@ -46,6 +47,7 @@
     "semver": "^5.5.0",
     "truffle-config": "^1.0.6",
     "truffle-resolver": "^4.0.4",
+    "web3": "^1.0.0-beta.37",
```
ajsantander: The Web3 dependency is used in cli only to map addresses to checksummed addresses in tests, so after we migrate to Truffle 5, it could be removed altogether.

packages/lib/test/src/utils/Transactions.test.js
```diff
@@ -110,58 +115,65 @@ contract('Transactions', function([_account1, account2]) {
...
-    });
-
-    it('retries estimating gas up to 3 times', async function () {
-      const stub = sinon.stub(this.instance.initialize, 'estimateGas')
-      _.times(4, i => stub.onCall(i).throws('Error', 'gas required exceeds allowance or always failing transaction'))
-      stub.returns(800000)
-
-      await sendTransaction(this.instance.initialize, [42, 'foo', [1,2,3]]).should.be.rejectedWith(/always failing transaction/);
-    });
+    // TODO: (STUB estimateGas problem) the tests below are disabled because
```
ajsantander: These are the only 2 tests that I couldn't port in this PR.

packages/lib/test/src/validations/Storage.test.js
```diff
@@ -287,8 +287,12 @@ contract('Storage', () => {
...
       const StorageMockChainChild = Contracts.getFromLib('StorageMockChainChild')
       const instance = await StorageMockChainChild.new();
-      const slots = await instance.slots();
-      slots.map(slot => slot.toNumber()).should.deep.eq([0, 1, 3, 5, 7])
+      const slots = await instance.methods.slots().call();
+      const mappedSlots = [];
+      _.mapKeys(slots, (value, key) => { 
+        if(isNaN(parseInt(key, 10))) mappedSlots.push(parseInt(value, 10));
+      });
+      mappedSlots.should.deep.eq([0, 1, 3, 5, 7])
```
ajsantander: Yup. This is not very nice... The new returned value not only contains keys like with string names, but also keys with numerical names (in string format). Not sure why this is. Such keys need to be purged in this test.

tslint.json
```diff
@@ -46,7 +46,8 @@
       "no-consecutive-blank-lines": [true, 1],
       "no-var-requires": [false],
       "curly": [true, "ignore-same-line"],
-      "adjacent-overload-signatures": [false]
+      "adjacent-overload-signatures": [false],
+      "no-string-literal": [false]
```
ajsantander: The linter wasn't allowing things like `myObject['myProp']`.

ajsantander:
Ok, all done!! Lib and cli tests should be passing now with Web3 v1 :tada:

packages/cli/test/scripts/push.test.js
```diff
@@ -266,7 +266,7 @@ contract('push script', function([_, owner]) {
       if (!this.networkFile.appAddress) return;
       const app = await App.fetch(this.networkFile.appAddress);
       const packageInfo = await app.getPackage(libName)
-      packageInfo.version.should.be.semverEqual(libVersion)
+      packageInfo.version.map(Number).should.be.semverEqual(libVersion)
```
spalladino: We could move the `.map(Number)` inside the semverEqual helper actually. But I'm fine with leaving this as is.

packages/cli/test/scripts/push.test.js
```diff
@@ -87,8 +87,8 @@ contract('push script', function([_, owner]) {
     it('should deploy and link contracts that require libraries', async function () {
       const address = this.networkFile.contract('WithLibraryImpl').address;
       const deployed = await WithLibraryImplV1.at(address);
-      const result = await deployed.double(10);
-      result.toNumber().should.eq(20);
+      const result = await deployed.methods.double(10).call();
+      parseInt(result, 10).should.eq(20);
```
spalladino: For the future, I'd consider having a custom helper instead of plain `parseInt`. I wouldn't rule out the fact that web3 may switch to native bignumbers instead of using strings, and then we would need a different function for parsing them. Centralizing how we go from web3-integers (currently strings) to real ints would help in these scenarios.

packages/lib/src/application/App.ts
```diff
@@ -37,27 +37,27 @@ export default class App {
...
   constructor(appContract: ContractWrapper, txParams: any = {}) {
     this.appContract = appContract;
-    this.txParams = txParams;
+    this.txParams = { ...txParams };
   }
 
   public async getPackage(name): Promise<{ package: Package, version: string }> {
-    const [address, version] = await this.appContract.getPackage(name);
-    const thepackage = await Package.fetch(address, this.txParams);
+    const { ['0']: address, ['1']: version } = await this.appContract.methods.getPackage(name).call();
```
spalladino: I'm afraid to ask why this change was needed. How are tuples returned now in web3 1.0?!

packages/lib/src/application/App.ts
```diff
@@ -37,27 +37,27 @@ export default class App {
...
+    const thepackage = await Package.fetch(address, { ...this.txParams });
```
spalladino: Can we push further down the need to clone txParams? I'd try to do it only on the calls to web3, and not when passing them through our intermediate objects.

packages/lib/test/src/utils/Transactions.test.js
spalladino: Use .skip instead of commenting them out, so they show up on the CI as pending

ajsantander:
Rebased to master with some important changes to note:

1)
A recent change from master, in Transactions.ts transformed checks like
```if(txParams.gas) // do something```
to
```if(txParams.gas || Contracts.getArtifactsDefaults) // do something```
and I was getting out of gas errors with that. Considering that ContractFactory.ts' life is not expected to continue for much longer, I solved the issue by doing:
```
  const defaultGas = Contracts.getArtifactsDefaults().gas;
  if (!txParams.gas && defaultGas) txParams.gas = defaultGas;
  if (txParams.gas) // do something
``` 

2.
Since Web3 uses checksummed addresses and that I had to checksum the addresses in Addresses.ts `uint256toAddress`, I started getting invalid address errors for addresses like `0x00da0452681c72356863b9048a52ce2630ed22b5`. The problem was that Web3's getStorageAt removes padded zeroes, so such address was becoming `0xda0452681c72356863b9048a52ce2630ed22b5` (notice the removal of the first two 0's). I fixed that by left padding the incoming "uint256" to 64 characters before actually removing the 0's. As far as I understand, even though this bug should have existed before this PR, it shouldn't have had very bad implications. The resulting code is:
```
export function uint256ToAddress(uint256: string): string {
  const padded = utils.leftPad(uint256.toString(), 64);
  const address = padded.replace('0x000000000000000000000000', '0x');
  return utils.toChecksumAddress(address);
}
```

packages/lib/src/application/App.ts
ajsantander: If a Solidity function returns named tuples, like
(Solidity) `function myFunction() public view returns (address adr1, address adr2)`
You can do
(Javascript) `const { adr1, adr2 } = await contract.methods.myFunction().call()`

But if the solidity function does not name the members of the tuple:
(Solidity) `function myFunction() public view returns (address, address)`
You can do
(Javascript) `const { ['0']: adr1, ['1']: adr2 } = await contract.methods.myFunction().call()`

Since Web3 v0 exposed tuples as arrays, this use to work:
(Javascript) `const [ adr1, adr2 ] = await contract.myFunction()`

ajsantander:
@spalladino all the changes requested by your review have been implemented!

PR #575, opened 2019-01-10
Use gas defaults when present

facuspagnuolo:
Fixes https://github.com/zeppelinos/zos/issues/467


packages/lib/src/utils/Transactions.ts
```diff
@@ -76,16 +76,17 @@ export async function deploy(contract: ContractFactory, args: any[] = [], txPara
  * Uses the node's estimateGas RPC call, and adds a 20% buffer on top of it, capped by the block gas limit.
  * @param contract contract instance to send the tx to
  * @param txParams all transaction parameters (data, from, gasPrice, etc)
+ * @param retries number of deploy retries
```
jcarpanelli: number of *deploy* retries?

packages/lib/src/utils/Transactions.ts
```diff
@@ -97,14 +98,23 @@ export async function sendDataTransaction(contract: ContractWrapper, txParams: a
  */
 async function _sendTransaction(contractFn: GenericFunction, args: any[] = [], txParams: any = {}) {
   // If gas is set explicitly, use it
-  if (txParams.gas) return contractFn(...args, txParams);
+  if (txParams.gas || Contracts.getArtifactsDefaults().gas) return contractFn(...args, txParams);
```
jcarpanelli: :clap:

PR #583, opened 2019-01-16
Update writing_contracts.md

ZumZoom:
Fix broken url


spalladino:
Good catch @ZumZoom, thanks!

PR #588, opened 2019-01-18
Add ProxyAdmin contract to manage proxy instances

jcarpanelli:
Fixes https://github.com/zeppelinos/zos/issues/435, fixes https://github.com/zeppelinos/zos/issues/419.
For more info, see: https://forum.zeppelin.solutions/t/how-the-new-proxyadmin-will-work/172

Implementations in both `lib` and `cli` are pretty much finished. I still have to polish some minor things though.

New project architecture so far:

architecture

Notes:

- I'll be pushing all migration-related code in a different PR after merging this one.
- I'm not  comforable with the new projects' names (`BaseSimpleProject`, `ProxyAdminProject`, `SimpleProject`) :disappointed:. As a matter of fact, I never liked the `SimpleProject` class name :stuck_out_tongue: If anybody comes up with something better,  please let me know.


packages/cli/src/models/TestHelper.ts
```diff
@@ -1,14 +1,14 @@
...
+import { ProxyAdminProject, AppProject } from 'zos-lib';
 
 /**
  * Initializes a zOS application testing and deploying it to the test network,
  * along with dependencies (if specified)
  * @param txParams optional txParams (from, gas, gasPrice) to use on every transaction
  * @param networkFile optional `ZosNetworkFile` object to use, instead of zos.test.json
  */
-export default async function(txParams: any = {}, networkFile?: ZosNetworkFile): Promise<SimpleProject | AppProject> {
+export default async function(txParams: any = {}, networkFile?: ZosNetworkFile): Promise<ProxyAdminProject | AppProject> {
```
spalladino: Good catch. It makes much more sense to use the ProxyAdmin for tests rather than the Simple one.

packages/cli/src/models/files/ZosNetworkFile.ts
```diff
@@ -59,6 +59,10 @@ interface ProviderInterface {
   address?: string;
 }
 
+interface ProxyAdminInterface {
+  address?: string;
```
spalladino: I'd expect the address to be mandatory, same as with the others. The proxyAdmin object itself should be optional on the package file. I mean, this should be ok:
```json
{
  "app": { "address": "0x1" },
  "proxyAdmin": { "address": "0x2" }
}
```

Same as this:
```json
{
  "app": { "address": "0x1" }
}
```

But not this:
```json
{
  "proxyAdmin": { }
}
```

packages/cli/src/models/files/ZosNetworkFile.ts
```diff
@@ -59,6 +59,10 @@ interface ProviderInterface {
   address?: string;
 }
 
+interface ProxyAdminInterface {
```
spalladino: What would be the difference on using an `type` instead of an `interface` here, given it's a rather small object? Something like `type ProxyAdminInfo = { address: string }`. Moreover, we could have `type HasAddress = { address: string }` and use it for Package, Provider, and App as well.

packages/cli/src/models/network/NetworkController.ts
```diff
@@ -428,17 +430,17 @@ export default class NetworkController {
     }
 
     log.info(`Publishing project to ${this.network}...`);
-    const simpleProject = <SimpleProject>(await this.fetchOrDeploy(this.currentVersion));
+    const proxyAdminProject = <ProxyAdminProject>(await this.fetchOrDeploy(this.currentVersion));
     const deployer = new AppProjectDeployer(this, this.packageVersion);
-    this.project = await deployer.fromSimpleProject(simpleProject);
+    this.project = await deployer.fromProxyAdminProject(proxyAdminProject);
```
spalladino: We need to make sure, when we implement the migrations, to first move from simple project to proxy admin project, if the user is coming from an older version.

packages/lib/src/project/ProxyAdminProject.ts
```diff
@@ -0,0 +1,53 @@
...
+    const proxyAdmin = proxyAdminAddress ? await ProxyAdmin.fetch(proxyAdminAddress, txParams) : null;
+    return new this(name, proxyAdmin, txParams);
+  }
+
+  constructor(name, proxyAdmin, txParams) {
+    super(name, txParams);
+    this.proxyAdmin = proxyAdmin;
+  }
+
+  public async createProxy(contractClass, { packageName, contractName, initMethod, initArgs, redeployIfChanged }: ContractInterface = {}): Promise<ContractWrapper> {
```
spalladino: Can't we move these 3 methods (create, upgrade, and changeAdmin) to the parent class? Aren't they shared with SimpleProject?

packages/cli/src/models/files/ZosNetworkFile.ts
jcarpanelli: Entirely agree. Will fix it asap.

spalladino:
Definitely no need to do ASAP

packages/lib/src/project/ProxyAdminProject.ts
jcarpanelli: Mmmh these methods in `ProxyAdminProject` interact with `ProxyAdmin`, while the same methods in `SimpleProject` don't. That doesn't mean that we can refactor a bit these functions, as most of the code is indeed repeated. Creating a protected function in the parent class to set up `initCallData`, `implementationAddress` and `initMethod` might be a good idea :slightly_smiling_face:

spalladino: How about `create`? I don't see any difference between the two, if you create the ProxyAdmin in getProxyAdmin if needed.

packages/cli/src/models/files/ZosNetworkFile.ts
jcarpanelli: After taking a look at this, I believe we need to define the address as not mandatory because `proxyAdmin`, `app`, `package` and `provider` getters may return an empty object if those address are still not created.

packages/lib/contracts/application/App.sol
```diff
@@ -100,73 +100,22 @@ contract App is Ownable {
...
-  function changeProxyAdmin(AdminUpgradeabilityProxy proxy, address newAdmin) public onlyOwner {
-    proxy.changeAdmin(newAdmin);
-  }
-
   /**
    * @dev Creates a new proxy for the given contract and forwards a function call to it.
    * This is useful to initialize the proxied contract.
    * @param packageName Name of the package where the contract is contained.
    * @param contractName Name of the contract.
+   * @param admin address of the proxy administrator.
```
facuspagnuolo:
```suggestion
   * @param admin Address of the proxy administrator.
```

packages/lib/contracts/upgradeability/AdminUpgradeabilityProxy.sol
```diff
@@ -40,17 +40,17 @@ contract AdminUpgradeabilityProxy is UpgradeabilityProxy {
 
   /**
    * Contract constructor.
-   * It sets the `msg.sender` as the proxy administrator.
    * @param _implementation address of the initial implementation.
+   * @param _admin address of the proxy administrator.
```
facuspagnuolo:
```suggestion
   * @param _admin Address of the proxy administrator.
```

packages/lib/contracts/upgradeability/ProxyAdmin.sol
```diff
@@ -0,0 +1,54 @@
+pragma solidity ^0.4.24;
+
+import "openzeppelin-solidity/contracts/ownership/Ownable.sol";
+import "./AdminUpgradeabilityProxy.sol";
+
+contract ProxyAdmin is Ownable {
```
facuspagnuolo: Please add a `@title` natspec here to explain a bit what is this contract for

packages/lib/src/project/BaseSimpleProject.ts
```diff
@@ -0,0 +1,148 @@
...
+interface Dependencies {
+  [packageName: string]: Dependency;
+}
+
+interface Dependency {
+  package: string;
+  version: string;
+}
+
+export default abstract class BaseSimpleProject {
```
facuspagnuolo: If there is no other kind of Base Projects, I would call this one `BaseProject` directly

jcarpanelli: Yeap, there is another Base Project. We have two main structures: `BasePackageProject` - `PackageProject` - `AppProject` and `BaseSimpleProject` - `SimpleProject` - `ProxyAdminProject`, the former in charge of managing published projects and the latter in charge of managing non-published ones. Hopefully, in the near future we'll be able to reduce at least this two structures by merging some classes and finally get to the point where only an `AppProject` and a `ProxyAdminProject` (maybe named `PublishedProject` and `UnpublishedProject`?) will be necessary.

PR #595, opened 2019-01-22
ContractFactory => ZosContract

### LINKED ISSUE #591: Provide plain Web3 (v1) objects in Contracts.ts
Atm, Contracts.ts returns ContractFactory objects, which are basically Web3 Contract objects wrapped with some custom stuff. This "custom stuff" was mostly useful when working with Web3 v0, involving things such as promisifying Web3 calls, etc. Now that we are using Web3 v1, a lot of this code is no longer necessary and actually redundant.

ContractFactroy should return plain Web3 Contracts, and ContractFactory could be removed altogether. Optionally, ZWeb3 no longer promisifies calls, which means we could eliminate this file as well.

Another implication of this would be that many custom interfaces that we created to type Web3 v0 objects, could be eliminated because Web3 v1 is already typed.

ajsantander:
Fix #591

Even though this PR addresses the issue "Provide plain Web3 (v1) objects in Contracts.ts", it does not do _exactly_ what was suggested in the issue. Why? The decision came after a lot of tinkering and exploration on the task at hand. It does not remove ZWeb3 (that could probably handled in a separate PR) and it only transforms ContractFactory into the new "ZosContract".

To understand the reasoning behind the decision, we need to understand what ContractFactory use to do:
1) It provided a few utility methods like `deploy`, `at` and `link` that encapsulated code that either manufactured Web3 Contract objects, or manipulated bytecode.
2) It stored information about the deployment of a contract such as the deployment transaction hash and receipt.
3) It stored information about the schema of the contract, such as bytecode, deployedBytecode, etc, and more elaborate information such as storage layout analysis information.

It did this by manufacturing Web3 Contract object + schema hybrids that were at times difficult to understand, and did so with code that was becoming a little fragile.

To remove ContractFactory would mean to have to store the items of 2 & 3 somewhere else, and to provide the utilities of 1 somewhere else as well.

Instead, ZosContract performs all the items 1 to 3, but in a much less convoluted manner, and one in which responsibilities are more clearly separated:
1) The utility methods are still stored in the ZosContract (ex ContractFactory) class, but greatly simplified.
2) All deployment information is stored in a small `zosData` object, which is injected into the Web3 Contract object itself. Otherwise the Web3 Contract object is unaltered.
3) ZosContract stores the schema, but it is the schema that stores all the additional schema-related information. This schema is an instance of ZosContractSchema, which extends SolidityContractSchema, and clearly shows where the zOS specific information is stored.

The result is that Contracts.ts simply produces the schema and passes it along to ZosContract. ZosContract stores the schema and has the ability to produce minimalistically augmented Web3 Contract objects.


packages/lib/src/artifacts/ZosContract.ts
```diff
@@ -0,0 +1,86 @@
...
+  }
+
+  // Extends Web3's Contract interface.
+  public schema: ContractSchema;
+  public binary: string; // TODO: remove
+  public deployedBinary: string; // TODO: remove
+  public storageInfo: StorageLayoutInfo;
+  public warnings: any;
+  public deploymentTransactionReceipt: TransactionReceipt;
+  public deploymentTransactionHash: string;
```
spalladino: Can we store all this under a separate object? I mean, web3 contract has all methods under "methods", events under "events", etc; it'd be nice to bundle all the extra info we need under one or two objects that contain all of them, to avoid polluting the contract interface.

ajsantander: The latest version of ZosContract stores schema related stuff (storageInfo and warnings) under ZosContract.schema and deployment related stuff in a `zosData` object, injected into the Web3 Contract object.

packages/lib/src/artifacts/ZosContract.ts
```diff
@@ -0,0 +1,49 @@
...
+    const self = this;
+    return new Promise(function(resolve, reject) {
+      const tx = contract.deploy({data: self.schema.linkedBytecode, arguments: args});
+      const zosData: any = {};
+      tx.send({ ...options })
+        .on('error', (error) => reject(error))
+        .on('receipt', (receipt) => zosData.deploymentTransactionReceipt = receipt)
+        .on('transactionHash', (hash) => zosData.deploymentTransactionHash = hash)
+        .then((instance) => {
+          instance.zosData = zosData;
```
spalladino: If we are just storing txReceipt and txHash, how about storing it in a "deployment" node, instead of naming it zosData?

ajsantander: Yeah, that'd make sense. However "zosData" is more explicit in the sense that the mere name suggests that this was something injected by zOS. People could confuse "deployment" as part of Web3 Contract's api. Also, we might want to inject additional data in the future that is not only related to deployment. I suggest:
```
zosData: {
  deployment: {...}
}
```

ajsantander:
@facuspagnuolo regarding your points:

> I don't get why you needed to change all the x.address to x._address
This is because x is an instance of Web3 Contract, which doesn't have an `.address` property but `._address` or `.options.address`. Of the two, I chose `_address` because it seemed more concise. I know that it doesn't look nice, but it is done for the sake of respecting the Web3 Contract api.

> I'd add some methods/properties to encapsulate ZosContract's behavior like bytecode, contractName, etc... otherwise, you will always have to know from the outside that there is something called schema within ZosContract
I don't think I like that idea. Knowing that there is a schema property from the outside should be trivial because of Typescript, and schema is also typed, so you would get completion for schema.contractName for example. I wanted to keep all non-Web3Contract stuff contained within the schema object, and I wouldn't like to have to forward the schema interface with redundant methods.

packages/lib/src/artifacts/ZosContract.ts
spalladino: Personally, I'd prefer to see it listed as just `deployment`, without the `zosData` step. I think that the risk of colliding with a future extension to the web3 contract is low.

spalladino:
> This is because x is an instance of Web3 Contract, which doesn't have an .address property but ._address or .options.address. Of the two, I chose _address because it seemed more concise. I know that it doesn't look nice, but it is done for the sake of respecting the Web3 Contract api.

How about adding a function to the object that returns the .options.address property, just as a shorthand, since we use it so much? I think it'd be handy, I'm reluctant to relying on something that is intended to be private or internal (as the leading underscore seems to imply).

> I don't think I like that idea. Knowing that there is a schema property from the outside should be trivial because of Typescript, and schema is also typed, so you would get completion for schema.contractName for example. I wanted to keep all non-Web3Contract stuff contained within the schema object, and I wouldn't like to have to forward the schema interface with redundant methods.

Agree with you here!

ajsantander:
@spalladino 

> How about adding a function to the object that returns the .options.address property, just as a shorthand, since we use it so much? I think it'd be handy, I'm reluctant to relying on something that is intended to be private or internal (as the leading underscore seems to imply).

Yes, I see your point about the "nono" of accessing an underscored property. I kind of liked the idea of knowing that I was interacting with a contract instead of with a custom zOS object, whenever I saw the underscore. If you confirm, I can make a second modification (the zosData injection being the 1st) to the Web3 Contract object that aliases "address" to `.options.address`. Additionally, I would raise an issue in Web3, because I see no reason for not simply exposing the address as they did with Web3 0.x.

spalladino:
Agree with both changes, thanks Ale!!!

facuspagnuolo:
> How about adding a function to the object that returns the .options.address property, just as a shorthand, since we use it so much? I think it'd be handy, I'm reluctant to relying on something that is intended to be private or internal (as the leading underscore seems to imply).
Let's do that

> I don't think I like that idea. Knowing that there is a schema property from the outside should be trivial because of Typescript, and schema is also typed, so you would get completion for schema.contractName for example. I wanted to keep all non-Web3Contract stuff contained within the schema object, and I wouldn't like to have to forward the schema interface with redundant methods.
Disagree, for the one who developed it is trivial, not for me at least, breaking encapsulation is not a good idea at all.

ajsantander:
@spalladino @facuspagnuolo thx for your review!

I've made the following changes:

1) Removed the need to use `._address` on Web3 Contract objects. ZosContract modifies the contract instances by injecting an `.address` property. I've also created an issue in Web3 to see if this can be changed: https://github.com/ethereum/web3.js/issues/2274

2) Injected `zosData` now looks like this:
```
Web3 Contract {
  ...
  zosData: {
    deployment: {
      transactionReceipt: {...}
      transactionHash: "..."
    }
  }
}
```

3) ZosContractSchema and SolidityContractSchema were deleted, and instead, ZosContract now breaks up the incoming generic schema object into a set of properties on itself:
```
export default class ZosContract {

  // Solidity contract schema properties.
  public schemaVersion: string;
  public contractName: string;
  public abi: any[];
  public bytecode: string;
  public deployedBytecode: string;
  public sourceMap: string;
  public deployedSourceMap: string;
  public source: string;
  public sourcePath: string;
  public ast: any;
  public legacyAST: any;
  public compiler: any;
  public networks: any;
  public updatedAt: string;

  // Custom schema properties.
  public linkedBytecode: string;
  public linkedDeployedBytecode: string;
  public warnings: any;
  public storageInfo: StorageLayoutInfo;

  constructor(schema: any) {
    Object.assign(this, schema);
  }

// ...
```

spalladino:
If I understand correctly, `ZosContract` is a custom class that only holds schema info for a contract class, and has `deploy` & `at` methods that spawn `web3.contract`s, adding a `zosData` to them with deployment info. Users of `zos-lib` get `ZosContract`s if they use our `Contracts.getFrom*()` methods, and get augmented `web3.contract`s if they use `Project.createProxy()` (or other methods that create instances).

While I'm fine with the latter (users should get something as similar as possible to a `web3.contract` when using a Project or similar), I'm not sold on exposing something entirely custom from the `Contracts.getFrom()` set of methods.

I'd like to hear your thoughts about returning `web3.contract`s from `Contracts.getFrom()`, but augmented with a `schema` (or `artifact`, or other name) property that has the info we require throughout the CLI (such as bytecode, ast, etc), instead of a custom `ZosContract` instance. This way, end users end up working with the same classes as we do for managing contract classes.

If we go that way, then `ZosContract` does not need to be a class by itself, but actually a file that contains:
- A function for creating a schema-augmented `web3.contract` from a truffle-generated artifact
- A static function for deploying the augmented contract and injecting deployment info
- Type definitions for the schema and deployment info

In very-pseudocode, would be sth like:

```
function createContractClass(schema) {
  // this function is called from Contracts to return a web3.contract-like object
  const c = new web3.eth.Contract(schema.abi, null, { data: schema.binary, ... defaultTxParams })
  c.schema = schema
  return c
}

async function deploy(contract, params, txParams) {
  // this is used internally by us to deploy and keep deployment info
  let receipt, hash;
  return new Promise(function(resolve, reject) {
    contract.deploy(... params).send(txParams)
      .on('error', (error) => reject(error))
      .on('receipt', (r) => receipt = r)
      .on('transactionHash', (h) => hash = h)
      .then((instance) => {
        instance.schema = contract.schema
        instance.deployment = { hash, receipt }
        addAddressProperty(contract)
        resolve(instance);
      })
      .catch((error) => reject(error));
  })
}

async function at(contract, address) {
  // this is used internally by us to have the happy address prop
  const instance = await contract.at(address)
  addAddressProperty(contract)
  return instance
}

function addAddressProperty(instance) {
  // this function is not exported by this module
  Object.defineProperty(instance, 'address', { get: function() { return this.options.address; } })
}
```

Regarding augmenting `web3.contract`, I'd have only two additional nodes that we add to the object: `schema` and `deployment`, and avoid zosData as a name. And for adding the address convenience method, I'd use a getter, to ensure it stays in sync with `options.address`, as shown above.

I'd like to hear both of your thoughts @facuspagnuolo @ajsantander. Agree that this is a sensitive change, and is worth to be discussed.

ajsantander:
Once again, thanks for your analysis @facuspagnuolo and @spalladino .

Putting all your comments together, I think they would result in the following changes:

- Instead of Contracts.ts returning a ZosContract object, it produces a modified Web3 Contract object that contains the following injections for: `.new()`, `.at()`, `.link()`, an `.address` getter that is not just a stored value but a proper getter, schema getters like `.contractName`, `.ast`, etc (only the ones the code actually needs) and `.deployment.transactionReceipt` plus `deployment.transactionHash`.
- The injections are not contained on a zosData object at all, but are loose in the Web3 Contract object.
- The injections are performed within Contracts.ts, so ZosContract.ts can be deleted.
- (optional) The modified Web3 Contract object could adhere to a Typescript interface which extends the original Web3 Contract interface (not the class), and so all our injections would be typed.

Please let me know if you would like to move forward with these changes.

spalladino:
> Instead of Contracts.ts returning a ZosContract object, it produces a modified Web3 Contract object that contains the following injections: .new(), .at(), .link(), an .address getter that is not just a stored value but a proper getter, schema getters like .contractName, .ast, etc (only the ones the code actually needs) and .deployment.transactionReceipt plus deployment.transactionHash.

I'd prefer keeping the contractName, ast, etc getter inside a `schema` object, so it's `contract.schema.ast`, to avoid polluting the top level; but I'm fine with either way. As for the additional methods, why not make them just plain functions that receive an instance of the contract and deploy it, or specialize it at an address, or link it; instead of making them instance functions? This way, we don't add new stuff to the contract object.

> The injections are not contained in a zosData object at all, but are loose in the Web3 Contract object.

Agree on not having zosData.

> The injections are performed within Contracts.ts, so ZosContract.ts can be deleted.

Either way is fine for me. As long as `Contracts` returns the web3 contract, it's ok!

> (optional) The modified Web3 Contract object could adhere to a Typescript interface which extends the original Web3 Contract interface (not the class), and so all our injections would be typed. The interface could be named ZosContract and would be 100% compatible with Web3 Contract (an extension of it).

Would love to have this in place!

ajsantander:
> I'd prefer keeping the contractName, ast, etc getter inside a schema object, so it's contract.schema.ast, to avoid polluting the top level; but I'm fine with either way. As for the additional methods, why not make them just plain functions that receive an instance of the contract and deploy it, or specialize it at an address, or link it; instead of making them instance functions? This way, we don't add new stuff to the contract object.

Sure, I could do both, but note that this implies going against two requests from @facuspagnuolo .
- He wanted to not have to do `myWeb3Instance.schema.<something>` and instead do `myWeb3Instance.<something>`. Pls see his comments above.
- He didn't want to have to do `deployContract(myWeb3Instance)` in favor of maintaining the Truffle style `myWeb3Instance.new()` syntax (note that the utiltity function style would imply a lot of changes in the usage of the interface, specially in tests).

spalladino:
> He didn't want to have to do deployContract(myWeb3Instance) in favor of maintaining the Truffle style myWeb3Instance.new() syntax (note that the utiltity function style would imply a lot of changes in the usage of the interface, specially in tests).

Is it possible to automate that change in our tests? If not, and it's going to require a lot of work, I'm ok with leaving them as part of the object, since they don't clash with other web3 contract methods. As for maintaining a truffle-like syntax, I wouldn't enforce it as a requirement, since we are now changing into web3 contracts.

facuspagnuolo:
Given that we are now planning to return a web3 contract, I'm ok with wrapping all the schema related info within a single property. Looks good to me guys!

ajsantander:
I don't see an easy way to automate the conversion of `deployContract(myWeb3Instance)` to `myWeb3Instance.new()` only in tests. I'm also in favor of not trying to reproduce the Truffle syntax. However, I am not totally convinced about having to import utility functions to deploy or retrieve a deployed contract. It could be annoying.

spalladino:
I have a hunch that it's doable via regexes, replacing `([A-Za-z_][A-Za-z0-9_]+)\.deploy\(\)` with `deploy($1)`, and something a bit smarter if deploy had args (such as replacing by `deploy($1,`). This would allow us more flexibility in the future if we need to change anything yet again (as we are doing with the Transactions module for sending txs). However, if the above doesn't work, I'm ok with adding `new`, `at`, and `link` methods to the object if needed (let's avoid `deploy` to not overwrite the existing web3 contract method).

spalladino:
And by the way, I didn't mean to change it *only in tests*! I thought that tests was the most painful place to make the change, that's all.

ajsantander:
Ok. Implementing the changes...

I hope that I don't dissapoint you, but I'm going to choose the easy way for now, that is, adding `.new()`, `.at()` methods into the Web3 Contract object.

spalladino:
> I hope that I don't dissapoint you, but I'm going to choose the easy way for now, that is, adding .new(), .at() methods into the Web3 Contract object.

Not my favourite option, but good enough! Go for it!

ajsantander:
Ok, all discussed changes implemented. I don't know if it's just me (because I'm not that used to Javascript) but the resulting ZosContract.ts code feels kinda weird. Is it me softening up and opening my soul into the wonders of Javascript, or is it just bad code?

ajsantander:
NOTE: The new typing enforcement of ZosContract detected a problem that wasn't failing before in the tests: https://github.com/zeppelinos/zos/blob/e3c8b53520f5c9e4b9758afc1952b90246b64579/packages/cli/src/models/network/NetworkController.ts#L488

Apparently this code is saving proxy creation info per network, but the property `.network_id` does not exist on non-truffle artifacts. My solution was to get the current network id from ZWeb3 and use that as the storing key.

packages/lib/src/artifacts/ZosContract.ts
```diff
@@ -0,0 +1,111 @@
...
+ * ZosContract is an interface that extends Web3's Contract interface, adding some properties and methods like:
+ * address getter: retrieves the deployed address
+ * schema: compilation artifacts
+ * new(): deployes a new contract
+ * at(): retrieves a deployed contract at the specified address
+ * link(): links libraries in a contract schema
+ */
+export default interface ZosContract {
+
+  // Web3 Contract interface.
```
spalladino: Is it possible to load these typings from web3 and having ZosContract extend from that interface, to avoid copy-pasting?

packages/lib/src/artifacts/ZosContract.ts
```diff
@@ -0,0 +1,111 @@
...
+
+/*
+ * ZosContract is an interface that extends Web3's Contract interface, adding some properties and methods like:
+ * address getter: retrieves the deployed address
+ * schema: compilation artifacts
+ * new(): deployes a new contract
+ * at(): retrieves a deployed contract at the specified address
+ * link(): links libraries in a contract schema
+ */
+export default interface ZosContract {
```
spalladino: Is it easy with the IDE you are using to rename ZosContract to plainly Contract?

packages/lib/src/artifacts/ZosContract.ts
ajsantander: Unfortunately, Web3's existing declarations define a class instead of an interface for the Contract type, and Typescript doesn't allow to extend clases from `.d.ts` files. Believe me, I tried everything! We could raise another issue in Web3 with this request.

spalladino: Damn! Yes, let's open an issue then, though I wouldn't expect any answer soon; I think they have more pressing ones.

packages/lib/src/artifacts/ZosContract.ts
spalladino: Yes, let's rename. I'll merge anyway now so we can start testing though, but it'd be good to have another PR renaming it.

examples/lib-complex/test/DonationsV1.test.js
```diff
@@ -4,17 +4,12 @@ require('./setup')
 import { Contracts, encodeCall } from 'zos-lib'
```
pcowgill: @ajsantander @spalladino @facuspagnuolo I think the import of `encodeCall` here can be removed in light of the changes in this PR

spalladino: Thanks for spotting this, Paul! We'll probably make a cleanup pass over all examples soon, and will take care of this then.

PR #607, opened 2019-01-25
Fixes in storage layout checks (master)

spalladino:
Port of #606 to master.

---

Fixes two issues with storage layout checks:

1- Certain append operations were not being properly recognized, they were identified as insertions and caused false positives. This was found during the latest deployment of openzeppelin-eth, and added as a regression test. The problem was that the edition distance matrix always preferred a diagonal path with zero cost, even if this meant an accumulated cost higher than taking another paths. This is typically not an issue with levenshtein, since diagonal paths are always optimal, but here we break that invariant as we have free appends.

2- The algorithm was failing if either the initial or target contract had no storage whatsoever. Not only something different from a matrix was being returned, but also the cost of the first row of the matrix was being incorrectly calculated, as appends were not being taken into consideration.

3- The first row of the distance matrix was being incorrectly calculated, since it must add insertion costs up until the point where the value stays constant due to appends.

### DIFF
--- a/packages/lib/contracts/mocks/StorageMocks2.sol
+++ b/packages/lib/contracts/mocks/StorageMocks2.sol
@@ -47,5 +47,3 @@ contract StorageMockChainChild is StorageMockChainA2, StorageMockChainB {
     }
   }
 }
-
-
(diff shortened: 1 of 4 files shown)


spalladino:
Found another minor issue, please wait until fixed for merging.

spalladino:
@facuspagnuolo ready to review!

PR #613, opened 2019-01-30
Optimize initializer modifier

k06a:
We have a hierarchy with more than 10 contracts, this optimization allowed us to significantly reduce gas cost

### DIFF
--- a/packages/lib/contracts/Initializable.sol
+++ b/packages/lib/contracts/Initializable.sol
@@ -32,12 +32,18 @@ contract Initializable {
     require(initializing || isConstructor() || !initialized, "Contract instance has already been initialized");
 
     bool wasInitializing = initializing;
-    initializing = true;
-    initialized = true;
+    if (!wasInitializing) {
+      initializing = true;
+    }
+    if (!initialized) {
+      initialized = true;
+    }
 
     _;
 
-    initializing = wasInitializing;
+    if (!wasInitializing) {
+      initializing = false;
+    }
   }
 
   /// @dev Returns true if and only if the function is running in the constructor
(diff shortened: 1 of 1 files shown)


spalladino:
@frangio can I ask you for your input here? The change seems harmless, and should help save gas, but I want to make sure I'm not missing anything.

packages/lib/contracts/Initializable.sol
```diff
@@ -32,12 +32,18 @@ contract Initializable {
     require(initializing || isConstructor() || !initialized, "Contract instance has already been initialized");
 
     bool wasInitializing = initializing;
-    initializing = true;
-    initialized = true;
+    if (!wasInitializing) {
```
frangio:
```suggestion
    if (!initializing) {
```

This is more straightforward, less confusing, and makes it clear that this `if` block and the next one are doing exactly the same thing.

frangio: I personally think that `if (initializing != true)` would result in more self-evident code. What do you guys think?

k06a:
Maybe you would also like this version:
```solidity
if (!initializing) {
  initialized = true;
  initializing = true;
  _;
  initializing = false;
} else {
  _;
}
```

k06a:
image
https://blog.polymath.network/solidity-tips-and-tricks-to-save-gas-and-reduce-bytecode-size-c44580b218e6

k06a:
@frangio what do you think?

frangio:
It seems like `_` will duplicate the code. I'd keep the PR's original proposal.

k06a:
@frangio fixed! Any suggestions?

spalladino:
> Note that these optimizations won't be as significant after Constantinople, due to [EIP 1283](https://eips.ethereum.org/EIPS/eip-1283). However, I still think this is worth merging.

@frangio given 1283 was reverted, do you have any new considerations on this PR?

frangio:
It makes it even more worth merging then.

k06a:
Ok, I'm ready if you're ready :)

spalladino:
Merged! Thanks a lot @k06a for the contribution!

PR #619, opened 2019-01-31
fix: change pragma to be truffle-flattener compatible

chentschel:
While extending Initializable pragma version to support solc 0.5 on [commit](https://github.com/zeppelinos/zos/pull/577/commits/912bf2dabb571716c40bb86e429153bda6ef3ad8) breaks truffle-flattener compatibility due to **'Only pinned or ^ versions are supported'** error.

### DIFF
--- a/packages/lib/contracts/Initializable.sol
+++ b/packages/lib/contracts/Initializable.sol
@@ -1,4 +1,4 @@
-pragma solidity >=0.4.24 <0.6.0;
+pragma solidity ^0.4.24;
 
 
 /**


spalladino:
Thanks for the PR @chentschel! The rationale for the version pragma we're using is to support contracts both in Solidity 0.4 and 0.5. I think the best place to fix this is actually on the flattener: let me contact @alcuadrado and see what we can do. Thanks for bringing this to our attention!

alcuadrado:
Hi,

There's no need to change them here. I'll release a new version this week that supports any kind of pragma. 

@frangio figured out that there's no need to mess with the pragmas in `truffle-flattener`, as `solc` can handle multiple pragmas as long as they don't contradict each other.

frangio:
The fix is being tracked in https://github.com/nomiclabs/truffle-flattener/issues/37.

alcuadrado:
Hey! I've just released version 1.3.0 of `truffle-flattener`, which supports any kind of pragma, and fixes this issue.

facuspagnuolo:
Thank you both @alcuadrado and @chentschel! Closing this one in favour of https://github.com/zeppelinos/zos/pull/682

PR #620, opened 2019-01-31
Set up a shared pre-commit hook that runs a linter when committing changes

### LINKED ISSUE #590: Run linter on CI tests
We are now using TSLint to define linting rules in the entire project (applicable for now in all .ts files): https://github.com/zeppelinos/zos/blob/master/tslint.json

These rules are not optional, and should be respected by anyone contributing to the project. To enforce this, I suggest that we run `npm run lint` in the CI tests.

Also, as suggested by @spalladino, "Another thing that would help is to share a git pre-commit hook to run the linter check".

ajsantander:
Fix #590 

This PR adds a global prepare script in `./package.json` that configures the location of git hooks in this repo to the path `./.githooks` instead of the default `.git/hooks`. That way, git hooks can be shared with all collaborators.

Then, a pre-commit hook is added, which simply runs `lerna run lint`. This command will run tslint on the cli and lib packages and, as a result, will abort commits which have linter failures.

The advantages of using a git pre-commit hook for this are:
- It doesn't slow down local testing, which would be that case if we added the linting routing to each package's `npm test` scripts.
- It avoids having to find out that your CI tests failed because of linting issues.
- Linting is performed locally, and only when you wish to commit changes to the repo.

### DIFF
--- a/.githooks/pre-commit
+++ b/.githooks/pre-commit
@@ -1,0 +1,2 @@
+#!/bin/sh
+lerna run lint
--- a/package.json
+++ b/package.json
@@ -1,6 +1,9 @@
 {
   "name": "root",
   "private": true,
+  "scripts": {
+    "prepare": "git config core.hooksPath .githooks"
+  },
   "devDependencies": {
     "lerna": "~3.3.2",
     "tslint": "^5.11.0",
--- a/packages/cli/src/models/files/ZosNetworkFile.ts
+++ b/packages/cli/src/models/files/ZosNetworkFile.ts
@@ -51,7 +51,7 @@ export interface DependencyInterface {
   customDeploy?: boolean;
 }
 
-type addressWrapper = { address?: string };
+interface AddressWrapper { address?: string; }
 
 export default class ZosNetworkFile {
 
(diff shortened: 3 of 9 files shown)


packages/cli/src/models/files/ZosNetworkFile.ts
```diff
@@ -51,7 +51,7 @@ export interface DependencyInterface {
   customDeploy?: boolean;
 }
 
-type addressWrapper = { address?: string };
+interface AddressWrapper { address?: string; }
```
spalladino: Is this part of the linter changes? If not, I'd move it out of the PR.

package.json
```diff
@@ -1,6 +1,9 @@
 {
   "name": "root",
   "private": true,
+  "scripts": {
+    "prepare": "git config core.hooksPath .githooks"
```
spalladino: Question: why on the `prepare` script? This is often used when packing a package, or when installing a dependency that needs to be compiled. Should we do this on a different script?

.githooks/pre-commit
```diff
@@ -0,0 +1,2 @@
+#!/bin/sh
+lerna run lint
```
spalladino: Does this require to have lerna globally installed? Does it work with local lerna as well?

spalladino:
Well, it was merged before I could submit. @ajsantander please take a look at the comments above, and only address them if you think they are worthwhile.

PR #621, opened 2019-01-31
Migration of zosversion from 2 to 2.2

jcarpanelli:
Highlights in this PR:
- Interactive migration for `create`, `update`, `set-admin` and `publish` commands.
- `migrate` function in charge of migrating a `SimpleProject` to a `ProxyAdminProject`, or an old published project (with the deprecated version of `App.sol` and `AdminUpgradeabilityProxy.sol` contracts) to a new `AppProject` with `ProxyAdmin`.
- A migration function in `lib`, in charge of doing a raw transaction to the blockchain  to interact with the deprectated `App.sol#changeProxyAdmin` function, and change the admin of a proxy from `App` to `ProxyAdmin`.
- Update of both `zos.json` and `zos.network.json` `zosversion` attribute.

EDIT: I also wrapped all `Transactions.js` functions into an object. I think that having standalone functions is ok if there are not many of them, but having ~16 is a bit annoying. Also, it is more readable if we have something such as `Transactions.deploy()` or `Transactions.sendRawTransaction()` instead of a standalone `deploy()`.


package.json
```diff
@@ -1,6 +1,9 @@
 {
   "name": "root",
   "private": true,
+  "scripts": {
+    "prepare": "git config core.hooksPath .githooks"
+  },
```
facuspagnuolo: Don't know why github is showing these stuff related to linter hook, since they were already merged into master.

packages/cli/src/commands/create.ts
```diff
@@ -20,6 +22,8 @@ const register: (program: any) => any = (program) => program
   .action(action);
 
 async function action(contractFullName: string, options: any): Promise<void> {
+  const zosversion = await ZosPackageFile.getZosversion();
+  if (!await willMigrateProjectIfNeeded(zosversion)) return;
```
facuspagnuolo: I would use sth more friendly like `hasToMigrateProject` or so, not a blocker tho.

facuspagnuolo: Very nit, but we can move this behavior into `/bin/options.ts` doing sth like:
```js
program.Command.prototype.withMigrateAction = function(): Command {
  return this.action(() => {
     const zosversion = await ZosPackageFile.getZosversion();
     if (!await willMigrateProjectIfNeeded(zosversion)) process.exit(0);
  });
};
```

packages/cli/src/models/files/ZosPackageFile.ts
```diff
@@ -17,6 +18,11 @@ export default class ZosPackageFile {
     publish: boolean;
   };
 
+  public static getZosversion(fileName: string = 'zos.json'): string | null {
+    const file = fs.parseJsonIfExists(fileName);
+    return file.zosversion || null;
+  }
```
facuspagnuolo: I think we should move the `checkVersion` thing outside the constructor, I imagine that's why you needed to put such behavior as a static method. Probably, we can move to a "before each" statement for any command.

packages/cli/src/models/files/ZosPackageFile.ts
```diff
@@ -137,7 +147,20 @@ export default class ZosPackageFile {
   }
 
   public write(): void {
-    fs.writeJson(this.fileName, this.data);
-    log.info(`Successfully written ${this.fileName}`);
+    if(this._hasChanged()) {
+      const exists = this._exists();
+      fs.writeJson(this.fileName, this.data);
+      exists ? log.info(`Updated ${this.fileName}`) : log.info(`Created ${this.fileName}`);
+    }
```
facuspagnuolo: 👍

packages/cli/src/models/files/ZosPackageFile.ts
```diff
@@ -137,7 +147,20 @@ export default class ZosPackageFile {
...
   }
+
+  public _hasChanged(): boolean {
+    const currentPackgeFile = fs.parseJsonIfExists(this.fileName);
+    return !isEqual(this.data, currentPackgeFile);
+  }
+
+  public _exists(): boolean {
+    return !!fs.parseJsonIfExists(this.fileName);
```
facuspagnuolo:
```suggestion
    return fs.exists(this.fileName);
```

packages/cli/src/models/files/ZosVersion.ts
```diff
@@ -1,13 +1,42 @@
...
   else if (isUndefined(version)) {
     throw Error(`zos version identifier not found in ${where}. This means the project was built with an older version of zos (1.x), and needs to be upgraded. Please refer to the documentation at https://docs.zeppelinos.org for more info.`);
-  }
-  else {
+  } else if (!isCurrentMajor(version) || (!isCurrentMinor(version) && !isUndefinedMinor(version))) {
     throw Error(`Unrecognized zos version identifier ${version} found in ${where}. This means the project was built with an unknown version of zos. Please refer to the documentation at https://docs.zeppelinos.org for more info.`);
   }
 }
+
+export function isNotMigratableZosversion(version: string): boolean {
```
facuspagnuolo: Since you always use it as `!isNotMigratableZosversion...`, I'd do 
```suggestion
export function isMigratableZosversion(version: string): boolean {
```

packages/cli/src/models/network/NetworkController.ts
```diff
@@ -422,13 +431,45 @@ export default class NetworkController {
...
+      this._tryRegisterProxyAdmin(proxyAdmin.address);
+
+      await allPromisesOrError(map(proxies, async (proxy) => {
+        const proxyInstance = await Proxy.at(proxy.address);
+        const currentAdmin = await proxyInstance.admin();
+        if (currentAdmin !== proxyAdmin.address) {
+          if (this.appAddress) {
+            return AppProxyMigrator(this.appAddress, proxy.address, proxyAdmin.address, this.txParams);
+          } else {
+            const simpleProject = await new SimpleProject(this.packageFile.name, this.txParams);
```
facuspagnuolo:
```suggestion
            const simpleProject = new SimpleProject(this.packageFile.name, this.txParams);
```

packages/cli/src/models/network/NetworkController.ts
```diff
@@ -515,6 +558,7 @@ export default class NetworkController {
 
   // Proxy model
   public async upgradeProxies(packageName: string, contractAlias: string, proxyAddress: string, initMethod: string, initArgs: string[]): Promise<ProxyInterface[]> {
+    if (!isNotMigratableZosversion(this.currentZosversion)) await this.migrate();
```
facuspagnuolo: Nit: we can extract the if statement to a private method `await _migrateIfNeeded()` or so

packages/cli/src/utils/prompt-migration.ts
```diff
@@ -0,0 +1,24 @@
...
+      });
+    });
+    if (!isValidResponse(response)) return willMigrateProjectIfNeeded(zosversion);
+    return response === 'y' || response === 'Y';
+  }
+  return true;
+}
+
+function isValidResponse(response): boolean {
+  return response == 'y' || response == 'Y' || response == 'n' || response == 'N';
```
facuspagnuolo:
```suggestion
  return response === 'y' || response === 'Y' || response === 'n' || response === 'N';
```

packages/lib/src/utils/Migrator.ts
```diff
@@ -0,0 +1,11 @@
+import Transactions from './Transactions';
+import encodeCall from '../helpers/encodeCall';
+import Logger from '../utils/Logger';
+
+const log: Logger = new Logger('Migrator');
+
+export default async function migrate(appAddress: string, proxyAddress: string, proxyAdminAddress: string, txParams: any = {}): Promise<void> {
+  const data = encodeCall('changeProxyAdmin', ['address', 'address'], [proxyAddress, proxyAdminAddress]);
```
facuspagnuolo: Should we add a line saying sth like 'will change proxy admin...'?

packages/lib/test/src/utils/Transactions.test.js
```diff
@@ -129,16 +128,15 @@ contract('Transactions', function(accounts) {
...
 
-            const receipt = await sendTransaction(this.instance.methods.initialize, DEFAULT_PARAMS);
+            const receipt = await Transactions.sendTransaction(this.instance.methods.initialize, DEFAULT_PARAMS);
             await assertGas(receipt.transactionHash, 800000 * 1.25 + 15000);
           });
 
-          it.skip('retries estimating gas up to 3 times', async function () {
-            const stub = sinon.stub(this.instance.methods.initialize, 'estimateGas')
+          it('retries estimating gas up to 3 times', async function () {
+            const stub = sinon.stub(Transactions, '_calculateActualGas');
```
facuspagnuolo: 👍

packages/cli/src/commands/create.ts
jcarpanelli: I'm not confortable with the `process.exit(0)` thing, but it's definitely simpler and clean.

packages/cli/src/models/files/ZosPackageFile.ts
jcarpanelli: Yeap... I didn't want to change the actual behavior of how we are checking `1.x` and `2` versions. OTOH, This is the only way that came into my mind to get the `zosversion` *before* initializing a `ZosPackageFile` inside the scripts functions.

PR #623, opened 2019-01-31
Adds ability to receive false booleans via the cli

### LINKED ISSUE #558: Ability to pass false booleans via the cli
Atm, calling something like 

```
zos create MyContract --init --args false
```

`cli/src/utils/input.ts` will parse false as the string "false", and when this value is passed to `lib/helpers/encodeCall.ts`, the encoding will interpret "false" as the boolean true.

`input.ts` should account for such booleans (as opposed to `encodeCall.ts` doing it in order to maintain compatibility with Web3).

ajsantander:
Fix #558 

This PR solves the issue "Ability to pass false booleans via the cli" by:
1) Modifying input.ts so that incoming string values that are either "false" or "true" are converted into `false` and `true` literals respectively. Notice that this makes booleans the only type of values that do not exit as strings from input.ts. The file input.test.js is also modified to test for this change.
2) Adds a test to the create command, ensuring that a "false" argument does indeed get parsed into `false` instead of `true`.
3) Adds tests to the create script, with a new Booleans.sol contract, ensuring that calling the script with boolean literals produces the expected results.

This does not modify encodeCall in any way to maintain compatibility with Web3's encoder. If "false" is passed to encodeCall, it will still interpret it as `true`.

Despite my belief that this should solve the issue at hand, please note that we could go a little further, and introduce a new type of test in the suite that tests end-to-end parameter parsing, from the cli, to the values obtained back from a contract. The implemented solution touches a few isolated components that, _in theory_, should produce the expected result, but we don't have a way to test all the components together at once.

### DIFF
--- a/packages/cli/src/utils/input.ts
+++ b/packages/cli/src/utils/input.ts
@@ -10,7 +10,17 @@ export function parseArgs(args: string): string[] | never {
     const MATCH_WORDS = /(^|([,[]\s*))(\w+)/g;
     args = args.replace(MATCH_WORDS, '$2"$3"'); // replace non quoted number by quoted number
 
-    return JSON.parse('[' + args + ']');
+    // Parse string to array.
+    let parsedArgs = JSON.parse('[' + args + ']');
+
+    // Replace boolean strings with boolean literals.
+    parsedArgs = parsedArgs.map((value) => {
+      if(value === 'false') return false;
+      else if(value === 'true') return true;
+      return value;
+    });
+
+    return parsedArgs;
   } catch (e) {
     throw Error(`Error parsing arguments: ${e}`);
   }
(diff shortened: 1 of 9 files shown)


packages/cli/src/utils/input.ts
```diff
@@ -10,7 +10,17 @@ export function parseArgs(args: string): string[] | never {
...
-    return JSON.parse('[' + args + ']');
+    // Parse string to array.
+    let parsedArgs = JSON.parse('[' + args + ']');
+
+    // Replace boolean strings with boolean literals.
+    parsedArgs = parsedArgs.map((value) => {
+      if(value === 'false') return false;
+      else if(value === 'true') return true;
+      return value;
+    });
```
spalladino: Can we make a case insensitive comparison here?

PR #629, opened 2019-02-06
Rename ZosContract to Contract

ajsantander:
For simplicity, renames ZosContract.ts to Contract.ts, which implies that all imports in the rest of the code are now:
```
import Contract from '../artifacts/Contract';
```

Note: The only file that imports Web3's Contract is Contract.ts, and it does so using an alias to avoid collision:
```
import { Contract as Web3Contract, TransactionObject, BlockType } from 'web3-eth-contract';
```

Also note: There are now two files with a very similar name, which could be confusing: Contract.ts and Contract*s*.ts. I suggest that we rename this file as well (since all it does is retrieve contract schemas from disk) to something like:
* Artifacts
* ContractsSchema
* ArtifactsLoader
* ...

### DIFF
--- a/packages/cli/src/interface/ValidationLogger.ts
+++ b/packages/cli/src/interface/ValidationLogger.ts
@@ -7,7 +7,7 @@ import {
   BuildArtifacts,
   StorageLayoutInfo,
   Operation,
-  ZosContract
+  Contract
 } from 'zos-lib';
 import { ContractInterface } from '../models/files/ZosNetworkFile';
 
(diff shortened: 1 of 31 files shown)


spalladino:
Thanks Ale! I'm not sure we need to rename Contracts, as [there is precedent](https://stackoverflow.com/questions/35860426/what-is-the-difference-between-array-and-arrays-class-in-java) for this :-P

Now, why was the building canceled? Could we resume it?

ajsantander:
> Now, why was the building canceled? Could we resume it?

I'm having trouble again with Web3's sudependency 'websocket', which is causing the CI tests to fail. The tests were failing all around, so I cancelled the bulds.

spalladino:
Ouch, it's not a good idea to unpin web3 for the tests. How about just removing any changes to `package-lock` from the PR?

ajsantander:
> Ouch, it's not a good idea to unpin web3 for the tests. How about just removing any changes to `package-lock` from the PR?

I pinned it in the first place because of the same problem. When I did, the tests passed... so I'm wondering why they don't anymore.

ajsantander:
Tests passed with unpinned Web3. However, will be reverting the change so that this PR doesn't unpin web3 and doesn't change any NPM files.

PR #630, opened 2019-02-06
Parse ZosContract#new arguments

jcarpanelli:
Sorry, I couldn't help myself ¯\_(ツ)_/¯

### DIFF
--- a/packages/cli/test/scripts/migrate.test.js
+++ b/packages/cli/test/scripts/migrate.test.js
@@ -129,12 +129,12 @@ contract('migrate-zosversion script', function(accounts) {
       await push({ network, txParams, networkFile: this.networkFile });
       this.implV1 = this.networkFile.contract('ImplV1');
       this.implV2 = this.networkFile.contract('ImplV2');
-      this.directory = await ImplementationDirectory.new([], { from: owner });
+      this.directory = await ImplementationDirectory.new({ from: owner });
       await this.directory.methods.setImplementation('ImplV1', this.implV1.address).send({ from: owner });
       await this.directory.methods.setImplementation('ImplV2', this.implV2.address).send({ from: owner });
-      this.package = await Package.new([], { from: owner });
+      this.package = await Package.new({ from: owner });
       await this.package.methods.addVersion(toSemanticVersion(this.version), this.directory.address, this.contentURI).send({ from: owner });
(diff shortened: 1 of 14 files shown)


packages/lib/src/artifacts/ZosContract.ts
```diff
@@ -109,3 +110,13 @@ export function createZosContract(schema: any): ZosContract {
   const contract = ZWeb3.contract(schema.abi, null, Contracts.getArtifactsDefaults());
   return _wrapContractInstance(schema, contract);
 }
+
+function parseArguments(passedArguments) {
+  let givenOptions = {};
+  if (passedArguments.length > 0) {
+    const lastArg = passedArguments[passedArguments.length - 1];
+    if (typeof(lastArg) === 'object' && !Array.isArray(lastArg) && !BN.isBigNumber(lastArg)) givenOptions = passedArguments.pop();
```
spalladino: How are structs passed in? Is it possible to use an object with the same fields as the struct when calling a function? If so, this could lead to ambiguous results.

jcarpanelli: Nice catch! We can verify whether the constructor receives any arguments in the ABI, and if it does, just `pop` the given options if `passedArguments.length` is greater by one than that number of arguments, something like:
```ts
const constructorAbi = abi.find((elem) => elem.type === 'constructor') || {};
const constructorArgs = constructorAbi.inputs && constructorAbi.inputs.length > 0 ? constructorAbi.inputs : [];
let givenOptions = {};


if (passedArguments.length === constructorArgs.length + 1) {
    const lastArg = passedArguments[passedArguments.length - 1];
    if (typeof(lastArg) === 'object' && !Array.isArray(lastArg) && !BN.isBigNumber(lastArg)) {
      givenOptions = passedArguments.pop();
    }
}
```

We can go one step further and verify if the last argument in the constructor ABI is indeed a `struct`/`tuple` and play a bit with that.

I have other things to do regarding testing, but I'm very interested in how this works, and I'll probably enhance it later this week :slightly_smiling_face:.

jcarpanelli:
hey @spalladino, I implemented what I proposed last week that solves those ambiguous cases you mentioned. Anyway, it seems that `web3` `Contract#new` cannot parse structs properly, but it's not our problem to solve.

PR #634, opened 2019-02-06
Add a note to "upgrading.md" about initializing new variables.

### LINKED ISSUE #616: Add hints on running migrations in contracts on upgrades in docs
See https://github.com/zeppelinos/zos/issues/615 for more info:

> Regarding the issue, it would be nice to have this information in the docs. Something like new section New variables initialization at the end of Upgrading your project will do the trick.

ajsantander:
Fix #616 
Part of #635

### DIFF
--- a/packages/docs/docs/docs/upgrading.md
+++ b/packages/docs/docs/docs/upgrading.md
@@ -166,7 +166,7 @@ variable `t` to our contract. But how do we set the initial value of `t`?
 The variables `x` and `s` were initialized with the `initialize` function,
 which was called when the proxy was created via the `zos create MyContract --init initialize --args ...` 
 command. The `update` command also accepts `--init` and `--args` parameters, so we can use a function
-to initialize the new variable once again. We cannot to use the same `initialize` function again, because
+to initialize the new variable once again. We cannot use the same `initialize` function again, because
 the `Initializable` modifier guards it against being called more than once. We need a new function. A good name for the 
 new function could be something like `initializeVersion2` or `migrateToV2`. This function would simply
 set the initial value of `t` and should be called with `zos update MyContract --init migrateToV2 --args 99`.
--- a/packages/docs/docs/docs/upgrading.md
+++ b/packages/docs/docs/docs/upgrading.md
@@ -160,7 +160,7 @@ truffle(local)> myContract.x().then(x => x.toString())
 43
 ```
 
-Now, just as a mental exercise, let's imagine that instead of just adding a new 
+Now let's imagine that instead of just adding a new 
 function to the contract (a change to functionality), we wanted to add a new 
 variable `t` (a change to storage layout). How do we set the initial value of `t`?
 The variables `x` and `s` were initialized with the `initialize` function,
--- a/packages/docs/docs/docs/upgrading.md
+++ b/packages/docs/docs/docs/upgrading.md
@@ -160,6 +160,24 @@ truffle(local)> myContract.x().then(x => x.toString())
 43
 ```
 
+Now, just as a mental exercise, let's imagine that instead of just adding a new 
+function to the contract (a change to functionality), we wanted to add a new 
+variable `t` (a change to storage layout). How do we set the initial value of `t`?
+The variables `x` and `s` were initialized with the `initialize` function,
+which was called when the proxy was created via the `zos create MyContract --init initialize --args ...` 
+command. The `update` command also accepts `--init` and `--args` parameters, so we can use a function
+to initialize the new variable once again. We cannot to use the same `initialize` function again, because
+the `Initializable` modifier guards it against being called more than once. We need a new function. A good name for the 
+new function could be something like `initializeVersion2` or `migrateToV2`. This function would simply
+set the initial value of `t` and should be called with `zos update MyContract --init migrateToV2 --args 99`.
+
+```
+function migrateToV2(uint256 _t) public {
+  require(_t == 0);
+  t = _t;
+}
+```
+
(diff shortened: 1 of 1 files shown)


packages/docs/docs/docs/upgrading.md
```diff
@@ -160,6 +160,24 @@ truffle(local)> myContract.x().then(x => x.toString())
 43
 ```
 
+Now, just as a mental exercise, let's imagine that instead of just adding a new
```
spalladino:
```suggestion
Now let's imagine that instead of just adding a new 
```

packages/docs/docs/docs/upgrading.md
```diff
@@ -160,6 +160,24 @@ truffle(local)> myContract.x().then(x => x.toString())
 43
 ```
 
+Now, just as a mental exercise, let's imagine that instead of just adding a new 
+function to the contract (a change to functionality), we wanted to add a new 
+variable `t` (a change to storage layout). How do we set the initial value of `t`?
```
spalladino:
```suggestion
variable `t` to our contract. But how do we set the initial value of `t`?
```

packages/docs/docs/docs/upgrading.md
```diff
@@ -160,6 +160,24 @@ truffle(local)> myContract.x().then(x => x.toString())
...
+The variables `x` and `s` were initialized with the `initialize` function,
+which was called when the proxy was created via the `zos create MyContract --init initialize --args ...` 
+command. The `update` command also accepts `--init` and `--args` parameters, so we can use a function
```
spalladino: I'd first explain that, even if we should add `t` to the initialize function (so we can initialize it on new instances we create), we cannot call that function again on contracts that have already been created. Then go on to explain that we need a new function, and only then mention that `zos update` also accepts the init option.

packages/docs/docs/docs/upgrading.md
```diff
@@ -160,6 +160,24 @@ truffle(local)> myContract.x().then(x => x.toString())
...
+to initialize the new variable once again. We cannot to use the same `initialize` function again, because
```
spalladino:
```suggestion
to initialize the new variable once again. We cannot use the same `initialize` function again, because
```

packages/docs/docs/docs/upgrading.md
```diff
@@ -160,6 +160,24 @@ truffle(local)> myContract.x().then(x => x.toString())
...
+the `Initializable` modifier guards it against being called more than once. We need a new function. A good name for the 
+new function could be something like `initializeVersion2` or `migrateToV2`. This function would simply
```
spalladino: Or maybe some name related to initializing T?

packages/docs/docs/docs/upgrading.md
```diff
@@ -160,6 +160,24 @@ truffle(local)> myContract.x().then(x => x.toString())
...
+set the initial value of `t` and should be called with `zos update MyContract --init migrateToV2 --args 99`.
```
spalladino: "set the initial value" and check that it has not been already initialized!

packages/docs/docs/docs/upgrading.md
```diff
@@ -160,6 +160,24 @@ truffle(local)> myContract.x().then(x => x.toString())
...
+
+```
+function migrateToV2(uint256 _t) public {
+  require(_t == 0);
```
spalladino: Mention that this trick only works if t cannot be zero in this contract, otherwise it could be reinitialized

packages/docs/docs/docs/upgrading.md
```diff
@@ -160,6 +160,24 @@ truffle(local)> myContract.x().then(x => x.toString())
...
+The variables `x` and `s` were initialized with the `initialize` function,
+which was called when the proxy was created via the `zos create MyContract --init initialize --args ...` 
+command. The `update` command also accepts `--init` and `--args` parameters, so we can use a function
+to initialize the new variable once again. We cannot to use the same `initialize` function again, because
+the `Initializable` modifier guards it against being called more than once. We need a new function. A good name for the 
+new function could be something like `initializeVersion2` or `migrateToV2`. This function would simply
+set the initial value of `t` and should be called with `zos update MyContract --init migrateToV2 --args 99`.
+
+```
+function migrateToV2(uint256 _t) public {
```
spalladino: I'd include the full contract code, either here or after this function, to stress the point that you should not just add the new migrate function, but also modify the initialize one.

ajsantander:
@spalladino thx for the review! All suggestions have been applied.

packages/docs/docs/docs/upgrading.md
```diff
@@ -160,6 +160,58 @@ truffle(local)> myContract.x().then(x => x.toString())
...
+
+import "zos-lib/contracts/Initializable.sol";
+
+contract MyContract is Initializable {
+
+  uint256 public x;
+  string public s;
+  uint256 public t;
+
+  function initialize(uint256 _x, string memory _s) initializer public {
```
spalladino: The original initialize should also be updated to include `t`, so any new instances can be fully initialized; otherwise any new deploy would require calling initialize + initializeT. 

I'd suggest also to make a note about this: perhaps earlier when you say "At first, you may think that the solution would be to add the initialization of `t` to the end of the initialize function, but that wouldn't work", we could clarify that it won't work for updating, but it's still needed for new instances.

spalladino:
@ajsantander do you want to add this under [guides and tutorials](https://forum.zeppelin.solutions/c/support/guides-and-tutorials) in the forum later? Maybe adding a little intro on the motivation. I think it'd be a nice addition!

PR #654, opened 2019-02-08
Checksum `from` address to avoid addresses comparison fail 

jcarpanelli:
Fixes https://github.com/zeppelinos/zos/issues/648.

### DIFF
--- a/packages/lib/src/artifacts/ZWeb3.ts
+++ b/packages/lib/src/artifacts/ZWeb3.ts
@@ -1,8 +1,12 @@
+import Logger from '../utils/Logger';
 import sleep from '../helpers/sleep';
 import Web3 from 'web3';
 import { TransactionReceipt } from 'web3/types';
 import { Eth, Block, Transaction } from 'web3-eth';
 import { Contract } from 'web3-eth-contract';
+import { toChecksumAddress } from 'web3-utils';
+
+const log: Logger = new Logger('ZWeb3');
 
 // Reference: see https://github.com/ethereum/EIPs/blob/master/EIPS/eip-155.md#list-of-chain-ids
 const NETWORKS = {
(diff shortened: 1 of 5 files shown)


packages/lib/src/artifacts/ZWeb3.ts
```diff
@@ -61,6 +65,14 @@ export default class ZWeb3 {
     return (await ZWeb3.accounts())[0];
   }
 
+  public static toChecksumAddress(address: string): string {
```
spalladino: Add a throw if checksum is incorrect. Logic should then be:

- If address seems checksummed (ie has at least an uppercase letter), validate it
  - If it is incorrect, throw
  - If it is correct, go
- If address has no uppercase letters, show a warning, and checksum it ourselves

jcarpanelli:
@spalladino, already implemented your suggestions. I also removed all `.toLowerCase()` I found, as `truffle-hdwallet-provider` does not complain if you provide a non-lowercased address while using private keys instead of a mnemonic.
Also, I added some tests :slightly_smiling_face:.

packages/lib/test/src/artifacts/ZWeb3.test.js
```diff
@@ -189,6 +189,34 @@ contract('ZWeb3', accounts => {
...
+      describe('checksum address', function () {
+        context('when address has uppercase hexa letters', function () {
+          context('when address is not well checksummed', function () {
+            it('fails', function () {
+              try {
+                const address = '0x82154A1cbcAb0461D4C12ee9Cb890244eead6d9F';
+                const checksummed = ZWeb3.toChecksumAddress(address);
+              } catch(error) {
+                error.message.should.match(/is not a valid Ethereum address or it has not been checksummed correctly/);
+              }
```
spalladino: Note that this test does not actually test that it fails! If the call does not throw, the catch is never executed, and the test succeeds. Look for the `expects(fn).to.throw()` method [in chai](https://www.chaijs.com/api/bdd/#method_throw) to implement this.

jcarpanelli: Fixed!

PR #669, opened 2019-02-12
Published docs for version 2.1

ajsantander:
This PR consists of the output of running `npm run bump-docs v2.1.0` with the latest code, which produces changes within the docs package, mainly docs/docs/website/build. This output was commited into the gh-pages branch of zos-docs as well, so that the changes are in production.

Also note that a small modification to the bump-docs script was made, so that the local static markdown files are used, instead of those of the target release tag.


spalladino:
Great job @ajsantander! One question though: aren't there autogenerated files missing from the build? I'm checking the [verify subcommand reference](https://docs.zeppelinos.org/docs/cli_verify.html) for instance, and it includes no mention of the `api-key` parameter for instance.

> Also note that a small modification to the bump-docs script was made, so that the local static markdown files are used, instead of those of the target release tag.
I'm not sure what this means, but maybe I'm missing some context. If you checked this with @jcarpanelli and you are both ok with it, fine for me!

jcarpanelli:
IMO we should (not now) keep both the previous functionality of bumping using an already released version on github, and bumping using the local files.
I would leave it as it is for now. Already created [an issue](https://github.com/zeppelinos/zos/issues/670) to solve this later.

ajsantander:
We will hold this PR for now because:
1) There are some improvements being done in the publish scripts,
2) The 2.1 docs still have some minor improvements to be made.

jcarpanelli:
@ajsantander, I'm closing this PR without merging it as we've already bumped the docs to `v2.1.0` [here](https://github.com/zeppelinos/zos/pull/694).

Issue #635, opened 2019-02-06
Publish docs for versions 2.1 and 2.2


ajsantander:
The latest version in docs is 2.0.0 and we have a few releases to the sources for versions 2.1 and 2.2. This issue is simply for tracking all the work related to this upgrade.

#### Make sure all guides work with zOS 2.1
- [x] Issue #639 "The quickstart guide from the documentation doesn't work with zOS v2.1.x" - Fixed by PR #643
- [x] Issue #656 "Programmatic usage guide needs to be updated for zOS 2.1" - Fixed by #658
- [x] Issue #655 "ERC20 Migration guide needs to be updated for zOS 2.1" - Fixed by #659

#### Document changes and new features
- [x] Delete or change notes like "The current version of ZeppelinOS uses Solidity 0.4.24..." - Fixed by #662
- [x] Document ability to use different Solidity compiler versions - Fixed by #662
- [x] Document contract verification on Etherscan - Fixed by PR #662
- [x] Re-generate dynamic docs
- [x] #566 Add note about upgrades among different compiler versions - Fixed by PR #662

#### Publish 2.1
- [x] Publish docs for version 2.1

#### Make sure all guides work with zOS 2.2
- [x] Migrate guides to use zOS 2.2 with Solidity 0.5, Truffle 5, Web3 1.0, async/await in truffle console, etc. - Fixed by #677

#### Document changes and new features
- [x] Remove mentions to transparent proxy. - Fixed by #686
- [x] Update section about transparent proxy. - Fixed by #686
- [x] Update contracts architecture section. - Fixed by #686
- [x] Update configuration files section. - Fixed by #686
- [x] Remove link to vouching contracts in vouching.md - Fixed by #690
- [x] #683 Add "Cannot call fallback function from the proxy admin" error to FAQ - Fixed by #690
- [x] #684 Add instructions on migration from 2.1 to 2.2 in documentation - Solved by #689

#### Publish 2.2
- [x] Publish docs for version 2.2

### Nice to haves
- [ ] #628 Link in documentation pointing to source are broken
- [x] #616 "Add hints on running migrations in contracts on upgrades in docs" - Fixed by PR #634
- [ ] #350 Consider renaming stake variable/methods to use 'vouch' instead
- [ ] #252 Doesn't work on windows 10
- [ ] #599 Outdated reference to BaseApp.sol in documentation
- [ ] #624 Deployment guide is unclear since it does not include creating a contract
[...]

PR #677, opened 2019-02-12
Updates the documentation guides for the zos 2.2 release

ajsantander:
Part of #635 

This PR simply reviews all guides in the documentation to make sure that they can be followed correctly with ZeppelinOS v2.2.0, which involved changes like:
- Changing all contracts embedded in the documentation to `pragma solidity ^0.5.0` and making minor modifications in their code.
- No longer requiring to install the specific version 4.x of Truffle.
- No longer requiring to install the specific version 2.0.2 of openzeppelin-eth.
- Minor changes to usages of truffle console, taking advantage of Truffle 5's support for await syntax.
- Minor modifications to embedded scripts and truffle console runs to use Web3 1.0.


packages/docs/docs/docs/zos_lib.md
```diff
@@ -139,20 +135,26 @@ And now, let's write our upgrading script in `index.js`:
...
+
 // Load the contract.
 const MyContractV0 = Contracts.getFromLocal('MyContractV0');
 const MyContractV1 = Contracts.getFromLocal('MyContractV1');
 
 async function main() {
+
   // Instantiate a project.
-  const project = new SimpleProject('MyProject', { from: web3.eth.accounts[0] });
+  const initializerAddress = (await web3.eth.getAccounts())[1];
```
spalladino: Just curious: why the change from account 0 to 1 here?

ajsantander: With 0 I was getting this error:
```
VM Exception while processing transaction: revert Cannot call fallback function from the proxy admin
```

ajsantander: @spalladino 

Re:
> One question: what do you think about keeping the version identifiers for dependencies, to avoid any problems in the future? So, instead of suggesting to install plainly `npm i truffle`, suggest `npm i truffle@5.0`; same for `zos-lib@2.2` (note that it does not include the patch version, so it installs the latest patch), etc.

I thought about it and my conclusion was that documentation for @next versions could not have to specify the dependency versions. This simplifies the guides for the reader, but asumes that all the latest versions will just work. We could not make that assumption and just fix the dependency versions. Which one do you like best?

spalladino: > With 0 I was getting this error:

My bad. That example is using SimpleProject, and not ProxyAdminProject, so it makes sense that it requires a different one!

> I thought about it and my conclusion was that documentation for @next versions could not have to specify the dependency versions. This simplifies the guides for the reader, but asumes that all the latest versions will just work.

Let's specify versions for all dependencies (such as truffle), so if they release a new breaking version, that does not break our guides. As for zos and zos-lib, agree on leaving them without version identifiers for simplicity.

PR #679, opened 2019-02-13
Force linter on pre-push instead of pre-commit

### LINKED ISSUE #661: Fix linter hook
We should run the linter as a pre-push instead of pre-commit. 
Please make sure to use locally installed lerna dependency instead of globally.

facuspagnuolo:
Fixes #661

### DIFF
--- a/.githooks/pre-push
+++ b/.githooks/pre-push
@@ -1,0 +1,2 @@
+#!/bin/sh
+npx lerna run lint
--- a/.githooks/pre-commit
+++ b/.githooks/pre-commit
@@ -1,2 +1,0 @@
-#!/bin/sh
-lerna run lint


.githooks/pre-push
```diff
@@ -0,0 +1,2 @@
+#!/bin/sh
+npx lerna run lint
```
jcarpanelli: This will just throw an error if any linter rule is not being followed. Neither will fix them nor will commit the changes.
We may want to:
- Run the linter
- `git add` the files modified by the linter and commit them with a generic commit message such as `Run lintfix`
- continue with the `git push`.

WDYT?

jcarpanelli: @facuspagnuolo may I push some changes to your branch so we could close this PR? :hugs:

jcarpanelli:
@facuspagnuolo may I push some changes to your branch so we could close this PR? :hugs:

jcarpanelli:
I'm closing this PR as we concluded that running the linter before committing instead of pushing is ok for now. The other issue that this PR solves (running the linter using lerna locally instead of globally) has been fixed in https://github.com/zeppelinos/zos/pull/794.

PR #687, opened 2019-02-13
Prompt migration question for each network independently

jcarpanelli:
The `zosversion` check has been changed in order to check it in the `zos.network.json` file rather than in the `zos.json` file. This means that we are going to prompt the migration question for each network independently instead of checking it once and then migrate without asking on the remaining networks.

### DIFF
--- a/packages/cli/src/models/files/ZosNetworkFile.ts
+++ b/packages/cli/src/models/files/ZosNetworkFile.ts
@@ -72,8 +72,8 @@ export default class ZosNetworkFile {
     dependencies: { [dependencyName: string]: DependencyInterface };
   };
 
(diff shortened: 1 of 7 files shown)


packages/cli/src/models/files/ZosNetworkFile.ts
```diff
@@ -72,6 +72,11 @@ export default class ZosNetworkFile {
     dependencies: { [dependencyName: string]: DependencyInterface };
   };
 
+  public static getZosversion(fileName: string): string | null {
```
facuspagnuolo: can we use simply `network` here instead

jcarpanelli: okabz

Issue #696, opened 2019-02-15
Improvements on docs


ajsantander:
This issue comprises the list of nice to haves from #635

- [x] #628 Link in documentation pointing to source are broken - Fixed by https://github.com/OpenZeppelin/solidity-docgen/pull/19
- [x] #616 "Add hints on running migrations in contracts on upgrades in docs" - Fixed by PR #634
- [x] #350 Consider renaming stake variable/methods to use 'vouch' instead - n/a, not relevant to docs anymore - moved to the vouching repo https://github.com/zeppelinos/zos-vouching/issues/9
- [x] #599 Outdated reference to BaseApp.sol in documentation - Fixed #697
- [x] #624 Deployment guide is unclear since it does not include creating a contract - Fixed by #698 
- [x] #372 Quickstart on the top bar should link to deploying - CLOSED, see comments on issue
- [x] "spacial" as in spacex in writing_contracts.md - Fixed by PR #660 
- [x] #339 Add guide to build an upgradeable ERC20 - Done already with zos_lib.md in docs?
- [x] #460 Create example project for creating upgradeable instances from a contract - Fixed by #724
- [x] #726 Update section on calling initialize manually - Fixed by #728

jcarpanelli:
Hey @ajsantander, just to be clear, remember that as we've already generated both `v2.1.0` and `v2.2.0` docs, you'll have to edit the already generated `.md`s in the folders `docs/docs/website/versioned_docs/version-2.1.0` and `docs/website/versioned_docs/version-2.2.0`. Also, you will have to update the `.md`s inside the `docs/docs/docs/` folder for `v2.2.0`.

ajsantander:
@jcarpanelli yes I understand. 

We may choose to just leave these improvements for the next update instead. @spalladino what do you think?

spalladino:
@ajsantander @jcarpanelli your call! If it's easy to incorporate the improvements on this version, then I'd say go for it.

PR #697, opened 2019-02-15
Replace reference in documentation to BaseApp with App

### LINKED ISSUE #599: Outdated reference to BaseApp.sol in documentation
In [the docs](https://docs.zeppelinos.org/docs/writing_contracts.html#creating-new-instances-from-your-contract-code), there is a reference to BaseApp contract, which no longer exists. Replace it by App.

ajsantander:
Fix #599 
Part of #696

### DIFF
--- a/packages/docs/docs/docs/writing_contracts.md
+++ b/packages/docs/docs/docs/writing_contracts.md
@@ -168,16 +168,16 @@ $ TOKEN=$(zos create TokenContract)
 $ zos create MyContract --init --args $TOKEN
 ```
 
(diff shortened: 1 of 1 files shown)


spalladino:
Looks good! One question though: should it be merged into `master` or `docs/2.2`, as you had pointed out?

ajsantander:
@spalladino this should be merged to master.
Reason: docs/2.2 only existed while we were in the situation of releasing two versions of the documentation at the same time (2.1 and 2.2), which is no longer the case.

spalladino:
Ack, thanks!!

PR #698, opened 2019-02-15
Commented about not interacting with logic contracts in the deploying guide

### LINKED ISSUE #624: Deployment guide is unclear since it does not include creating a contract
The [Deploying your first project](https://docs.zeppelinos.org/docs/deploying.html) guide ends with the project being `push`ed to the network, but with no `create` operations. This is confusing to the reader, who runs a push and sees a log that their contract has been deployed, so thinks that *they have to use the logic contract*, as no `create` operations were mentioned in that article.

Consider moving part of the [following guide](https://docs.zeppelinos.org/docs/upgrading.html) so at least a `zos create` is included on the _Deploying..._ one, and the user understands that they have to use that contract and not the logic one.

(feedback from @rrecuero)

ajsantander:
Fix #624 
Part of #696

### DIFF
--- a/packages/docs/docs/docs/deploying.md
+++ b/packages/docs/docs/docs/deploying.md
@@ -180,7 +180,7 @@ upgradeable instances, as we will see later in the
 
 For example, you wouldn't interact with the instance of MyContract that we just deployed. This instance is only intended for defining the behavior or logic of the contract. You can find the address of this logic contract in the `zos.dev-<network_id>.json` file, under `contracts.MyContract.address`.
 
-So what _do_ you interact with? As we'll see in depth in the next guide, what you do is create a proxy for the logic contract, and interact with this proxy contract instead. 
+So what _do_ you interact with? As we'll see in depth in the next guide, what you do is create an upgradeable instance for the logic contract, and interact with this instance instead. 
 
 ```console
 npx zos create MyContract --init initialize --args 42,hitchhiker
--- a/packages/docs/docs/docs/deploying.md
+++ b/packages/docs/docs/docs/deploying.md
@@ -186,7 +186,7 @@ So what _do_ you interact with? As we'll see in depth in the next guide, what yo
 npx zos create MyContract --init initialize --args 42,hitchhiker
 ```
 
-Once you create a proxy, you always interact with the same proxy, even if you decide to upgrade it and define a new logic contract for it. Executing the command above will return the address of the proxy, but you can also find it in the `zos.dev-<network_id>.json` file, under `proxies.my-project.MyContract`. You will always interact with this address, and you should never interact with the address of the logic contract.
+Once you create an upgradeable instance of MyContract, you always interact with the same instance, even if you decide to upgrade it and define a new logic contract for it. Executing the command above will return the address of the upgradeable contract instance, but you can also find it in the `zos.dev-<network_id>.json` file, under `proxies.my-project.MyContract`. You will always interact with this address, and you should never interact with the address of the logic contract.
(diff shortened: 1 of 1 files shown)


packages/docs/docs/docs/deploying.md
```diff
@@ -178,6 +178,16 @@ to understand is that the contracts deployed by the `push` command are logic con
 upgradeable instances, as we will see later in the
 [Upgrading your project](https://docs.zeppelinos.org/docs/upgrading.html) section.
 
+For example, you wouldn't interact with the instance of MyContract that we just deployed. This instance is only intended for defining the behavior or logic of the contract. You can find the address of this logic contract in the `zos.dev-<network_id>.json` file, under `contracts.MyContract.address`. Remember to never interact with this address!
```
```suggestion
For example, you wouldn't interact with the instance of MyContract that we just deployed. This instance is only intended for defining the behavior or logic of the contract. You can find the address of this logic contract in the `zos.dev-<network_id>.json` file, under `contracts.MyContract.address`.
```

I think the point is already clear :-)

spalladino: Also, I'd avoid referring to the logic contract as a "MyContract instance", since it could be confused with the "upgradeable instance". I'd just say "would not interact with the MyContract that we just deployed", etc.

packages/docs/docs/docs/deploying.md
```diff
@@ -178,6 +178,16 @@ to understand is that the contracts deployed by the `push` command are logic con
...
+
+So what _do_ you interact with? As we'll see in depth in the next guide, what you do is create a proxy for the logic contract, and interact with this proxy contract instead.
```
spalladino: Note that in the following guide we hardly ever use the term "proxy". I'd refer to them as "upgradeable instances" for consistency.

ajsantander:
Thanks @spalladino !
Suggestions implemented.

packages/docs/docs/docs/deploying.md
spalladino: > I'd avoid referring to the logic contract as a "MyContract instance", since it could be confused with the "upgradeable instance". I'd just say "would not interact with the MyContract that we just deployed", etc.

I think this part got hidden when accepting the suggestion!

packages/docs/docs/docs/deploying.md
```diff
@@ -178,6 +178,16 @@ to understand is that the contracts deployed by the `push` command are logic con
...
 
+For example, you wouldn't interact with the instance of MyContract that we just deployed. This instance is only intended for defining the behavior or logic of the contract. You can find the address of this logic contract in the `zos.dev-<network_id>.json` file, under `contracts.MyContract.address`.
+
+So what _do_ you interact with? As we'll see in depth in the next guide, what you do is create an upgradeable instance for the logic contract, and interact with this instance instead. 
+
+```console
+npx zos create MyContract --init initialize --args 42,hitchhiker
+```
+
+Once you create a proxy, you always interact with the same proxy, even if you decide to upgrade it and define a new logic contract for it. Executing the command above will return the address of the proxy, but you can also find it in the `zos.dev-<network_id>.json` file, under `proxies.my-project.MyContract`. You will always interact with this address, and you should never interact with the address of the logic contract.
```
spalladino: Note that in this paragraph you are talking about proxies again

ajsantander: Right! Sorry, I'm too used to the term proxy, but I think I got it now :laughing: 
Removed all references to proxy, and will use "upgradeable instance" from now on.

PR #725, opened 2019-02-21
Interactive init

jcarpanelli:
Fixes https://github.com/zeppelinos/zos/issues/713.
This PR makes the `init` command interactive, by asking both `project-name` and `version` parameters if not provided, and hinting the `name` and `version` values from the `package.json` file if possible:

interactive-init

### DIFF
--- a/packages/cli/package-lock.json
+++ b/packages/cli/package-lock.json
@@ -127,6 +127,11 @@
 			"resolved": "https://registry.npmjs.org/aes-js/-/aes-js-3.0.0.tgz",
 			"integrity": "sha1-4h3xCtbCBTKVvLuNq0Cwnb6ofk0="
 		},
+		"ansi-escapes": {
+			"version": "3.2.0",
+			"resolved": "https://registry.npmjs.org/ansi-escapes/-/ansi-escapes-3.2.0.tgz",
+			"integrity": "sha512-cBhpre4ma+U0T1oM5fXg7Dy1Jw7zzwv7lt/GoCpr+hDQJoYnKVPLL4dCvSEFMmQurOQvSrwT7SL/DAlhBI97RQ=="
+		},
 		"ansi-regex": {
 			"version": "2.1.1",
 			"resolved": "https://registry.npmjs.org/ansi-regex/-/ansi-regex-2.1.1.tgz",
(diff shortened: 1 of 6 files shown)


packages/cli/package-lock.json
```diff
@@ -4454,11 +4603,27 @@
...
+			"requires": {
+				"is-promise": "^2.1.0"
+			}
+		},
 		"rustbn.js": {
 			"version": "0.1.2",
 			"resolved": "https://registry.npmjs.org/rustbn.js/-/rustbn.js-0.1.2.tgz",
 			"integrity": "sha512-bAkNqSHYdJdFsBC7Z11JgzYktL31HIpB2o70jZcGiL1U1TVtPyvaVhDrGWwS8uZtaqwW2k6NOPGZCqW/Dgh5Lg=="
 		},
+		"rxjs": {
```
spalladino: Damn, `rxjs` is heavy (17mb). Doesn't matter though, inquirer is beautiful and is totally worth it :-)

packages/cli/src/commands/init.ts
```diff
@@ -1,8 +1,10 @@
 import push from './push';
 import init from '../scripts/init';
+import { promptForArgumentsIfNeeded } from '../utils/prompt';
+import { FileSystem } from 'zos-lib';
 
 const name: string = 'init';
-const signature: string = `${name} <project-name> [version]`;
+const signature: string = `${name} [project-name] [version]`;
 const description: string = `initialize your ZeppelinOS project. Provide a <project-name> and optionally an initial [version] name`;
```
spalladino: We should review these descriptions, especially when it does not add any useful information. "Provide a project-name" is already implied by its arguments. We can make a pass at it in a future PR.

packages/lib/src/utils/FileSystem.ts
```diff
@@ -45,13 +45,8 @@ export function parseJson(filename: string): any {
...
 
-export function parseJsonIfExists(filename: string): any | null {
-  if (exists(filename)) {
-    return JSON.parse(read(filename));
-  }
-  else {
-    return null;
-  }
+export function parseJsonIfExists(filename: string): {} | null {
+  return exists(filename) ? JSON.parse(read(filename)) : null;
```
spalladino: THANK YOU

packages/cli/src/commands/init.ts
jcarpanelli: Addressed in https://github.com/zeppelinos/zos/issues/727

jcarpanelli:
> Looks great! There is one thing missing we need to take into account (either on this PR or another), and that is running `zos` in a non-interactive script. If the user is running `zos` in a CI (for instance), then we cannot pause and prompt - we need to throw an exception stating that they missed a required parameter.
> 
> Note that this requires detecting whether the user is in an interactive prompt, which may involve different methods for linux, mac, or windows; even worse, different terminals (bash, zsh, fish) may have different ways for detecting this (see [this SO question](https://unix.stackexchange.com/questions/26676/how-to-check-if-a-shell-is-login-interactive-batch)). My suggestion would be to start out with a simple method that works on bash, and errs on the side of thinking that the environment is interactive, just making sure that it properly detects a non-interactive debian (which is what most CIs run).

Good catch. Already adressed in https://github.com/zeppelinos/zos/issues/729.

PR #728, opened 2019-02-21
Added a note on calling overloaded functions with Truffle 4 and 5.

### LINKED ISSUE #726: Update section on calling initialize manually
In ["Calling Initialize Functions Manually in Your Unit Tests"](https://docs.zeppelinos.org/docs/advanced.html#calling-initialize-functions-manually-in-your-unit-tests) we are suggesting to use `zos-lib`'s `encodeCall` to manually craft the transaction and send it to the contract. 

However, since truffle 5 (which is the default truffle we suggest since v2.2), this suggested way breaks since truffle arbitrarily clears the `data` field when using `sendTransaction` (see [here](https://forum.zeppelin.solutions/t/revert-when-calling-initialize-manually/314/6) for more info). 

As suggested by @nventuro on the thread, we should suggest using truffle's new way to [call overloaded methods](https://github.com/trufflesuite/truffle/releases/tag/v5.0.0#user-content-what-s-new-in-truffle-v5-interacting-with-your-contracts-overloaded-solidity-functions), without requiring `encodeCall` at all.

ajsantander:
Fix #726

### DIFF
--- a/packages/docs/docs/docs/testing.md
+++ b/packages/docs/docs/docs/testing.md
@@ -81,9 +81,12 @@ contract('Sample', function ([_, owner]) {
 
 Use this convenient tool to write tests for your code and storage migrations before executing them in production.
 
-## Calling Initialize Functions Manually in Your Unit Tests
+## Calling Initialize Functions Manually in Your Unit Tests with Truffle 4
 
-Truffle does not know how to resolve situations where a contract 
+_Note: If you're using Truffle 5, as is recommended with ZeppelinOS 2.2, please see 
+[Calling Initialize Functions Manually in Your Unit Tests with Truffle 5](https://docs.zeppelinos.org/docs/advanced.html#calling-initialize-functions-manually-in-your-unit-tests-with-truffle-5) instead._
+
+Truffle 4 does not know how to resolve situations where a contract 
 has functions that have matching names, but different arities. 
 Here's an example of a `TimedCrowdsale` contract that inherits 
 from `Crowdsale` which results in a contract that has two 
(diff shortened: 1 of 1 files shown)


packages/docs/docs/docs/testing.md
```diff
@@ -81,9 +81,12 @@ contract('Sample', function ([_, owner]) {
 
 Use this convenient tool to write tests for your code and storage migrations before executing them in production.
 
-## Calling Initialize Functions Manually in Your Unit Tests
+## Calling Initialize Functions Manually in Your Unit Tests with Truffle 4
```
spalladino: @ajsantander are we maintaining duplicate sections for truffle 4 and 5 in the guides? If not, let's remove all references to truffle 4 here. If yes, then let's discuss how do we want to support them, given the extra effort.

packages/docs/docs/docs/testing.md
```diff
@@ -128,3 +131,24 @@ data = encodeCall(
 const timeCrowdsale = await TimeCrowdsale.new()
 await timeCrowdsale.sendTransaction({ data, from: owner })
 ```
+
+## Calling Initialize Functions Manually in Your Unit Tests with Truffle 5
+
+_Note: If you're using Truffle 4, please see 
+[Calling Initialize Functions Manually in Your Unit Tests with Truffle 4](https://docs.zeppelinos.org/docs/advanced.html#calling-initialize-functions-manually-in-your-unit-tests-with-truffle-4) instead._
+
+As of version 5, Truffle uses Web3 version 1, whose Contract object has the ability to
```
spalladino: I wouldn't bother mentioning web3 1.0 here, we just want to tell the user how to call `initialize`

ajsantander:
Thanks @spalladino .
Applied your suggestions.

PR #730, opened 2019-02-22
Interactive add

jcarpanelli:
Fixes https://github.com/zeppelinos/zos/issues/714.


packages/cli/src/models/initializer/truffle/Truffle.ts
```diff
@@ -52,6 +52,21 @@ const Truffle = {
     }
   },
 
+  getContractNames(): string[] {
+    const buildDir = this.getBuildDir();
+    if (buildDir) {
+      return FileSystem.readDir(buildDir)
+        .filter((name) => {
+          const contract = FileSystem.parseJson(`${buildDir}/${name}`);
```
spalladino: We should attempt to open only json files, in case there are other artifacts in there

packages/cli/src/models/initializer/truffle/Truffle.ts
```diff
@@ -52,6 +52,21 @@ const Truffle = {
...
+          const projectDir = buildDir.replace('build/contracts', '');
+          return contract.bytecode.length > 2 && contract.sourcePath.match(projectDir, 'g') !== null;
```
spalladino: Note that `match` will turn `projectDir` into a regex, which may not be what we want here - maybe an `indexOf` is a better option.

packages/cli/src/commands/add.ts
```diff
@@ -21,13 +32,20 @@ async function action(contractNames: string[], options: any): Promise<void> {
...
   else {
-    const contractsData = contractNames.map((rawData) => {
-      const [ aName, alias ] = rawData.split(':');
-      return { name: aName, alias: (alias || aName) };
-    });
+    const contractList = Truffle.getContractNames();
+    const promptedArgs = await promptForArgumentsIfNeeded({ args: { contractNames }, props: argsProps(contractList) });
+    const contractsData = contractNames.length !== 0
+      ? contractNames.map(splitContractName)
+      : promptedArgs.contractNames.map(splitContractName);
```
spalladino: Is it necessary to run the result of `promptedArgs` via `split`?

jcarpanelli: Nope, it's not. I  just wanted to maintain consistency with what we had before. As a matter of fact, I noticed that we're also double-checking whether the `alias` property is defined or not (well, not strictly verifying this but checking the `contractNames` type) inside `scripts/add.ts`. I'll probably review why we are doing this (maybe it's just a hotfix to make some tests work?) and remove it if possible.

packages/cli/src/scripts/add.ts
```diff
@@ -4,7 +4,6 @@ import { AddParams } from './interfaces';
 
 export default function add({ contractsData, packageFile }: AddParams): void | never {
   if (contractsData.length === 0) throw new Error('At least one contract name must be provided to add.');
-  contractsData = contractsData.map((data) => typeof(data) === 'string' ? { name: data, alias: data } : data);
```
spalladino: I wouldn't remove this line, as it makes using the `script` from the programmable interface much easier

packages/lib/src/utils/FileSystem.ts
```diff
@@ -42,7 +42,11 @@ export function ifNotExistsThrow(filename: string, message: string): void {
 
 // TS-TODO: Returned object could be of a more specific type
 export function parseJson(filename: string): any {
-  return JSON.parse(read(filename));
+  try {
+    return JSON.parse(read(filename));
+  } catch(error) {
+    return null;
+  }
```
spalladino: I prefer to fail loudly if we are trying to parse a file that is not a valid JSON, or does not exist. It also helps to more descriptive error messages (such as "file not exists" or "invalid JSON at line N", instead of "cannot read property X of null"). I understand that this helps on the `getContractNames` method, but I'd rather check for json files there.

packages/cli/src/models/initializer/truffle/Truffle.ts
```diff
@@ -52,6 +52,23 @@ const Truffle = {
     }
   },
 
+  getContractNames(): string[] {
+    const buildDir = this.getBuildDir();
+    if (buildDir) {
+      return FileSystem.readDir(buildDir)
+        .filter((name) => {
```
spalladino: We can manually filter for json extensions, or use [glob](https://www.npmjs.com/package/glob) to do so

packages/cli/src/commands/share.ts
```diff
@@ -0,0 +1,12 @@
+import Truffle from '../models/initializer/truffle/Truffle';
+
+export function getContractsList(message, type) {
```
spalladino: Agree on adding these functions to prompt.ts, to avoid having a generic share.ts file, especially since it is related.

packages/cli/test/scripts/add.test.js
```diff
@@ -41,6 +41,14 @@ contract('add script', function() {
     this.packageFile.contract(contractAlias).should.eq(customContractName)
   });
 
+  it('should allow contractsData to be an array of contract names instead of an object', function() {
```
spalladino: Good catch adding a test for this

PR #743, opened 2019-02-26
Handle spaces in project path

### LINKED ISSUE #708: zos add command fails if your project has space in its path on Mac OS
If your project is inside folder such as '/folder with space', then running 'zos add' command would fail due to npx running path unquoted.

```
Bytezilla:folder with space iYalovoy$ zos add LinkedList
Compiling contracts with Truffle...
/bin/sh: /folder: No such file or directory

Command failed: /folder with space/node_modules/.bin/truffle compile --all
/bin/sh: /folder: No such file or directory
```

Not a zos issue per se, but can cause confusion.
The workaround is to remove/uninstall truffle locally and install it globally, so npx would run it from global installation with no issues.

spalladino:
Use execFile to provide path to file and options separately, instead of using plain exec with a manually built command string.

Fixes #708

### DIFF
--- a/packages/cli/src/models/compiler/Compiler.ts
+++ b/packages/cli/src/models/compiler/Compiler.ts
@@ -1,4 +1,4 @@
-import { exec } from 'child_process';
+import { execFile, ExecException } from 'child_process';
 import { FileSystem, Contracts, Logger } from 'zos-lib';
 
 const log = new Logger('Compiler');
@@ -10,7 +10,7 @@ const Compiler = {
     if (!FileSystem.exists(truffleBin)) truffleBin = 'truffle'; // Attempt to load global truffle if local was not found
 
     return new Promise((resolve, reject) => {
-      exec(`${truffleBin} compile --all`, (error, stdout, stderr) => {
+      execFile(truffleBin, ['compile', '--all'], (error: ExecException, stdout, stderr) => {
         if (error) {
           if (error.code === 127) console.error('Could not find truffle executable. Please install it by running: npm install truffle');
           reject(error);


packages/cli/src/models/compiler/Compiler.ts
```diff
@@ -10,7 +10,7 @@ const Compiler = {
     if (!FileSystem.exists(truffleBin)) truffleBin = 'truffle'; // Attempt to load global truffle if local was not found
 
     return new Promise((resolve, reject) => {
-      exec(`${truffleBin} compile --all`, (error, stdout, stderr) => {
+      execFile(truffleBin, ['compile', '--all'], (error: ExecException, stdout, stderr) => {
```
jcarpanelli: WDYT if we simplify `Compiler#call` to something prettier like:

```ts
import { promisify } from 'util';
const execFileAsync = promisify(execFile);

const Compiler = {
  async call(): Promise<void | ExecException> {
    let truffleBin = `${process.cwd()}/node_modules/.bin/truffle`;
    log.info('Compiling contracts with Truffle...');
    if (!FileSystem.exists(truffleBin)) truffleBin = 'truffle'; // Attempt to load global truffle if local was not found
    try {
      const { stdout } = await execFileAsync(truffleBin, ['compile', '--all']);
      log.warn(stdout); //or log.info?
    } catch(error) {
      throw Error(error.stdout);
    }
  }
};
```
I believe that in both cases (whether or not there is a compilation error) `stderr` will be empty, and the output will be in `stdout`

Not a blocker though.

PR #745, opened 2019-02-26
Interactive verify

### LINKED ISSUE #719: Interactive verify command
Ask which contracts to verify by showing the user a list (allow them to select all). Only list pushed contracts here.

Prompt which remote to use (default to etherscan). Ask for API key if it's etherscan, and don't store it anywhere (for now). Load optimizer settings from truffle.js, or prompt user if not available (defaulting to no optimizer).

jcarpanelli:
Fixes #719.
(btw, I don't care about uploading the etherscan API KEY in the gif :joy:) 
Verify gif

I'm aware that the command tests are failing. Will fix them soon :man_health_worker: .

### DIFF
--- a/packages/cli/src/commands/verify.ts
+++ b/packages/cli/src/commands/verify.ts
@@ -21,7 +21,6 @@ const props = (optimizerEnabled) => {
     optimizerRuns: {
       type: 'input',
       message: 'Specify the optimizer runs',
-      default: '200',
       when: ({ optimizer }) => optimizer || optimizerEnabled
     },
     remote: {
@@ -52,7 +51,8 @@ async function action(contractName: string, options: any): Promise<void | never>
   const { optimizer, optimizerRuns, remote, apiKey, network: networkName } = options;
   const currentOpts = { network: networkName, optimizer, optimizerRuns, remote, apiKey };
 
-  const promptedArgsAndOpts = await promptIfNeeded({ args: { contractName }, opts: currentOpts, props: props(optimizer) });
+  const defaultArgs = Truffle.getCompilerInfo();
+  const promptedArgsAndOpts = await promptIfNeeded({ args: { contractName }, opts: currentOpts, defaults: defaultArgs, props: props(optimizer) });
   const { network, txParams } = await ConfigVariablesInitializer.initNetworkConfiguration(promptedArgsAndOpts);
 
   await verify(promptedArgsAndOpts.contractName, { ...promptedArgsAndOpts, network, txParams });
--- a/packages/cli/src/commands/verify.ts
+++ b/packages/cli/src/commands/verify.ts
@@ -2,29 +2,60 @@ import pickBy from 'lodash.pickby';
 
 import verify from '../scripts/verify';
 import ConfigVariablesInitializer from '../models/initializer/ConfigVariablesInitializer';
+import Truffle from '../models/initializer/truffle/Truffle';
+import { promptIfNeeded, getContractsList, getNetworkList } from '../utils/prompt';
 
 const name: string = 'verify';
-const signature: string = `${name} <contract-alias>`;
+const signature: string = `${name} [contract-alias]`;
 const description: string = 'verify a contract with etherscan or etherchain. Provide a contract name.';
+
+const props = (optimizerEnabled) => {
+  return {
+    ...getContractsList('contractName', 'Choose a contract', 'list'),
+    ...getNetworkList('list'),
+    optimizer: {
+      type: 'confirm',
+      message: 'Was the optimizer enabled when you compiled your contracts?',
+      default: false
+    },
+    optimizerRuns: {
+      type: 'input',
+      message: 'Specify the optimizer runs',
+      default: '200',
+      when: ({ optimizer }) => optimizer || optimizerEnabled
+    },
+    remote: {
+      type: 'list',
+      message: 'Select an endpoint',
+      choices: ['etherscan', 'etherchain'],
+      default: 'etherscan'
+    },
+    apiKey: {
+      type: 'input',
+      message: 'Provide an etherscan API KEY',
+      when: ({ remote }) => remote === 'etherscan'
+    },
+  };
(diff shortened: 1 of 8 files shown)


packages/cli/src/commands/verify.ts
```diff
@@ -2,29 +2,60 @@ import pickBy from 'lodash.pickby';
 
 import verify from '../scripts/verify';
 import ConfigVariablesInitializer from '../models/initializer/ConfigVariablesInitializer';
+import Truffle from '../models/initializer/truffle/Truffle';
+import { promptIfNeeded, getContractsList, getNetworkList } from '../utils/prompt';
 
 const name: string = 'verify';
-const signature: string = `${name} <contract-alias>`;
+const signature: string = `${name} [contract-alias]`;
```
spalladino: Keep in mind that if we later add a flag to disable interactive, we may need to keep the old signatures around

packages/cli/src/commands/verify.ts
```diff
@@ -2,29 +2,60 @@ import pickBy from 'lodash.pickby';
...
+    remote: {
+      type: 'list',
+      message: 'Select an endpoint',
+      choices: ['etherscan', 'etherchain'],
+      default: 'etherscan'
+    },
+    apiKey: {
+      type: 'input',
+      message: 'Provide an etherscan API KEY',
+      when: ({ remote }) => remote === 'etherscan'
```
spalladino: <3

packages/cli/src/commands/verify.ts
```diff
@@ -2,29 +2,60 @@ import pickBy from 'lodash.pickby';
...
-  .option('--optimizer-runs <runs>', 'specify number of runs if optimizer enabled.')
+  .option('--optimizer-runs [runs]', 'specify number of runs if optimizer enabled.')
   .option('--remote <remote>', 'specify remote endpoint to use for verification')
   .option('--api-key <key>', 'specify etherscan API key. To get one, go to: https://etherscan.io/myapikey')
   .action(action);
 
-async function action(contractAlias: string, options: any): Promise<void | never> {
-  const { optimizer, optimizerRuns, remote, apiKey } = options;
-  if (optimizer && !optimizerRuns) {
-    throw new Error('Cannot verify contract without defining optimizer runs');
```
spalladino: Same as before: we may need to keep some checks around for non-interactive mode. However, in this case, I think we could default to 200 runs instead of failing.

packages/cli/src/commands/verify.ts
```diff
@@ -2,29 +2,60 @@ import pickBy from 'lodash.pickby';
...
+const props = (optimizerEnabled) => {
+  return {
+    ...getContractsList('contractName', 'Choose a contract', 'list'),
+    ...getNetworkList('list'),
+    optimizer: {
+      type: 'confirm',
+      message: 'Was the optimizer enabled when you compiled your contracts?',
+      default: false
+    },
+    optimizerRuns: {
```
spalladino: Let's default to 200

packages/cli/src/commands/verify.ts
```diff
@@ -2,29 +2,60 @@ import pickBy from 'lodash.pickby';
...
   .option('--api-key <key>', 'specify etherscan API key. To get one, go to: https://etherscan.io/myapikey')
   .action(action);
 
-async function action(contractAlias: string, options: any): Promise<void | never> {
-  const { optimizer, optimizerRuns, remote, apiKey } = options;
-  if (optimizer && !optimizerRuns) {
-    throw new Error('Cannot verify contract without defining optimizer runs');
-  }
-  const { network, txParams } = await ConfigVariablesInitializer.initNetworkConfiguration(options);
-  const opts = pickBy({ optimizer, optimizerRuns, remote, apiKey, network, txParams });
```
spalladino: Was this variable even used?!

packages/cli/src/utils/prompt.ts
```diff
@@ -2,39 +2,64 @@ import isEmpty from 'lodash.isempty';
...
-export async function promptForArgumentsIfNeeded({ args, defaults, props }: PromptParams): Promise<any> {
-  const questions = Object.keys(args)
+export async function promptIfNeeded({ args = {}, opts = {}, defaults, props }: Args): Promise<any> {
+  const argsQuestions = Object.keys(args)
     .filter((argName) => !args[argName] || isEmpty(args[argName]))
     .map((argName) => promptFor(argName, defaults, props));
 
-  return { ...args, ...(await Inquirer.prompt(questions)) };
+  const optsQuestions = Object.keys(opts)
+    .filter((optName) => !opts[optName])
```
spalladino: What happens if the user explicitly sets `false` to an opt via command line? Shouldn't we be checking for undefined instead?

packages/cli/src/commands/verify.ts
jcarpanelli: nope :joy:

packages/cli/src/commands/verify.ts
jcarpanelli: I was thinking about doing these validations inside the scripts. WDYT?

packages/cli/src/commands/verify.ts
jcarpanelli: We are setting a default value for the `runs` field in the script function declaration :slightly_smiling_face:. Anyway, I will double-check this just to be sure we won't have any problem

packages/cli/src/commands/verify.ts
spalladino: Could work, and it's a good idea since it also checks programmatic usage. Let's see how it goes!

packages/cli/src/utils/prompt.ts
jcarpanelli: `false` would be passed as a string, so there shouldn't be any problem. Anyway, It's a good practice to be more explicit in these cases as it will be clearer to understand whether it should filter the option or not. Good gatch :slightly_smiling_face:

PR #750, opened 2019-02-27
Interactive link and unlink

jcarpanelli:
Fixes https://github.com/zeppelinos/zos/issues/720. Bonus track: interactive `unlink`ing.
link
Still polishing up some stuff.

EDIT: it seems that the `inquirer`'s checkbox feature does not work properly with `asciinema` :cry:

### DIFF
--- a/packages/cli/src/commands/link.ts
+++ b/packages/cli/src/commands/link.ts
@@ -30,7 +30,7 @@ const register: (program: any) => any = (program) => program
 
 async function action(dependencies: string[], options: any): Promise<void> {
   const { install: installDependencies, interactive } = options;
-  const promptedArgs = await promptIfNeeded({ args: { dependencies, installDependencies }, props }, interactive);
+  const promptedArgs = await promptIfNeeded({ args: { dependencies }, opts: { installDependencies }, props }, interactive);
   if (promptedArgs.dependencies && typeof promptedArgs.dependencies === 'string') {
     promptedArgs.dependencies = [promptedArgs.dependencies];
   }
--- a/packages/cli/src/commands/link.ts
+++ b/packages/cli/src/commands/link.ts
@@ -1,6 +1,7 @@
 import push from './push';
 import link from '../scripts/link';
 import { promptIfNeeded } from '../utils/prompt';
+import Dependency from '../models/dependency/Dependency';
 
 const name: string = 'link';
 const signature: string = `${name} [dependencies...]`;
@@ -10,7 +11,6 @@ const props = {
   dependencies: {
     type: 'input',
     message: 'Provide an EVM-package name and version',
-    default: ['openzeppelin-eth']
   },
   installDependencies: {
     type: 'confirm',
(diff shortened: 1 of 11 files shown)


packages/cli/src/commands/link.ts
```diff
@@ -1,21 +1,41 @@
...
 
 const name: string = 'link';
 const signature: string = `${name} [dependencies...]`;
 const description: string = 'links project with a list of dependencies each located in its npm package';
 
+const props = {
+  dependencies: {
+    type: 'input',
+    message: 'Provide an EVM-package name and version',
+    default: ['openzeppelin-eth']
```
spalladino: Let's somehow explain that the full format should be `name@version`

packages/cli/src/commands/unlink.ts
```diff
@@ -1,9 +1,20 @@
...
 
 const name: string = 'unlink';
 const signature: string = `${name} [dependencies...]`;
 const description: string = 'unlinks dependencies from the project. Provide a list of whitespace-separated dependency names';
+const props = (depNames) => {
+  return {
+    dependencies: {
+      message: 'Select the dependencies you want to unlink',
+      type: 'checkbox',
+      choices: depNames
```
spalladino: What happens if depNames is empty? We should handle that case nicely

packages/cli/src/models/files/ZosPackageFile.ts
```diff
@@ -18,6 +18,13 @@ export default class ZosPackageFile {
     publish: boolean;
   };
 
+  public static getInstalledDependencies(fileName: string = 'zos.json'): string[] {
+    const file = fs.parseJsonIfExists(fileName);
+    if (file && file.dependencies) {
+      return Object.keys(file.dependencies).map((depName) => `${depName}@${file.dependencies[depName]}`);
```
spalladino: I'm sure there is a lodash function to do this more elegantly

packages/cli/src/models/local/LocalController.ts
```diff
@@ -141,19 +141,32 @@ export default class LocalController {
...
-    await Promise.all(dependencies.map(async (depNameVersion: string) => {
+    const linkedDependencies = await Promise.all(dependencies.map(async (depNameVersion: string) => {
       const dependency = installDependencies
         ? await Dependency.install(depNameVersion)
         : Dependency.fromNameWithVersion(depNameVersion);
       this.packageFile.setDependency(dependency.name, <string>dependency.requirement);
+      return dependency.name;
     }));
+
+    if (linkedDependencies.length > 0) {
```
spalladino: Award for attention to detail :-)

packages/cli/src/commands/link.ts
jcarpanelli: Have a crazy idea :nerd_face:: we can run a `npm view openzeppelin-eth` to get the current package version and show `openzeppelin-eth@2.1.3` (or whichever the latest version is) instead of `openzeppelin-eth`. This way, we don't have to update the "example" version each time the openzeppelin team releases a new version.

jcarpanelli: Isi pisi:
```ts
  public static async fetchVersionFromNpm(name: string): Promise<string | null> {
    const execAsync = promisify(exec);
    try {
      const { stdout }= await execAsync(`npm view ${name} | grep latest`);
      const versionMatch = stdout.match(/([0-9]*\.){2}[0-9]*/)
      if (versionMatch.length) return versionMatch[0];
    } catch(error) {
      return null;
    }
  }

```

packages/cli/src/commands/unlink.ts
jcarpanelli: Good question. Will check it!

PR #756, opened 2019-03-01
Rename openzeppelin contracts in zos-lib

spalladino:
This generated artifacts with potentially duplicated names, for example if the user imported both App.sol and openzeppelin-eth (given that Ownable would be present in both). Due to https://github.com/zeppelinos/zos/issues/389, this causes errors when trying to `push` the project to any network.

The following two contracts trigger this issue:
```solidity
// Foo.sol
pragma solidity ^0.4.24;
import "zos-lib/contracts/application/App.sol"; // indirectly imports zos-lib/contracts/openzeppelin/Ownable.sol
contract Foo {
  function foo() public {
    App app = new App();
  }
}
```

```solidity
// Bar.sol
pragma solidity ^0.4.24;
import 'openzeppelin-eth/contracts/ownership/Ownable.sol';
contract Bar is Ownable {  }
```

Running `zos push` fails with `Cannot read property 'forEach' of undefined`, since it cannot find the source corresponding to one the `Ownable`s present in the project.

This commit renames them to prevent any conflicts with the contracts from openzeppelin. Note that it is pending to be cherry-picked into master, and we should also add a test to properly reproduce the case described above.

### DIFF
--- a/packages/lib/contracts/application/App.sol
+++ b/packages/lib/contracts/application/App.sol
@@ -3,14 +3,14 @@ pragma solidity ^0.4.24;
 import "./ImplementationProvider.sol";
 import "./Package.sol";
 import "../upgradeability/AdminUpgradeabilityProxy.sol";
-import "../openzeppelin-solidity/ownership/Ownable.sol";
+import "../ownership/Ownable.sol";
 
 /**
  * @title App
  * @dev Contract for upgradeable applications.
  * It handles the creation of proxies.
  */
-contract App is ZOSOwnable {
+contract App is ZOSLibOwnable {
   /**
    * @dev Emitted when a new proxy is created.
    * @param proxy Address of the created proxy.
--- a/packages/lib/contracts/application/App.sol
+++ b/packages/lib/contracts/application/App.sol
@@ -10,7 +10,7 @@ import "../openzeppelin-solidity/ownership/Ownable.sol";
  * @dev Contract for upgradeable applications.
  * It handles the creation of proxies.
  */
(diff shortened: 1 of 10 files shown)


spalladino:
Updated! @jcarpanelli mind re-reviewing?

packages/lib/contracts/ownership/Ownable.sol
```diff
@@ -4,9 +4,13 @@ pragma solidity ^0.4.24;
  * @title Ownable
  * @dev The Ownable contract has an owner address, and provides basic authorization control
  * functions, this simplifies the implementation of "user permissions".
+ *
  * Source: https://raw.githubusercontent.com/OpenZeppelin/openzeppelin-solidity/v2.0.0/contracts/ownership/Ownable.sol
+ * This contract is copied here and renamed from the original to avoid clashes in the compiled artifacts
```
jcarpanelli: Thanks for clarifying it in a comment :slightly_smiling_face:

packages/lib/contracts/application/App.sol
```diff
@@ -3,14 +3,14 @@ pragma solidity ^0.4.24;
 import "./ImplementationProvider.sol";
 import "./Package.sol";
 import "../upgradeability/AdminUpgradeabilityProxy.sol";
-import "../openzeppelin-solidity/ownership/Ownable.sol";
+import "../ownership/Ownable.sol";
```
jcarpanelli: I would also change the filename to `ZOSLibOwnable`, but it's ok like this.

Issue #754, opened 2019-03-01
Command `add --all` should not add libraries or contracts from another packages


spalladino:
Running `zos add --all` should only add to the project contracts that belong to the project itself. Contracts imported from other projects, such as Initializable, should not be added. Solidity libraries should never be added.

siromivel:
@jcarpanelli I believe this one was also addressed in https://github.com/zeppelinos/zos/pull/758 :smiley:

PR #758, opened 2019-03-04
zos add --all should only add non-library project contracts

siromivel:
Addresses #754

This fixes the issue where `zos add --all` adds non-project and library contracts. I identified project contracts by recursing through the contracts directory and aggregating an array of contract names stripped of file extensions. We then check against that array of project contracts while going through the built contracts to add and only add those that are found in the project. 

Libraries are excluded by checking the through the contracts' compiled JSON interface, this may be fragile as it appeared on first glance that the `contractKind` always appears on the second element of the `contractData.ast.nodes` array - if that's not the case it would be simple to refactor this to iteratively check for that value.

### DIFF
--- a/packages/cli/src/models/initializer/truffle/Truffle.ts
+++ b/packages/cli/src/models/initializer/truffle/Truffle.ts
@@ -72,8 +72,11 @@ const Truffle = {
         .filter((name) => {
           const contract = FileSystem.parseJsonIfExists(`${buildDir}/${name}`);
           if (contract) {
+            let isLibrary = contract.ast ? contract.ast.nodes[1].contractKind === "library" : false;
             const projectDir = buildDir.replace('build/contracts', '');
-            return contract.bytecode.length > 2 && contract.sourcePath.indexOf(projectDir) === 0;
+            return (contract.bytecode.length > 2 &&
+                    contract.sourcePath.indexOf(projectDir) === 0) &&
+                    !isLibrary
           } else return false;
         })
         .map((name) => name.replace('.json', ''));
--- a/packages/cli/src/models/initializer/truffle/Truffle.ts
+++ b/packages/cli/src/models/initializer/truffle/Truffle.ts
@@ -71,12 +71,13 @@ const Truffle = {
         .filter((name) => {
           const contract = FileSystem.parseJsonIfExists(`${buildDir}/${name}`);
           if (contract) {
-            const isLibrary = contract.ast && contract.ast.nodes.find((node) => node.name === contract.contractName && node.contractKind === 'library');
+            const isLibrary = contract.ast && contract.ast.nodes
+              .find((node) => node.name === contract.contractName && node.contractKind === 'library');
             const projectDir = buildDir.replace('build/contracts', '');
 
-            return (contract.bytecode.length > 2 &&
-                    contract.sourcePath.indexOf(projectDir) === 0 &&
-                    !isLibrary);
+            return !isLibrary
+                    && contract.sourcePath.indexOf(projectDir) === 0
+                    && contract.bytecode.length > 2;
           } else return false;
         })
         .map((name) => name.replace('.json', ''));
--- a/packages/cli/src/models/initializer/truffle/Truffle.ts
+++ b/packages/cli/src/models/initializer/truffle/Truffle.ts
@@ -75,8 +75,8 @@ const Truffle = {
             let isLibrary = contract.ast ? contract.ast.nodes[1].contractKind === "library" : false;
             const projectDir = buildDir.replace('build/contracts', '');
             return (contract.bytecode.length > 2 &&
-                    contract.sourcePath.indexOf(projectDir) === 0) &&
-                    !isLibrary
+                    contract.sourcePath.indexOf(projectDir) === 0 &&
+                    !isLibrary);
           } else return false;
         })
         .map((name) => name.replace('.json', ''));
--- a/packages/cli/src/models/initializer/truffle/Truffle.ts
+++ b/packages/cli/src/models/initializer/truffle/Truffle.ts
@@ -1,6 +1,5 @@
 import pickBy from 'lodash.pickby';
 import pick from 'lodash.pick';
-import { promisify } from 'util';
 import { FileSystem } from 'zos-lib';
 
 const Truffle = {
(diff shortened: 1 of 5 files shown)


spalladino:
This is great, thanks so much @siromivel! @jcarpanelli could you take a look at this, since you are currently working on the commands interface?

siromivel:
I just pushed a quick fix forr #760 to this branch as I agree it makes sense to clean up these issues together - let me know if that's an issue and I can prune that commit.

packages/cli/src/models/initializer/truffle/Truffle.ts
```diff
@@ -72,8 +72,11 @@ const Truffle = {
         .filter((name) => {
           const contract = FileSystem.parseJsonIfExists(`${buildDir}/${name}`);
           if (contract) {
+            let isLibrary = contract.ast ? contract.ast.nodes[1].contractKind === "library" : false;
             const projectDir = buildDir.replace('build/contracts', '');
-            return contract.bytecode.length > 2 && contract.sourcePath.indexOf(projectDir) === 0;
+            return (contract.bytecode.length > 2 &&
+                    contract.sourcePath.indexOf(projectDir) === 0) &&
+                    !isLibrary
```
siromivel: This addresses #760 - note that it will completely omit libraries from the return of this method, if that's an issue this could be done in a more granular fashion. All tests passed locally and I couldn't immediately see a problem with omitting the libraries.

packages/cli/src/models/local/LocalController.ts
```diff
@@ -57,13 +57,37 @@ export default class LocalController {
   }
 
   public addAll(): void {
+    function trimExtension(file: string) {
+      let fileExtRegex = new RegExp(/\.[^\.]*$/);
+      return file.replace(fileExtRegex, "");
+    }
+
+    function getAllFilenamesAtPath(path: string): string[] {
+      return fs.readDir(path).reduce((contracts: any, child: string) => {
```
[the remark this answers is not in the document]
siromivel: This function is actually completely unnecessary now that I know relying on `sourcePath` is reasonable; it needed to be recursive originally as it was meant to go through the uncompiled contracts directory and in the test cases was not finding contracts in subdirectorys (specifically the `/contracts/mocks` path where the various Impl contracts live.)

Next commit will remove this monstrosity in favor of the simpler solution :grin:

siromivel:
I'll look at fleshing out the test cases this afternoon and get the PR updated with those!

siromivel:
Just pushed a commit cleaning up the logic for checking if a contract belongs to the project and broke the tests for this out into their own test block. I also added a `WithExternalPackageImplV1` contract to `ImplV1.sol` that relies on the `mock-stdlib` package and assertions ensuring the external package and library aren't being added.

packages/cli/src/models/initializer/truffle/Truffle.ts
```diff
@@ -72,8 +72,12 @@ const Truffle = {
         .filter((name) => {
           const contract = FileSystem.parseJsonIfExists(`${buildDir}/${name}`);
           if (contract) {
+            const isLibrary = contract.ast ? contract.ast.nodes[1].contractKind === "library" : false;
```
jcarpanelli: I believe this is not generic enough. Image that you have a `Foo.sol` in which you have, some contracts and libraries defined:
```solidity
pragma solidity ^0.5.0;

contract MyContract {}
library MyLibrary {}
library MyOtherLibrary {}
```
In this case, when reading the contract ast, you will find the pragma-related info in the first position and the `MyContract` contract-related info in the second position (i.e., `contract.ast[1]`), and not the info related to the library. You will struggle with the same problem if you have imports before declaring your libraries.
In order to solve this, you can check not only the `contractKind` value but also the `name` to be sure you are picking the right element of the `nodes` array.
For instance:
```ts
const isLibrary = !!contract.ast.nodes.find(node => node.contractKind === 'library' && node.name === contract.contractName);
```
(Please double check that I'm not commiting any mistake with the node's attributes :laughing:)

jcarpanelli: nit: use single quotes instead of double quotes or the linter will complain.

packages/cli/src/models/local/LocalController.ts
```diff
@@ -57,13 +57,20 @@ export default class LocalController {
...
+    const buildFolder = Contracts.getLocalBuildDir();
+    const sourceFolder = Contracts.getLocalContractsDir();
+
+    fs.readDir(buildFolder).forEach((file) => {
+      const path = `${buildFolder}/${file}`;
       if(this.hasBytecode(path)) {
         const contractData = fs.parseJson(path);
-        const contractName = contractData.contractName;
-        this.add(contractName, contractName);
+        const isLibrary = contractData.ast.nodes[1].contractKind === "library";
```
jcarpanelli: Same as above regarding how to pick the correct ast node item and regarding the quotation.

packages/cli/test/scripts/add.test.js
```diff
@@ -78,6 +79,17 @@ contract('add script', function() {
     this.packageFile.contract('ImplV2').should.eq('ImplV2')
   })
 
+  it('should not add solidity libraries or external packages when adding all', function() {
```
jcarpanelli: nit: `should not add solidity libraries or contracts from external packages when adding all`

siromivel:
Took care of the small stuff in here yesterday; planning to take a deeper look at the nodes data structure and get this finished up this afternoon 🚀

packages/cli/src/models/local/LocalController.ts
```diff
@@ -64,10 +64,10 @@ export default class LocalController {
       const path = `${buildFolder}/${file}`;
       if(this.hasBytecode(path)) {
         const contractData = fs.parseJson(path);
-        const isLibrary = contractData.ast.nodes[1].contractKind === "library";
         const isProjectContract = contractData.sourcePath.indexOf(sourceFolder) === 0;
+        const isLibrary = contractData.ast && contractData.ast.nodes.find((node) => node.name === contractData.contractName && node.contractKind === 'library');
```
siromivel: The existence check for contractData.ast may be unnecessary - it was an edge case I ran into in the Truffle tests when working with the contract with bytecode `0x`, that contract has no ast on it. I spent a while trying to compose a contract that would compile into the same an was unsuccessful; a completely blank contract file won't compile at all and even just a pragma declaration will generate some viable bytecode. If there's no real world scenario that leads to this I'd suggest possibly modifying the truffle test and removing this existence check.

siromivel:
@jcarpanelli I took care of the feedback this afternoon; an integration test is timing out on Travis but I think it may just be flaky - ran the suite 5 times locally and the same timeout occured on one of those runs.

I was also thinking it may be nice to abstract the logic for checking if a contract is a library so that we can avoid duplicating that code(with this PR it will be used in both `Truffle.js` and `add.test.js`). I poked around the repo a bit and found the various `helpers` and `utilities` folders. It also might make sense to expose it as a property at the `contract` level so that we don't have to traverse the nodes every time we want to check it. Would one of those be an appropriate location for this logic or is there a better place that I'm not familiar with yet? Thanks!

PR #761, opened 2019-03-04
initializeT

acrdlph:
t should be checked not _t


spalladino:
Good catch, thanks a lot @acrdlph!
@ajsantander or @jcarpanelli, can we get this fix into the 2.2 docs and redeploy?

jcarpanelli:
> @ajsantander or @jcarpanelli, can we get this fix into the 2.2 docs and redeploy?

Done :slightly_smiling_face:

Issue #759, opened 2019-03-04
Windows spawn ENOENT on truffle compile


imbenwolf:
When executing "zos add XY" on Windows 10, I get following error:
`
spawn C:\PATH\TO\PROJECT\node_modules\.bin\truffle ENOENT
`

When replacing line 13 in zos/packages/cli/src/models/compiler/Compiler.ts from:
`
execFile(truffleBin, ['compile', '--all'], (error: ExecException, stdout, stderr) => {
`
to:
``
exec(`${truffleBin} compile --all`, (error: ExecException, stdout, stderr) => {
``
Everything works fine. 

I propose to change this to fix issue on Windows.

siromivel:
I just opened a PR for this - I can't test it in a Windows environment until tomorrow but will update once I do!

spalladino:
Fixed by @siromivel in #764

PR #764, opened 2019-03-05
Fix `zos add` in Windows environments

siromivel:
Addresses #759 

This was a very quick fix and I have both Windows and Linux environments available at home. I've tested it in my Linux environment and everything seems fine; I'll try running the tests in my Windows environment tomorrow when I get home.

### DIFF
--- a/packages/cli/src/models/compiler/Compiler.ts
+++ b/packages/cli/src/models/compiler/Compiler.ts
@@ -1,4 +1,4 @@
-import { execFile, ExecException } from 'child_process';
+import { exec, ExecException } from 'child_process';
 import { FileSystem, Contracts, Logger } from 'zos-lib';
 
 const log = new Logger('Compiler');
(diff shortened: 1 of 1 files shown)


siromivel:
Ah unfortunately I don't have access to a Mac machine for testing at the moment; if there is an issue there it looks like it should be possible to check if the OS is windows-based and conditionally run an `exec` in that case although it would be nicer to not need switching logic.

spalladino:
Actually the issue with having spaces in the path affected both Linux and MacOS. The quick and dirty solution was to quote the executable, but using `execFile` instead was a cleaner solution. That said, I didn't expect it to break Windows, though it seems that the differences are [platform-dependent indeed](https://nodejs.org/api/child_process.html#child_process_spawning_bat_and_cmd_files_on_windows).

@siromivel can I ask you to test using `execFile` with the `shell` flag set to true, and see if it works in Windows? I don't have a Windows dev station setup here to test it out now. Thanks!

siromivel:
Just an update: I'm working on testing this on my Windows machine, running into some issues as the current User directory has spaces and is causing different issues 😂

siromivel:
Ok I've confirmed that simply passing in the `shell: true` arg fixes this issue on a Windows machine and doesn't appear to break Linux compatibility. I can't test this on MacOS currently.

image
🍾

packages/cli/src/models/compiler/Compiler.ts
```diff
@@ -1,5 +1,5 @@
 import { execFile, ExecException } from 'child_process';
-import { FileSystem, Contracts, Logger } from 'zos-lib';
+import { FileSystem, Logger } from 'zos-lib';
```
siromivel: This was an unused import

siromivel:
I'm seeing a build issue on Travis with this change, I'll look into that a bit later this morning.

siromivel:
Build is fixed - I was passing the args as an object literal which TypeScript didn't like, not sure why it was compiling locally but moving those out into a const assignment got the build working again.

PR #766, opened 2019-03-07
interactive push (first part)

### LINKED ISSUE #707: Interactive prompt to deploy dependencies if they are not present in the current network
If the user is working on a network (such as rinkeby) with an evm package that is not available on that network, then (instead of failing) we should prompt whether they want to deploy that dependency for testing.

If the node is ganache, then deploy anyway without even asking.

### LINKED ISSUE #716: Interactive network params
When the user runs a command that requires a network info (such as create or push), ask which network they want to connect to. Next time the user is prompted about this, default to the latest choice (consider saving it as a session with duration 0 as a hack to implement it).

jcarpanelli:
Fixes #707, fixes #716.

This PR includes:
- Automatic deployment of dependencies if working with `ganache`
- Prompt to deploy dependencies if necessary
- Save last network used for using it as default for next time

interactive-push

EDIT: I also refactored a bit the commands tests. Sorry for doing  this here PR :eyes:

### DIFF
--- a/packages/cli/src/models/dependency/Dependency.ts
+++ b/packages/cli/src/models/dependency/Dependency.ts
@@ -29,7 +29,7 @@ export default class Dependency {
   }
 
   public static satisfiesVersion(version: string | semver.SemVer, requirement: string | semver.Range): boolean {
-    return !requirement || version === requirement || semver.satisfies(version, requirement);
+    return !requirement || version === requirement || semver.satisfies(semver.coerce(version), requirement);
   }
 
   public static async fetchVersionFromNpm(name: string): Promise<string> {
@@ -41,6 +41,20 @@ export default class Dependency {
     } catch(error) {
       return name;
     }
+  }
+
+  public static hasDependenciesForDeploy(network: string): boolean {
+    const dependencies = ZosPackageFile.getLinkedDependencies() || [];
+    const networkDependencies = ZosNetworkFile.getDependencies(`zos.${network}.json`) || {};
+    const hasDependenciesForDeploy = dependencies.find(depNameAndVersion => {
+      const [name, version] = depNameAndVersion.split('@');
+      const networkFilePath = `node_modules/${name}/zos.${network}.json`;
+      const projectDependency = networkDependencies[name];
+      const satisfiesVersion = projectDependency && this.satisfiesVersion(projectDependency.version, version);
+      return !fs.exists(networkFilePath) && !satisfiesVersion;
+    });
+
+    return !!hasDependenciesForDeploy;
   }
 
   public static async install(nameAndVersion: string): Promise<Dependency> {
--- a/packages/cli/src/models/files/ZosNetworkFile.ts
+++ b/packages/cli/src/models/files/ZosNetworkFile.ts
@@ -77,6 +77,11 @@ export default class ZosNetworkFile {
     return file ? file.zosversion : null;
   }
 
+  public static getDependencies(fileName: string): { [key: string]: any } | undefined {
+    const file = fs.parseJsonIfExists(fileName);
+    return file ? file.dependencies : undefined;
+  }
+
   // TS-TODO: type for network parameter (and class member too).
   constructor(packageFile: ZosPackageFile, network: any, fileName: string) {
     this.packageFile = packageFile;
--- a/packages/cli/src/models/network/Session.ts
+++ b/packages/cli/src/models/network/Session.ts
@@ -14,21 +14,33 @@ interface SessionOptions {
   network?: string;
   from?: string;
   timeout?: number;
+  expires?: Date;
 }
 
 const Session = {
 
   getOptions(overrides: SessionOptions = {}): SessionOptions {
     const session = this._parseSession();
-    if (!session) return this._setDefaults(overrides);
+    if (!session || this._hasExpired(session)) return this._setDefaults(overrides);
     log.info(`Using session with ${describe(omitBy(session, (v, key) => overrides[key]))}`);
+
     return { ...session, ...overrides };
   },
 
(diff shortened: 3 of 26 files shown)


packages/cli/src/models/dependency/Dependency.ts
```diff
@@ -43,6 +43,20 @@ export default class Dependency {
     }
   }
 
+  public static hasDependenciesForDeploy(network: string): boolean {
+    const dependencies = ZosPackageFile.getLinkedDependencies() || [];
+    const networkDependencies = ZosNetworkFile.getDependencies(`zos.${network}.json`) || {};
+    const hasDependenciesForDeploy = dependencies.find(depNameAndVersion => {
```
spalladino: Can we use `isDeployedOnNetwork`, or a modified version if needed, to avoid remplementing this logic here?

packages/cli/src/models/dependency/Dependency.ts
```diff
@@ -43,6 +43,20 @@ export default class Dependency {
     }
   }
 
+  public static hasDependenciesForDeploy(network: string): boolean {
```
spalladino: How about changing to a `getDependenciesPendingDeployment`, so we can actually show the user a list of the deps that would get deployed? Should be just a matter of changing the `find` to a `filter`.

packages/cli/src/models/files/ZosNetworkFile.ts
```diff
@@ -77,6 +77,11 @@ export default class ZosNetworkFile {
     return file ? file.zosversion : null;
   }
 
+  public static getDependencies(fileName: string): { [key: string]: any } | undefined {
```
spalladino: Can we remove this method, and use something like `packageFile.networkFile(network).dependencies`? I'd assume that networkFile for a new network would yield an empty ZosNetworkFile, that would have empty dependencies as well.

packages/cli/src/models/network/Session.ts
```diff
@@ -14,21 +14,33 @@ interface SessionOptions {
...
 
-  open({ network, from, timeout }: SessionOptions, expires: number = DEFAULT_EXPIRATION_TIMEOUT): void {
+  setDefaultNetworkIfNeeded(network: string): void {
+    const session = this._parseSession();
+    if (!session || this._hasExpired(session)) this.open({ network }, 0, false);
+  },
+
+  getDefaultNetwork(): string | undefined {
+    const session = this._parseSession();
+    return session && this._hasExpired(session) ? session.network : undefined;
```
spalladino:
```suggestion
    return session ? session.network : undefined;
```

I think that the session network can still be considered to be the default network regardless of whether it had expired or not.

packages/cli/test/commands/push.test.js
```diff
@@ -1,14 +1,30 @@
...
-  itShouldParse('should call push script with options', 'push', 'zos push --network test --skip-compile -d --reset -f', function(push) {
-    push.should.have.been.calledWithExactly({ force: true, deployDependencies: true, reupload: true, network: 'test', txParams: {} })
-  })
+  context('when network uses ganache', function() {
+    itShouldParse('should call push script with options', 'push', 'zos push --network test --skip-compile -d --reset -f', function(push) {
+      push.should.have.been.calledWithExactly({ force: true, deployDependencies: true, reupload: true, network: 'test', txParams: {} })
+    })
+  });
+
+  context('when network does not use ganache', function () {
```
spalladino: Awesome

packages/cli/test/scripts/session.test.js
```diff
@@ -11,22 +11,22 @@ describe('session script', function () {
   const opts = { network: 'foo', from: '0x1', timeout: 10 }
 
   describe('opening a new session', function () {
-    describe('when there was no session opened before', function () {
-      describe('when the time out does not expire', function () {
+    context('when there was no session opened before', function () {
```
spalladino: Just curious: why this change?

tslint.json
```diff
@@ -49,7 +49,8 @@
       "adjacent-overload-signatures": [false],
       "no-string-literal": [false],
       "max-classes-per-file": [false],
-      "one-variable-per-declaration": [false]
+      "one-variable-per-declaration": [false],
+      "arrow-parens": [false]
```
spalladino: I typically dislike a PR that attempts to sneak in a change like this, but since this one is so awesome, I'll look the other way :-P

packages/cli/src/models/dependency/Dependency.ts
jcarpanelli: I'll address it in my TODO list, an evaluate in which cases we should stick to class methods and when it's appropriate to instantiate the class.

packages/cli/src/models/dependency/Dependency.ts
jcarpanelli: Its a good idea :slightly_smiling_face:

packages/cli/test/scripts/session.test.js
jcarpanelli: I know that `context` is a synonym for `describe`, but when talking about contexts (*when X is happening*) I prefer using context rather than `describe`. FYI, I started drafting an issue (pending to publish) with a list of conventions we should follow.

packages/cli/src/models/network/Session.ts
jcarpanelli: Mmmh but if the session hasn't expired, we won't be prompting the network list (and thus we won't show a default network)

spalladino: Gotcha! I had understood that this was used not for skipping the prompt, but for providing a default option among the networks listed.

PR #770, opened 2019-03-08
Skip validation errors upon push force

### LINKED ISSUE #755: Skip errors during contracts validation if force is supplied
There are several errors that occur during contracts validation that are related to incorrect artifacts compilations (such as #465). We should try-catch the validation of every contract, warn the user if an error occured during the process, and allow them to proceed with the `zos push` if `--force` was included (or prompt on whether to continue if in interactive mode).

spalladino:
Fixes #755

### DIFF
--- a/packages/cli/src/models/network/NetworkController.ts
+++ b/packages/cli/src/models/network/NetworkController.ts
@@ -305,7 +305,7 @@ export default class NetworkController {
       contract.schema.storageInfo = getStorageLayout(contract, buildArtifacts);
       return validationPasses(newWarnings);
     } catch (err) {
-      log.error(`Error while validating contract ${contract.schema.contractName}: ${err.message}`);
+      log.error(`Error while validating contract ${contract.schema.contractName}`, err);
       return false;
     }
   }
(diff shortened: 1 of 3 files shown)


packages/cli/src/models/network/NetworkController.ts
```diff
@@ -210,10 +210,9 @@ export default class NetworkController {
       await this._setSolidityLibs(contract);
       log.info(`Uploading ${contract.schema.contractName} contract as ${contractAlias}`);
       const contractInstance = await this.project.setImplementation(contract, contractAlias);
+      const { types, storage } = contract.schema.storageInfo || { types: null, storage: null };
```
jcarpanelli: If there's no difference between having `type` and `storage` assigned as `undefined` instead of `null`, you can simply change this to:
```suggestion
      const { types, storage } = contract.schema.storageInfo;
```

packages/lib/src/utils/Logger.ts
```diff
@@ -36,8 +36,16 @@ export default class Logger {
     this.log(msg, 'yellow');
   }
 
-  public error(msg: string): void {
-    this.log(msg, 'red');
+  public error(msg: string, ex?: Error): void {
```
jcarpanelli: :+1:

packages/cli/src/models/network/NetworkController.ts
spalladino: That was my first try, but actually, the ts compiler complained if I did that

Issue #767, opened 2019-03-07
Add test for truffle-migrate example


spalladino:
The truffle-migrate example is not currently being compiled nor tested in the CI, though testing it should be straightforward by running the commands listed on its README in a bash script that is exercised by the CI.

> To run this project locally, first install all dependencies with `npm install`, and then start a ganache instance by  running `npx ganache-cli -p 9545`. Next, run all migrations via `npx truffle migrate --network local`. Finally, check  the state of the deployment via `npm run check`.

The only change required should be modifying `index.js` (the script called as part of `npm run check`) to `throw` (and thus have a non-zero exit code) if the value returned is not the expected one (52 IIRC).

This test should be somehow added to `.travis.yml` as another stage to be run.

PR #775, opened 2019-03-09
Add test to truffle-migrate example #767

paulinablaszk:
Add test for truffle-migrate example

### DIFF
--- a/examples/truffle-migrate/index.js
+++ b/examples/truffle-migrate/index.js
@@ -8,6 +8,9 @@ async function main() {
   const value = await deployed.value();
   const version = await deployed.version();
   console.log(`Value of MyContract (${version}) deployed instance: ${value.toNumber()}`);
+  if(value != 52){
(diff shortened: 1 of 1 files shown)


examples/truffle-migrate/index.js
```diff
@@ -8,6 +8,9 @@ async function main() {
   const value = await deployed.value();
   const version = await deployed.version();
   console.log(`Value of MyContract (${version}) deployed instance: ${value.toNumber()}`);
+  if(value != 52){
```
jcarpanelli:
```suggestion
  if (value !== 52) {
```

paulinablaszk: Hi :) This PR adds a test to script in `index.js` that throws an error if the returned value is not the expected one.
fixes #767

spalladino:
@jcarpanelli I think this PR is good to merge, WDYT?

Issue #774, opened 2019-03-08
Validate if salt has been already used for deployment in current network


spalladino:
Before running a `zos create2` command (without the `--query` flag), we should check whether the target deployment address is already taken or not (a simple getCodeAt should do the trick), and fail elegantly.

spalladino:
Fixed in #788

PR #788, opened 2019-03-15
feature/check address prior running create2 #774

siromivel:
I wanted to look at the `create2` stuff a bit more and this seemed like a good issue(#774) to dive into it with. Added a check in `NetworkController.createProxy` to ensure the deployment address isn't occupied prior to calling `createProxyWithSalt`. Added a case to `create.test.js` covering this change.

### DIFF
--- a/packages/cli/src/models/network/NetworkController.ts
+++ b/packages/cli/src/models/network/NetworkController.ts
@@ -505,6 +505,12 @@ export default class NetworkController {
     this.checkInitialization(contract, initMethod, initArgs);
 
     const createArgs = { packageName, contractName: contractAlias, initMethod, initArgs };
+
+    if (salt) {
+      const deploymentAddress = await this.getProxyDeploymentAddress(salt);
+      if (await ZWeb3.getCode(deploymentAddress) != '0x') throw new Error("deployment address already in use");
+    }
+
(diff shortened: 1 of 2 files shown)


thesolyd:
This is not working guys. Everytime I get a new address.
const instance = await project.createProxyWithSalt(MyContractV0,"2");

spalladino:
Hey @thesolyd, can I ask you to ask in the forum, sharing the full code snippet you are running? Also, if you are running the same script multiple times, note that creating a new `Project` instance will be deploying a new `ProxyFactory` every time, causing addresses to change on every run.

You should initialize the Project with an existing proxyFactory address, to make sure it does not deploy a new one:
https://github.com/OpenZeppelin/openzeppelin-sdk/blob/3e2860ef7d5ea2d0397e0c5d25ed1d215d1e79f2/packages/lib/src/project/ProxyAdminProject.ts#L13-L22

To retrieve the proxyFactory address, you can call `ensureProxyFactory` which will deploy it if needed, and return the address:
https://github.com/OpenZeppelin/openzeppelin-sdk/blob/3e2860ef7d5ea2d0397e0c5d25ed1d215d1e79f2/packages/lib/src/project/BaseSimpleProject.ts#L201-L206

PR #794, opened 2019-03-28
Migrate to CircleCI

spalladino:
This PR adds configuration for CircleCI (see [here](https://circleci.com/workflow-run/a59c4bc2-82b1-4160-bf24-193d204fbae6) for a successful run of this branch), which yielded much faster results than travis (even when comparing a free plan to a paid one), and provides ssh login to the test boxes for debugging.

As part of this PR, I also restructured the folders for the integration tests, so that `workdir` was not a subfolder of `test`. Apparently, lerna cannot properly manage one package nested within another; this could have been the reason why we needed to run several `bootstrap`s within the integration test. Note that this broke travis tests, unfortunately.

I also leveraged circle's secondary containers to pull geth as a second container instead of installing it on every run. This required a few changes on the test setup as well.

Last but not least, I also renamed most packages to follow a convention: packages now begin with zos, example, or test depending their role. This allows targeting them with lerna scopes much more easily.

Also, fixes https://github.com/zeppelinos/zos/issues/661.


tests/cli/test/src/integration.test.js
```diff
@@ -8,7 +8,8 @@ const {
   getNetwork,
   truffleExec,
   run,
-  copy
+  copy,
+  setMockStdlibVersion
```
jcarpanelli: As we are going to do a bit of tech debt this sprint (are we? :stuck_out_tongue:), we may want to replace all the `stdlib` to `dependency` or `package` as appropiate.

.githooks/pre-commit
```diff
@@ -1,2 +1,2 @@
 #!/bin/sh
-lerna run lint
+./node_modules/.bin/lerna run lint
```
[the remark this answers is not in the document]
spalladino: Got it! I'll let you close it when you think it's ok.

tests/cli/test/src/integration.test.js
spalladino: Hmm I'd prefer to tackle other more pressing tech debt items than renaming stdlib nomenclature in tests.

PR #804, opened 2019-04-04
Adds ability to change owner of a proxy admin to set-admin command

ylv-io:
* Adds required functionality to a `set-admin` command in the following way:
`zos set-admin <new-proxy-admin-address>`
* Use interactive check to prevent admin address input errors.
* Adds tests to support required changes.

### DIFF
--- a/packages/cli/src/commands/set-admin.ts
+++ b/packages/cli/src/commands/set-admin.ts
@@ -1,4 +1,5 @@
 import pickBy from 'lodash.pickby';
+import inquirer from 'inquirer';
 
 import setAdmin from '../scripts/set-admin';
 import { fromContractFullName } from '../utils/naming';
@@ -7,13 +8,16 @@ import ConfigVariablesInitializer from '../models/initializer/ConfigVariablesIni
 
 const name: string = 'set-admin';
 const signature: string = `${name} [alias-or-address] [new-admin-address]`;
-const description: string = 'change upgradeability admin of a contract instance. Provide the [alias] or [package]/[alias] of the contract to change the ownership of all its instances, or its [address] to change a single one. Note that if you transfer to an incorrect address, you may irreversibly lose control over upgrading your contract.';
+const description: string = `change upgradeability admin of a contract instance, all instances or proxy admin.
+Provide the [alias] or [package]/[alias] of the contract to change the ownership of all its instances,
+or its [address] to change a single one, or none to change all contract instances to a new admin. 
+Note that if you transfer to an incorrect address, you may irreversibly lose control over upgrading your contract.`;
 
 const register: (program: any) => any = (program) => program
   .command(signature, undefined, { noHelp: true })
   .usage('[alias-or-address] [new-admin-address] --network <network> [options]')
   .description(description)
-  .option('-y, --yes', 'accept transferring admin rights (required)')
+  .option('-f, --force', 'bypass a manual check')
   .withNetworkOptions()
   .action(action);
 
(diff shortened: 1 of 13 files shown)


packages/lib/test/src/ProxyAdmin.test.js
```diff
@@ -11,8 +11,8 @@ const ImplV1 = Contracts.getFromLocal('DummyImplementation');
 const ImplV2 = Contracts.getFromLocal('DummyImplementationV2');
 const ProxyAdminContract = Contracts.getFromLocal('ProxyAdmin');
 
-contract('ProxyAdmin class', function(accounts) {
-  const [_, proxyAdminOwner, newAdmin, otherAccount] = accounts.map(utils.toChecksumAddress);
+contract.only('ProxyAdmin class', function(accounts) {
```
spalladino: Watch out! By the way, we need to add something that makes the CI fail if there is a `.only` in the specs. Or a git hook that runs on commit and checks it.

packages/cli/test/commands/setAdmin.test.js
```diff
@@ -3,22 +3,26 @@ require('../setup')
 
 import { stubCommands, itShouldParse } from './share';
 
-describe('set-admin command', function() {
+describe.only('set-admin command', function() {
```
spalladino: Remember to remove the `.only` from the tests!

packages/lib/test/src/ProxyAdmin.test.js
ylv-io: Would great to have cli to catch `.only`. Funny enough I was thinking the same.

packages/cli/src/commands/set-admin.ts
```diff
@@ -7,29 +8,48 @@ import ConfigVariablesInitializer from '../models/initializer/ConfigVariablesIni
...
+    newAdmin = contractFullNameOrAddress;
+    contractFullNameOrAddress = '';
+  }
+
+  if (!options.force) {
+    const answers = await inquirer.prompt([
+      {
+        name: 'address',
+        type: 'string',
+        message: 'Please tripe check your address and type the last 4 letters of the new admin address. If you provide a wrong address you will lose control forever.',
```
spalladino:
```suggestion
        message: 'Please double check your address and type the last 4 characters of the new admin address. If you provide a wrong address, you will lose control over your contracts.',
```

packages/cli/src/commands/set-admin.ts
```diff
@@ -7,29 +8,48 @@ import ConfigVariablesInitializer from '../models/initializer/ConfigVariablesIni
...
+      }
+    ]);
+    if (answers.address !== newAdmin.slice(-4)) {
+      throw Error('Addresses do not much. Pay attention to your address. If you provide wrong address you will lose control forever.');
```
spalladino:
```suggestion
      throw Error('Last 4 characters of the admin address do not match');
```
Keep an eye on spelling for `match`!

packages/cli/package-lock.json
```diff
@@ -1,6639 +1,6099 @@
...
-		},
-		"dom-walk": {
-			"version": "0.1.1",
-			"resolved": "https://registry.npmjs.org/dom-walk/-/dom-walk-0.1.1.tgz",
-			"integrity": "sha1-ZyIm3HTI95mtNTB9+TaroRrNYBg="
-		},
-		"domelementtype": {
-			"version": "1.3.1",
-			"resolved": "https://registry.npmjs.org/domelementtype/-/domelementtype-1.3.1.tgz",
-
```
spalladino: This is odd - why this change on the CLI package-lock? If there were no changes to dependencies, maybe remove it from the PR?

ylv-io: Hm. Strange. Should be no changes in package-lock at all.

PR #805, opened 2019-04-04
CREATE2 with signer

### LINKED ISSUE #731: Eagerly deploy proxyAdmin upon publish
The proxyAdmin instance is being deployed lazily when the first proxy is created. This makes it difficult to create a Solidity "proxy factory" that creates upgradeable instances and assigns them to the proxyAdmin, since a random proxy must be created first in order to set it up.

Instead, when the project is published, let's eagerly create the proxyAdmin.

### LINKED ISSUE #795: Update command fails to update proxies explicitly owned by the proxy admin
The upgrade command only attempts to upgrade proxies whose ownership has not been transferred. It does this by checking whether each proxy in the network file has an `admin` property defined: if it doesn't, it assumes it can still control it; if it does, it checks it against the current admin.

The issue is that it is picking the current admin as the `from` address, and not as the `proxyAdmin` contract. See https://github.com/zeppelinos/zos/blob/master/packages/cli/src/models/network/NetworkController.ts#L680.

### LINKED ISSUE #799: Add meta transaction support to proxy factory
The current implementation of `zos create2` depends on the `msg.sender` for determining the deployment address. This makes it impossible for meta-txs to be used for deployment, since the sender is an arbitrary relay and not the end user.

Add support for accepting a signature on the creation parameters, and use the signer instead of the sender in determining the deployment address.

spalladino:
This PR adds support for running CREATE2 in the ProxyFactory using a signer instead of the sender for determining the deployment address, allowing for a relayer to execute the deployment of the proxy (as long as the original user signs the initialization data).

It adds a `--signature` option to `zos create2` that will be sent to the proxy factory if specified, and used instead of the `--from` for calculating the address. When used in combination with `--query`, the address resulting from sending that signature will be returned.

This PR also adds `--proxy-admin` and `--proxy-factory` flags to `zos push`, allowing to eagerly deploy both contracts. This is required for predicting the deployment address of a proxy, since the address of the proxy factory is involved in the calculation, and the proxy admin is the default owner set during initialization of the proxy.

There is also a new example project included to showcase how to use this feature, which is also included in the CI to be run along other examples.

Last but not least, this PR also includes fixes to the Nomic audit.

Fixes #799 
Fixes #795 
Fixes #731 (via new `--proxy-admin` flag)

_Note to reviewer: checking this PR by-commit may be easier._


packages/cli/src/commands/create2.ts
```diff
@@ -2,48 +2,64 @@ import pickBy from 'lodash.pickby';
...
   .usage('[alias] --network <network> --salt <salt> [options]')
   .description(description)
   .option('--salt <salt>', `salt used to determine the deployment address`)
-  .option('--query', `do not create the contract and just return the deployment address`)
-  .option('--init [function]', `call function after creating contract. If none is given, 'initialize' will be used`)
-  .option('--args <arg1, arg2, ...>', 'provide initialization arguments for your contract if required')
+  .option('--query [sender]', `do not create the contract and just return the deployment address, optionally specifying the sender used to derive the deployment address (defaults to 'from')`)
+  .option('--init [function]', `optional initialization function to call after creating contract (defaults to 'initialize')`)
+  .option('--args <arg1, arg2, ...>', 'optional arguments to the initialization function')
+  .option('--admin <admin>', 'optional admin of the proxy (uses the project\'s proxy admin if not specified)')
```
jcarpanelli: Stating the concept of 'optional' here (and in the `--admin` and `--signature` flags) may cause confusion to users, as they may want to run the command providing the flag but not the value associated. I believe `commander.js` will throw an error here though, but it would be nice to rephrase it.

packages/cli/src/commands/create2.ts
```diff
@@ -2,48 +2,64 @@ import pickBy from 'lodash.pickby';
...
+  .option('--signature <signature>', `optional signature of the request, uses the signer instead of sender to derive the deployment address`)
   .option('--force', 'force creation even if contracts have local modifications')
   .withNetworkOptions()
   .action(action);
 
 async function action(contractFullName: string, options: any): Promise<void> {
   const { network, txParams } = await ConfigVariablesInitializer.initNetworkConfiguration(options);
   if (!await hasToMigrateProject(network)) process.exit(0);
+  if (!options.salt) throw new Error('option `--salt\' is required');
```
jcarpanelli: Using ` and \' is a convention I don't know, or is it a typo?

packages/cli/src/commands/create2.ts
```diff
@@ -2,48 +2,64 @@ import pickBy from 'lodash.pickby';
...
 
-  if (options.query) await runQuery(options, network, txParams);
-  else await runCreate(options, contractFullName, network, txParams);
+  const { initMethod, initArgs } = parseInit(options, 'initialize');
+  const { contract: contractAlias, package: packageName } = fromContractFullName(contractFullName);
+  const opts = { ... options, initMethod, initArgs, contractAlias, packageName };
```
jcarpanelli:
```suggestion
  const opts = { ...options, initMethod, initArgs, contractAlias, packageName };
```

packages/cli/src/commands/create2.ts
```diff
@@ -2,48 +2,64 @@ import pickBy from 'lodash.pickby';
...
-async function runCreate(options: any, contractFullName: string, network: string, txParams: any) {
-  const { force, salt } = options;
-  const { initMethod, initArgs } = parseInit(options, 'initialize');
-  const { contract: contractAlias, package: packageName } = fromContractFullName(contractFullName);
+async function runSignatureQuery(options: any, network: string, txParams: any) {
+  const { query, initMethod, initArgs, contractAlias, packageName, force, salt, signature: signatureOption, admin } = options;
+  if (!contractAlias) throw new Error('missing required argument: alias');
+  if (typeof query === 'string') throw new Error('cannot specify argument `sender\' as it is inferred from `signature\'');
+  const args = pickBy({ packageName, contractAlias, initMethod, initArgs, force, salt, signature: signatureOption, admin });
+  await querySignedDeployment({ ... args, network, txParams });
```
jcarpanelli:
```suggestion
  await querySignedDeployment({ ...args, network, txParams });
```

packages/cli/src/commands/push.ts
```diff
@@ -32,11 +32,13 @@ const register: (program: any) => any = (program) => program
   .option('-d, --deploy-dependencies', 'deploys dependencies to the network if there is no existing deployment')
   .option('--reset', 'redeploys all contracts (not only the ones that changed)')
   .option('-f, --force', 'ignores validation errors and deploys contracts')
+  .option('--proxy-admin', 'eagerly deploys the project\'s proxy admin (if not deployed yet on this network)')
+  .option('--proxy-factory', 'eagerly deploys the project\'s proxy factory (if not deployed yet on this network)')
```
jcarpanelli: WDYT about being a bit more explicit with naming here, and change both flags to `--deploy-proxy-admin` and `--deploy-proxy-factory` ?

I think it will be so much clear!
```sh
zos push --network local --deploy-proxy-admin --deploy-proxy-factory
```
It's a personal opinion though.

packages/lib/src/artifacts/ZWeb3.ts
```diff
@@ -67,6 +67,8 @@ export default class ZWeb3 {
   }
 
   public static toChecksumAddress(address: string): string {
```
jcarpanelli:
```suggestion
  public static toChecksumAddress(address: string): string | null {
```

packages/lib/src/project/AppProject.ts
```diff
@@ -165,53 +166,65 @@ export default class AppProject extends BasePackageProject {
...
       initMethod = 'initialize';
     }
-    return { contractName, packageName, initArgs, initMethod };
+
+    return { ...opts, contractName, packageName, initMethod };
   }
 
-  public async createProxyWithSalt(contract: Contract, salt: string, contractInterface: ContractInterface = {}): Promise<Contract> {
-    const { contractName, packageName, initMethod, initArgs } = this.getContractInterface(contract, contractInterface);
+  public async createProxyWithSalt(contract: Contract, salt: string, signature?: string, contractInterface: ContractInterface = {}): Promise<Contract> {
```
jcarpanelli: I believe this is not how conditional arguments should be used (as a matter of fact, if you don't set a defualt value to `contractIinterface`, tsc will fail) They should be defined always at the end, so you can avoid passing `null` in calls. 
As this may lead to future errors, I would reorder the parameters as: `contract`, `contractInterface`, `salt` and `signature?`. WDYT?

packages/lib/test/src/ProxyFactory.test.js
```diff
@@ -5,72 +5,102 @@ import utils from 'web3-utils';
...
+  before('set implementations', async function() {
+    this.implementationV1 = await ImplV1.new();
+    this.implementationV2 = await ImplV2.new();
+  });
+
+  beforeEach('create factory', async function() {
+    this.factory = await ProxyFactory.deploy({ from: sender });
+  });
+
+  describe('#deploy', function () {
```
jcarpanelli: :heart_eyes:

packages/cli/src/commands/create2.ts
spalladino: It's how commander.js displays errors

packages/cli/src/commands/push.ts
spalladino: Like the idea!

packages/lib/src/project/AppProject.ts
spalladino: I'd agree, but contractInterface is also optional here (since it has a default value), so it's ok to flag signature as optional.

PR #810, opened 2019-04-12
Fix Contract#at to create a clone of the current instance

### LINKED ISSUE #784: TestHelper "createProxy" does not create unique proxies.
This is a quirky issue, and maybe I am overlooking something here:

**Expected Behavior**

When I create multiple proxies in a test using the testhelper's 'createProxy()' function, I expect that each proxy created is unique (deployed at a different address).

**Current Behavior**

When creating multiple proxies of the same contract inside a it() block (or even creating one inside a beforeEach and one inside an it() block) the resulting proxies are identical. 

When following the documentation here: [https://docs.zeppelinos.org/docs/testing.html](https://docs.zeppelinos.org/docs/testing.html)

If I attempt to create two proxies in a single it() block, the second proxy has the same address as the first- they are identical.

**Example:** 
```
  it("should create two proxies", async function() {
    const proxy1 = await this.project.createProxy(Sample);
    const proxy2 = await this.project.createProxy(Sample);

    const add1 = proxy1.address;
    const add2 = proxy2.address;
    console.log(add1, " ", add2);
    add1.should.eq(add2);
  });

```
In this case, this test will pass, and the addresses from the console.log will be identical. 

**My complete test code is:** 

Note the commented out Sample2 variable. If I call the second createProxy with Sample2 as the argument, I do get a unique proxy. However, I am not clear of what the behavior is in this case. I would expect createProxy(Sample) to generate me a unique proxy each time. 

```
const { TestHelper } = require("zos");
const { Contracts, ZWeb3 } = require("zos-lib");

ZWeb3.initialize(web3.currentProvider);

const Sample = Contracts.getFromLocal("Sample");
// const Sample2 = Contracts.getFromLocal("Sample");
const ERC20 = Contracts.getFromNodeModules("openzeppelin-eth", "ERC20");

require("chai").should();

contract("Sample", function() {
  beforeEach(async function() {
    this.project = await TestHelper();
  });

  it("should create two proxies", async function() {
    const proxy1 = await this.project.createProxy(Sample);
    const proxy2 = await this.project.createProxy(Sample);

    const add1 = proxy1.address;
    const add2 = proxy2.address;
    //console.log(proxy1);
    console.log(add1, " ", add2);
    proxy1.should.eq(proxy2);
  });
});
```

spalladino:
Function `Contract#at` was modifying the current instance instead of creating a new one. This caused two invocations to `at` on the same contract with different addresses to end up yielding the same instance.

Fixes #784

### DIFF
--- a/packages/lib/src/artifacts/Contract.ts
+++ b/packages/lib/src/artifacts/Contract.ts
@@ -82,8 +82,10 @@ function _wrapContractInstance(schema: any, instance: Web3Contract): Contract {
 
   instance.at = function(address: string): Contract | never {
     if(!ZWeb3.isAddress(address)) throw new Error('Given address is not valid: ' + address);
-    instance.options.address = instance._address = address;
-    return instance;
+    const newWeb3Instance = instance.clone();
+    newWeb3Instance._address = address;
+    newWeb3Instance.options.address = address;
+    return _wrapContractInstance(instance.schema, newWeb3Instance);
   };
 
   instance.link = function(libraries: { [libAlias: string]: string }): void {
--- a/examples/creating-instances-from-solidity/index.js
+++ b/examples/creating-instances-from-solidity/index.js
@@ -90,8 +90,7 @@ async function main() {
   console.log(`Factory proxy created at ${factoryProxy.address}`);
 
   // Use the Factory proxy address to create a javascript Factory contract object.
-  const factoryContract = Contracts.getFromLocal('Factory');
-  factoryContract.at(factoryProxy.address);
+  const factoryContract = Contracts.getFromLocal('Factory').at(factoryProxy.address);
 
   // Construct the call data for the initialize method of Instance.sol.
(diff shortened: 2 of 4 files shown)


packages/lib/src/artifacts/Contract.ts
```diff
@@ -82,8 +82,9 @@ function _wrapContractInstance(schema: any, instance: Web3Contract): Contract {
 
   instance.at = function(address: string): Contract | never {
     if(!ZWeb3.isAddress(address)) throw new Error('Given address is not valid: ' + address);
-    instance.options.address = instance._address = address;
-    return instance;
+    const newWeb3Instance = instance.clone();
+    newWeb3Instance.options.address = newWeb3Instance._address = address;
```
ylv-io: Can we not use chained assignments? I believe they are error prone because how they are implemented in JS and just in general not easy flow. Still a bit opinionated statement. Would be great to have shared guidelines.

PR #811, opened 2019-04-12
Add test to check for unlinked libraries on deploy

### LINKED ISSUE #250: Check if there are unlinked solidity libraries before uploading a contract
If a contract slips through the CLI with unlinked solidity libraries, the error the user gets is an `out of gas` directly from Ganache, or something even friendlier like `cannot unmarshal invalid hex string into Go struct field CallArgs.data of type hexutil.Bytes` from Geth.

To prevent this, we can check if the bytecode we are about to deploy in `Transactions.js#deploy` has any placeholders left, and throw a friendlier error in that scenario.

spalladino:
Fixes #250

### DIFF
--- a/packages/lib/contracts/mocks/WithLibrary.sol
+++ b/packages/lib/contracts/mocks/WithLibrary.sol
@@ -1,0 +1,13 @@
+pragma solidity ^0.5.0;
+
+library MyLibrary {
+  function answer() external pure returns(uint256) {
+    return 42;
+  }
+}
+
+contract WithLibraryMock {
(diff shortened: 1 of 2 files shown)


packages/lib/contracts/mocks/WithLibrary.sol
```diff
@@ -0,0 +1,13 @@
+pragma solidity ^0.5.0;
+
+library MyLibrary {
+  function answer() external pure returns(uint256) {
+    return 42;
```
ylv-io: love reference

PR #818, opened 2019-04-22
Implement function to look for project root dir

jcarpanelli:
Fixes https://github.com/zeppelinos/zos/issues/646. Also, it could be useful for https://github.com/zeppelinos/zos/pull/792 and running `zos init` anywhere in the project :eyes:.

### DIFF
--- a/packages/cli/src/bin/helpers.ts
+++ b/packages/cli/src/bin/helpers.ts
@@ -1,0 +1,14 @@
+import fs from 'fs';
+
+export default function lookForRootDirectory(dirPath: string): string | null {
+  if (dirPath.length === 0) return null;
+  const dirFiles = fs.readdirSync(dirPath);
+
+  if ((dirFiles.includes('package.json') && dirFiles.includes('package-lock.json')) || dirFiles.includes('zos.json')) {
+    return dirPath;
+  } else {
+    const dir = dirPath.split('/');
+    dir.pop();
+    return lookForRootDirectory(dir.join('/'));
+  }
+}
--- a/packages/cli/src/bin/helpers.ts
+++ b/packages/cli/src/bin/helpers.ts
@@ -4,7 +4,5 @@ import findUp from 'find-up';
 export default function findRootDirectory(cwd: string): string | null {
   const filePath = findUp.sync(['package.json', 'zos.json'], { cwd });
   if (!filePath) return null;
-  const dir = filePath.split(path.sep);
-  dir.pop();
-  return dir.join(path.sep);
+  return path.dirname(filePath);
 }
--- a/packages/cli/src/bin/helpers.ts
+++ b/packages/cli/src/bin/helpers.ts
@@ -1,14 +1,10 @@
-import fs from 'fs';
+import path from 'path';
+import findUp from 'find-up';
 
-export default function lookForRootDirectory(dirPath: string): string | null {
-  if (dirPath.length === 0) return null;
-  const dirFiles = fs.readdirSync(dirPath);
-
-  if ((dirFiles.includes('package.json') && dirFiles.includes('package-lock.json')) || dirFiles.includes('zos.json')) {
-    return dirPath;
-  } else {
-    const dir = dirPath.split('/');
-    dir.pop();
-    return lookForRootDirectory(dir.join('/'));
-  }
+export default function findRootDirectory(cwd: string): string | null {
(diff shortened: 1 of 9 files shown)


packages/cli/src/bin/helpers.ts
```diff
@@ -0,0 +1,14 @@
+import fs from 'fs';
+
+export default function lookForRootDirectory(dirPath: string): string | null {
+  if (dirPath.length === 0) return null;
+  const dirFiles = fs.readdirSync(dirPath);
+
+  if ((dirFiles.includes('package.json') && dirFiles.includes('package-lock.json')) || dirFiles.includes('zos.json')) {
+    return dirPath;
+  } else {
+    const dir = dirPath.split('/');
```
ylv-io: Not every OS use '/' as path separator. User [path.sep](https://nodejs.org/api/path.html#path_path_sep) instead.

jcarpanelli: Good catch!

jcarpanelli:
> On a side note. What is the reason for us writing these helpers when there is a [library](https://www.npmjs.com/package/find-up) with the same functionality and 16kk downloads a week?
> Like `(await findUp(['package.json', 'package-lock.json', 'zos.json']))`?

I didn't know about the existence of that library! Thanks :heart: 
I double-checked it, and it seems that `find-up#findUp` function will return the full path of the first occurrence (e.g., `/home/foo/project/package.json`), so we still have to remove the file name from the string returned. 
I also noticed that it depends on 5 other packages (directly or indirectly, see [the dependency tree](http://npm.broofa.com/?q=find-up)), which could be overkill compared to the function implemented in this PR. Anyways, I will probably replace it later today.

packages/cli/src/bin/zos-cli.ts
```diff
@@ -3,11 +3,15 @@
...
+import lookForRootDirectory from './helpers';
+
+const rootDirectory = lookForRootDirectory(process.cwd()) || process.cwd();
+if (rootDirectory !== null && rootDirectory !== process.cwd()) process.chdir(rootDirectory);
 
 // Acquire file lock to ensure no other instance is running
 const LOCKFILE: string = '.zos.lock';
 try {
-  lockSync(LOCKFILE, { retries: 0 });
+  lockSync(`${rootDirectory}/${LOCKFILE}`, { retries: 0 });
```
spalladino:
```suggestion
  lockSync(LOCKFILE, { retries: 0 });
```
Should not be needed, as we are doing a chdir right before

spalladino:
As discussed offline, let's add an exception when the user is explicitly running `zos init` to ensure is run on the current folder. Otherwise, we make it impossible to create a new project from scratch using `zos init` in a current folder if there is a `package.json` up in the tree (as is, for instance, in a monorepo).

jcarpanelli:
> As discussed offline, let's add an exception when the user is explicitly running `zos init` to ensure is run on the current folder. Otherwise, we make it impossible to create a new project from scratch using `zos init` in a current folder if there is a `package.json` up in the tree (as is, for instance, in a monorepo).

I've just pushed this change together with @ylv-io suggestions. Ready for review!

packages/cli/src/bin/helpers.ts
```diff
@@ -0,0 +1,10 @@
+import path from 'path';
+import findUp from 'find-up';
+
+export default function findRootDirectory(cwd: string): string | null {
+  const filePath = findUp.sync(['package.json', 'zos.json'], { cwd });
+  if (!filePath) return null;
+  const dir = filePath.split(path.sep);
```
ylv-io: My two cents.
There is a method [path.dirname](https://nodejs.org/api/path.html#path_path_dirname_path) so you can do `return path.dirname(filePath);`. Feels more natural.

jcarpanelli:
I've just implemented @ylv-io last suggestion :smile: 

@spalladino I also removed all calls to the `init` command. It will be very easy to add them again later. I think we should further discuss how we can include that feature, as I feel that just running `zos create` anywhere in the project and doing everything under the hood (_init_, add, push) will be the easiest, cleanest and more awesome way of creating a proxy without bothering the user with unnecessary steps.

That being said, we can think how to readd this later and merge this PR now.

spalladino:
Agree, I think we're good to merge. @ylv-io could you update your review?

Issue #798, opened 2019-04-03
New command for creating a zepkit box


spalladino:
Add a new command (`zos create-dapp`, or an option to `zos init` like `--box=react`) so that the `zos` CLI creates a new zepkit box. 

By default, the box should *not* include any of the tutorials, but should have one of the examples (the counter one) already set up, so the user understands how everything is wired.

spalladino:
Fixed in #822

PR #822, opened 2019-04-24
Adds `unpack` command

ylv-io:
The PR introduce `zos unpack` command.
Goal behind the command is to enable easy distribution of various ZOS kits made by Zeppelin or other developers.
All ZOS kits required to have `stable` branch with `kit.json` at the root level.
[ZepKit](https://github.com/zeppelinos/zepkit) is an example of a such kit.
Changes affect zos-cli project:
* `unpack` command introduced
* `unpack` script introduced
* `unpack` controller introduced
* test for `unpack` introduced

Related to #798

---
Unpack


packages/cli/src/models/local/UnpackController.ts
```diff
@@ -0,0 +1,72 @@
...
+      e.message = `Failed to download and unpack kit from ${url}. Details: ${e.message}`;
+      throw e;
+    }
+  }
+
+  public async verifyRepo(url: string): Promise<KitConfig | never> {
+    if (!url) throw Error('A url must be provided.');
+
+    try {
+      return JSON.parse(await request(url
```
spalladino: Let's validate that the JSON file has the expected fields. Also, it'd be good to add some form of versioning to zepkits (similar to what we are doing in the zos.json files), in case we want to change the config format in the future.

packages/cli/src/models/local/UnpackController.ts
```diff
@@ -0,0 +1,72 @@
...
+
+    // because zos always spawns '.zos.lock' file
+    if ((await readdir(workingDirPath)).length > 1) throw Error('The directory must be empty');
+
+    try {
+      let spinner = ora(`Downloading kit from ${url}`).start();
+      const git = simpleGit(workingDirPath);
+      await git.init();
+      await git.addRemote('origin', url);
+      await git.pull('origin', 'stable');
```
spalladino: What do you think about changing the `pull` command for a `fetch` followed by a [sparse checkout](https://stackoverflow.com/a/33934496/12791), using the ignore paths, so we don't need to delete anything afterwards?

packages/cli/src/models/local/UnpackController.ts
```diff
@@ -0,0 +1,72 @@
...
+      spinner.succeed();
+
+      spinner = ora(`Unpacking kit`).start();
+      // always delete .git folder 
+      config.ignore.push(`.git`);
+      // delete all files/folders from ignore
+      const paths = config.ignore.map(file => path.join(workingDirPath, file));
```
spalladino: If we don't use the sparse checkout, let's verify that no paths include a `..`, so no files from outside the working dir are deleted.

spalladino: I know that they could still be deleted with a hook, but well...

packages/cli/src/models/local/UnpackController.ts
```diff
@@ -0,0 +1,72 @@
...
+      const exists = await Promise.all(paths.map(path => pathExists(path)));
+      await Promise.all(exists.map((ext, i)=> ext ? remove(paths[i]) : null ).filter(Boolean));
+      
+      // run kit commands like `npm install`
+      await exec(config.hooks['post-unpack']);
+      spinner.succeed();
+
+      stdout(`Done. Amazing!`);
+      stdout(`Please, continue at ${url}`);
```
spalladino: Shouldn't the kit have the responsibility of showing the next steps? Maybe we could include a "description" or "next steps" field in the `kit.json` that is shown here, or just make it part of the `post-unpack`.

packages/cli/src/models/local/UnpackController.ts
```diff
@@ -0,0 +1,72 @@
...
+      // delete all files/folders from ignore
+      const paths = config.ignore.map(file => path.join(workingDirPath, file));
+      const exists = await Promise.all(paths.map(path => pathExists(path)));
+      await Promise.all(exists.map((ext, i)=> ext ? remove(paths[i]) : null ).filter(Boolean));
+      
+      // run kit commands like `npm install`
+      await exec(config.hooks['post-unpack']);
+      spinner.succeed();
+
+      stdout(`Done. Amazing!`);
```
spalladino:
```suggestion
      stdout(`The kit is ready to use. Amazing!`);
```
If we have access to the name, so we can use it instead of "the kit", much beter!

packages/cli/src/models/local/UnpackController.ts
```diff
@@ -0,0 +1,72 @@
...
+  hooks: Object,
+}
+
+export default class UnpackController {
+  public async unpack(url: string, workingDirPath: string = '', config: KitConfig): Promise<void | never> {
+    if (!url) throw Error('A url must be provided.');
+    if (!config) throw Error('A config must be provided.');
+
+    const { exec } = cp;
+    const { readdir, remove, pathExists } = fs;
```
spalladino: If we remove the need for `remove` (pun not intended), could we also remove (still not intended) `fs-extra` and just use regular `fs`, so we don't introduce another dependency?

packages/cli/src/models/local/UnpackController.ts
spalladino: Alternatively, we could replace the `ignore` field of the `kit.json` by a `files` property, so the kit owner must manually list which files or folders are part of the kit (as we do in both zos and zos-lib `package.json`s). In that case, the `checkout` would be much cleaner, since we can just call checkout with the list of files and folders to unpack.

```
git checkout [<tree-ish>] [--] <pathspec>...
           Overwrite paths in the working tree by replacing with the contents in the index or in
           the <tree-ish> (most often a commit).
```

I think this option is cleaner and safer than the sparse checkout, or the manual remove. WDYT?

packages/cli/src/models/local/UnpackController.ts
```diff
@@ -0,0 +1,72 @@
...
+
+const ora = patch('ora');
+const simpleGit = patch('simple-git/promise');
+
+interface KitConfig {
+  ignore: string[],
+  hooks: Object,
+}
+
+export default class UnpackController {
```
spalladino: I like that you are already starting with breaking apart the controllers. Good initiative!

packages/cli/test/scripts/unpack.test.js
```diff
@@ -0,0 +1,75 @@
...
+import cp from '../../src/utils/cp';
+import patch, { cache } from '../../src/utils/patch';
+import unpack from '../../src/scripts/unpack';
+
+const simpleGit = patch('simple-git/promise');
+
+const repo = 'zeppelinos/zepkit';
+const url = `https://github.com/zeppelinos/zepkit.git`;
+
+describe('unpack script', function() {
```
spalladino: We definitely need an integration test for running this command. Maybe we can tackle it along with the integration tests of the zepkit itself?

packages/cli/package.json
```diff
@@ -73,6 +74,9 @@
     "lodash.uniq": "^4.5.0",
     "lodash.uniqwith": "^4.5.0",
     "npm-programmatic": "0.0.10",
+    "ora": "^3.4.0",
+    "request": "^2.88.0",
```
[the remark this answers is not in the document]
ylv-io: Sure, I just completely missed the fact that we have `axios`.

packages/cli/package.json
```diff
@@ -89,7 +93,7 @@
     "chai-as-promised": "^7.1.1",
     "chai-bignumber": "^2.0.2",
     "chai-string": "^1.5.0",
-    "ganache-cli": "^6.4.0",
+    "ganache-cli": "^6.3.0",
```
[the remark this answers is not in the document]
ylv-io: Because I have tested 6.3.0 to work properly with ZepKit and pinned ZepKit to that version but didn't test 6.4.0. I thought it wouldn't hurt to stay for everyone on the same version.

packages/cli/test/scripts/unpack.test.js
```diff
@@ -0,0 +1,75 @@
+'use strict'
+require('../setup');
+
+import sinon from 'sinon';
+import fs from 'fs-extra';
+import cp from '../../src/utils/cp';
```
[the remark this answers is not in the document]
ylv-io: Okay, even it is common to us `cp` for the module.

packages/cli/src/models/local/UnpackController.ts
```diff
@@ -0,0 +1,72 @@
...
+    if (!config) throw Error('A config must be provided.');
+
+    const { exec } = cp;
+    const { readdir, remove, pathExists } = fs;
+
+    // because zos always spawns '.zos.lock' file
+    if ((await readdir(workingDirPath)).length > 1) throw Error('The directory must be empty');
+
+    try {
+      let spinner = ora(`Downloading kit from ${url}`).start();
```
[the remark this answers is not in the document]
ylv-io: Totally love idea hiding it behind proxy. I'll ask @jcarpanelli if we have a spinner or not. I am pretty certain we need a spinner for ZOS.

packages/cli/src/models/local/UnpackController.ts
ylv-io: Yeah. I've skipped it as a decision to deliver MVP faster but It seems it didn't work!

packages/cli/src/models/local/UnpackController.ts
ylv-io: My first implementation was like that and then I realized it is wrong.
I was thinking about it for quite a bit. Problem with showing next step or description in terminal are:
* Terminal has limited formatting support 
* It is hard to go back to terminal after running more commands and `clear` command would wipe it out
I believe that the way other projects do it is wrong. The print long list of commands which is gone after 10 minutes of working in terminal and there is no way to go back.
That is why I decided to focus to bringing user to the kit readme file which can be opened with rich content and sustained for longer time.
Link in terminal is clickable in Visual Studio Code or can be copied.

packages/cli/src/models/local/UnpackController.ts
ylv-io: I also use `pathExists`. `fs-extra` is helpful library because it provides function which are not present in just `fs` and we'll most likely need them anyway. For example `remove` will remove both dir and file. 
Also it provide async versions of `fs` methods out of the box.
My question is how much we are concerned with dependencies? As my understanding we are not building a website so we are not super sensitive to package size.

packages/cli/test/scripts/unpack.test.js
ylv-io: Yeah. But I wouldn't do it as a part of `cli` because downloading and installing a kit few times will take like 10 minutes. We have to be smart about it. It should be a separate test suite for sure.

packages/cli/src/models/local/UnpackController.ts
```diff
@@ -0,0 +1,72 @@
...
+      
+      // run kit commands like `npm install`
+      await exec(config.hooks['post-unpack']);
+      spinner.succeed();
+
+      stdout(`Done. Amazing!`);
+      stdout(`Please, continue at ${url}`);
+    } catch(e) {
+      // TODO: remove all files from directory on fail except .zos.lock
+      e.message = `Failed to download and unpack kit from ${url}. Details: ${e.message}`;
```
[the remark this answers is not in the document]
ylv-io: It is actually used on ZOS to some extent like:
````
error.message = `${contractAlias} deployment failed with error: ${error.message}`;
throw error;
````
I am actually more concerned about code which spawns new error killing old stracktrace, like this
````
} catch(error) {
    throw new Error(error.message || 'Error while trying to verify contract');
  }
````
I think I'll start `Coding Guide` document rather soon so we all can collaborate and define guidelines for us and future developers to work on project.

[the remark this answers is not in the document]
Comment (spalladino) [reply packages/cli/src/models/local/UnpackController.ts]: > I think I'll start Coding Guide document rather soon so we all can collaborate and define guidelines for us and future developers to work on project.

Sounds good. FWIW, I'd actually start by adopting a guide from an existing project, rather than draft one from scratch.

File: packages/cli/src/models/local/UnpackController.ts (hunk 1 above)
Comment (spalladino) [reply packages/cli/src/models/local/UnpackController.ts]: Hehe I like the idea of pushing the MVP, but versioning is like backups in that you only notice you need them when it's too late.

Or like smart contract upgrades, for that matter.

File: packages/cli/src/models/local/UnpackController.ts (hunk 4 above)
Comment (spalladino) [reply packages/cli/src/models/local/UnpackController.ts]: You make a good point. Nevertheless, I'm not sure they are mutually exclusive: a project may want to include some basic next steps, as well as redirecting to the README. 

My main concern with the redirect is that only a fraction of the users will actually click on the link. As an example, OZ recently got an issue opened from a very good dev, that was solved just by clicking on a link included on the error msg. So, it would be good to include the very basic points right in the terminal.

File: packages/cli/src/models/local/UnpackController.ts (hunk 6 above)
Comment (spalladino) [reply packages/cli/src/models/local/UnpackController.ts]: Size it's definitely not an issue in the CLI (it is on lib, since we'll want to add browser support soon), but I'd still be careful in how much we add. If you think it adds value, then let's go for it - I trust your judgement there!

File: packages/cli/test/scripts/unpack.test.js (hunk 8 above)
Comment (spalladino) [reply packages/cli/test/scripts/unpack.test.js]: Check out the `test` folder at the root of the project, there is already an integration test suite. We could add a second one there.

File: packages/cli/src/models/local/UnpackController.ts (hunk 12 above)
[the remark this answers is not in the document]
Comment (jcarpanelli) [reply packages/cli/src/models/local/UnpackController.ts]: We are already discussing this offline with @ylv-io . I believe that it's not bad to use `ora` in these cases for now, but we have to take into account that it has some limitations (e.g., you cannot run multiple spinners at the same time, and that's something that we definitely will need when `zos push`ing several contracts) and that we might remove it in the near future.
That being said, this introduces some sort of inconsistency with the actual logging throughout the tool, but I'm not uncomfortable with it.

File: packages/cli/src/models/local/UnpackController.ts [hunk 14]
```diff
@@ -0,0 +1,72 @@
...
+      // run kit commands like `npm install`
+      await exec(config.hooks['post-unpack']);
+      spinner.succeed();
+
+      stdout(`The kit is ready to use. Amazing!`);
+      stdout(`Please, continue at ${url}`);
+    } catch(e) {
+      // TODO: remove all files from directory on fail except .zos.lock
+      e.message = `Failed to download and unpack kit from ${url}. Details: ${e.message}`;
+      throw e;
```
jcarpanelli: What would happen with `ora` if something goes wrong inside the `try` statement? 
I would add a `spinner.fail()` inside the `catch` statement. Otherwise, the spinner animation will 

Also, could you please check if `ora` automatically adds a new line after failing (i.e, after calling `spinner.fail()`? At least if we don't fail, no new line will be added and the console will log like this:

log

packages/cli/src/scripts/unpack.ts
```diff
@@ -0,0 +1,26 @@
+import process from 'process';
+
+import UnpackController from '../models/local/UnpackController';
```
jcarpanelli: I love that you are implementing this in a new controller! We will slowly reduce (well, at least not making bigger) the `NetworkController` and `LocalController` files :+1: 
That being said I think we can be more descriptive with the controller's name. Using a _noun_  instead of using the verb `unpack` might be better. Maybe `PackController`? (considering pack is a noun in this context and I'm not using it as a verb)

packages/cli/test/scripts/unpack.test.js
```diff
@@ -0,0 +1,75 @@
...
+import patch, { cache } from '../../src/utils/patch';
+import unpack from '../../src/scripts/unpack';
+
+const simpleGit = patch('simple-git/promise');
+
+const repo = 'zeppelinos/zepkit';
+const url = `https://github.com/zeppelinos/zepkit.git`;
```
jcarpanelli:
```suggestion
const url = 'https://github.com/zeppelinos/zepkit.git';
```

packages/cli/src/commands/unpack.ts
```diff
@@ -0,0 +1,17 @@
+import unpack from '../scripts/unpack';
+
+const name: string = 'unpack';
+const signature: string = `${name} [kit]`;
+const description: string = `Download and install a ZeppelinOS kit to the current directory`;
```
jcarpanelli:
```suggestion
const description: string = `download and install a ZeppelinOS kit to the current directory`;
```
We're starting all command's descriptions with small letters, let's try to keep that consistent.

packages/cli/src/models/local/UnpackController.ts
```diff
@@ -0,0 +1,72 @@
...
+
+    try {
+      let spinner = ora(`Downloading kit from ${url}`).start();
+      const git = simpleGit(workingDirPath);
+      await git.init();
+      await git.addRemote('origin', url);
+      await git.pull('origin', 'stable');
+      spinner.succeed();
+
+      spinner = ora(`Unpacking kit`).start();
```
jcarpanelli:
```suggestion
      spinner = ora('Unpacking kit').start();
```

packages/cli/src/models/local/UnpackController.ts
```diff
@@ -0,0 +1,72 @@
...
+      // always delete .git folder 
+      config.ignore.push(`.git`);
```
jcarpanelli:
```suggestion
      config.ignore.push('.git');
```

packages/cli/src/models/local/UnpackController.ts
```diff
@@ -0,0 +1,72 @@
...
+      // delete all files/folders from ignore
+      const paths = config.ignore.map(file => path.join(workingDirPath, file));
+      const exists = await Promise.all(paths.map(path => pathExists(path)));
+      await Promise.all(exists.map((ext, i)=> ext ? remove(paths[i]) : null ).filter(Boolean));
```
jcarpanelli:
```suggestion
      await Promise.all(exists.map((ext, i) => ext ? remove(paths[i]) : null).filter(Boolean));
```

packages/cli/src/models/local/UnpackController.ts
```diff
@@ -0,0 +1,72 @@
...
+      
+      // run kit commands like `npm install`
+      await exec(config.hooks['post-unpack']);
+      spinner.succeed();
+
+      stdout(`The kit is ready to use. Amazing!`);
```
jcarpanelli:
```suggestion
      stdout('The kit is ready to use. Amazing!');
```

packages/cli/src/models/local/UnpackController.ts
```diff
@@ -0,0 +1,72 @@
...
+      throw e;
+    }
+  }
+
+  public async verifyRepo(url: string): Promise<KitConfig | never> {
+    if (!url) throw Error('A url must be provided.');
+
+    try {
+      return JSON.parse(await request(url
+        .replace(".git", "/stable/kit.json")
```
jcarpanelli:
```suggestion
        .replace('.git', '/stable/kit.json')
```

It's somewhat bizarre that the githook didn't warn you before commiting this. Are you skipping it?

packages/cli/src/models/local/UnpackController.ts
```diff
@@ -0,0 +1,72 @@
...
+        .replace("github.com", "raw.githubusercontent.com")
```
jcarpanelli:
```suggestion
        .replace('github.com', 'raw.githubusercontent.com')
```

packages/cli/src/models/local/UnpackController.ts
```diff
@@ -0,0 +1,72 @@
...
+      spinner.succeed();
+
+      spinner = ora(`Unpacking kit`).start();
+      // always delete .git folder 
+      config.ignore.push(`.git`);
+      // delete all files/folders from ignore
+      const paths = config.ignore.map(file => path.join(workingDirPath, file));
+      const exists = await Promise.all(paths.map(path => pathExists(path)));
+      await Promise.all(exists.map((ext, i)=> ext ? remove(paths[i]) : null ).filter(Boolean));
+
```
jcarpanelli:
```suggestion
```

packages/cli/src/models/local/UnpackController.ts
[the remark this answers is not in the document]
ylv-io: Yes, I was thinking the same? Do you have project in mind?

packages/cli/src/models/local/UnpackController.ts
ylv-io: Was thinking about this for quite a while. I suggest yet another option. I want to reduce entrance effort for kit developers to minimum. What is why I don't think it is that good to ask them to either create a list of include/exclude files because it is time consuming and error prone.
Also, all extra files still have to be downloaded. I am thinking, since we have `stable` branch only for releases, what if we ask kit's developers to remove all extra files from that branch? Essentially implementing it on a git/branch level? That way files can't be offsync and don't need to be downloaded?
WDYT?

packages/cli/src/models/local/UnpackController.ts
ylv-io: I agree we need both. It totally makes sense for to provide kit's developers a way to show message after unpacking. I would say we should both with heavy and rich guide at github and short and to the point instructions at terminal.

packages/cli/src/models/local/UnpackController.ts
ylv-io: Am I missing something or ZosPackageFile doesn't really validate `zos.json` except zosversion? I mean It doesn't check if any other fields are present or some extra fields not present? Do we have a npm package for validating JSON? If not I'll use https://www.npmjs.com/package/ajv

packages/cli/src/models/local/UnpackController.ts
spalladino: If I understand correctly, it'd something like a `gh-pages` branch, but specific for the zepkit publishing, right? If we go that way, then I would not use the `stable` branch: it already has semantic in the context of git-flow, and it should not ignore files that are tracked under `master` or `develop`. We should ask them to create a dedicated branch, such as `zos-kit`, but that could add more friction than the whitelist.

How about making the `files` property in the `kit.json` optional? If it is not present, then everything is included; otherwise, it is passed as an argument to `checkout` so that only the chosen files are downloaded. I don't think it's much of a hassle if all files are downloaded by default.

ylv-io: Love it. No downloads and zero friction.

packages/cli/src/models/local/KitController.ts
```diff
@@ -0,0 +1,116 @@
...
+export default class KitController {
+  public async unpack(url: string, workingDirPath: string = '', config: KitConfig): Promise<void | never> {
+    if (!url) throw Error('A url must be provided.');
+    if (!config) throw Error('A config must be provided.');
+
+    const { exec } = child;
+    const { readdir, remove, pathExists } = fs;
+
+    // because zos always spawns '.zos.lock' file
+    if ((await readdir(workingDirPath)).length > 1) throw Error('The directory must be empty');
```
spalladino: I'd suggest checking that the content is indeed `zos.lock`. If the user specifies a different working dir, then it should be absolutely empty, since `zos.lock` should not be there.

packages/cli/src/models/local/KitController.ts
```diff
@@ -0,0 +1,116 @@
...
+        await git.pull('origin', 'stable');
+      } else {
+        // if there are some files then do tree-ish checkout
+        // http://nicolasgallagher.com/git-checkout-specific-files-from-another-branch/
+        await git.checkout([
+          `origin/stable`,
+          `--`,
+          ...config.files,
+        ]);
+      }
```
spalladino: Looks *much* cleaner now, awesome work!!

packages/cli/src/models/local/KitController.ts
```diff
@@ -0,0 +1,116 @@
...
+
+    try {
+      const config = (await axios.get(url
+        .replace('.git', '/stable/kit.json')
+        .replace('github.com', 'raw.githubusercontent.com')
+      )).data;
+
+      // validate our json config
+      // TODO: derive the schema directly from the KitConfig type
+      const ajv = new Ajv({ allErrors: true });
```
spalladino: I wasn't familiar with `ajv`, very cool! We should start using more throughout the app.

packages/cli/src/models/local/KitController.ts
```diff
@@ -0,0 +1,116 @@
...
+  hooks: object;
+}
+
+/* tslint:disable */
+// schema has to have doulbe quotes
+const kitConfigSchema = {
+  "type": "object",
+  "$schema": "http://json-schema.org/draft-07/schema#",
+  "properties": {
+    "version": { "type": "string" },
```
spalladino: Let's name the field differently (we use `zosversion` in `zos.json` files for instance, `manifestVersion` would also work), so we avoid confusion with a potential version of the kit itself.

PR #823, opened 2019-04-29
Transfer command

jcarpanelli:
Fixes #823 

I've started implementing a command in the cli to transfer an amount of ether/wei/gwei between accounts.
For this, I've created a `TransactionController` in the `cli` package that will probably host almost all functions that will interact with the `Transactions` lib's object.

A demo of how the command works:

transfer

### DIFF
--- a/packages/cli/src/commands/transfer.ts
+++ b/packages/cli/src/commands/transfer.ts
@@ -1,38 +1,41 @@
 import { ZWeb3 } from 'zos-lib';
 import transfer from '../scripts/transfer';
 import { promptIfNeeded, networksList, InquirerQuestions } from '../prompts/prompt';
+import { isValidUnit } from '../utils/units';
 import ConfigVariablesInitializer from '../models/initializer/ConfigVariablesInitializer';
 
 const name: string = 'transfer';
 const signature: string = name;
-const description: string = 'send funds in ether to the specified address';
+const description: string = 'send funds in ether, wei or gwei to a given address';
 
 const register: (program: any) => any = (program) => program
   .command(signature, undefined, { noHelp: true })
   .usage('[options]')
   .description(description)
-  .option('--units [units]', 'specify units.')
-  .option('--to <to>', 'to')
-  .option('--value <value>', 'value')
+  .option('--to <to>', 'specify recipient address')
+  .option('--value <value>', 'the amount of ether units to be transferred')
+  .option('--unit <unit>', 'units name. To see supported units, go to: https://web3js.readthedocs.io/en/1.0/web3-utils.html#id85')
   .withNetworkOptions()
   .withNonInteractiveOption()
   .action(action);
 
 async function action(options: any): Promise<void> {
-  const { network: networkInOpts, units, to, value, from, interactive } = options;
-  const promptedNetwork = await promptIfNeeded({ opts: { network: networkInOpts }, props: getCommandProps() }, interactive);
-  const { txParams } = await ConfigVariablesInitializer.initNetworkConfiguration(promptedNetwork, true);
-  const accounts = await ZWeb3.accounts();
-  const opts = { units, to, from, value };
-  console.log(opts);
-  const prompted = await promptIfNeeded({ opts, props: getCommandProps(accounts) }, interactive);
+  const { network: networkInOpts, unit, to, value, from, interactive } = options;
+  const configOpts = { network: networkInOpts, from };
+  const configProps = getCommandProps();
+  const promptedConfig = await promptIfNeeded({ opts: configOpts, props: configProps }, interactive);
+  const { txParams } = await ConfigVariablesInitializer.initNetworkConfiguration(promptedConfig, true);
 
-  await transfer({ ...txParams, ...prompted });
+  const transferOpts = { to, from, value };
+  const transferProps = getCommandProps(await ZWeb3.accounts(), unit);
+  const promptedTransfer = await promptIfNeeded({ opts: transferOpts, props: transferProps }, interactive);
+
+  await transfer({ ...promptedTransfer, unit, txParams });
 
   if (!options.dontExitProcess && process.env.NODE_ENV !== 'test') process.exit(0);
 }
 
-function getCommandProps(accounts: string[] = []): InquirerQuestions {
+function getCommandProps(accounts: string[] = [], unit: string = 'ether'): InquirerQuestions {
   return {
     ...networksList('network', 'list'),
     from: {
@@ -43,18 +46,17 @@ function getCommandProps(accounts: string[] = []): InquirerQuestions {
         value: account
       }))
     },
-    units: {
-      type: 'list',
-      message: 'Select a unit',
-      choices: () => ['Gwei', 'Wei', 'Ether']
-    },
     to: {
       type: 'input',
       message: 'Enter a receiver account'
     },
     value: {
       type: 'input',
-      message: 'Enter a value in wei'
+      message: 'Enter an amount to transfer',
+      transformer: (value) => {
+        if (value.length === 0 || !isValidUnit(unit)) return value;
+        return `${value} ${unit.toLowerCase()}`;
+      },
     }
   };
 }
--- a/packages/cli/src/commands/transfer.ts
+++ b/packages/cli/src/commands/transfer.ts
@@ -6,7 +6,7 @@ import ConfigVariablesInitializer from '../models/initializer/ConfigVariablesIni
 
 const name: string = 'transfer';
 const signature: string = name;
-const description: string = 'send funds in ether, wei or gwei to a given address';
+const description: string = 'send funds to a given address';
 
 const register: (program: any) => any = (program) => program
   .command(signature, undefined, { noHelp: true })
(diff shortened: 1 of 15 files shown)


packages/cli/src/scripts/transfer.ts
```diff
@@ -0,0 +1,8 @@
+import TransactionController from '../models/network/TransactionController';
+import ScriptError from '../models/errors/ScriptError';
+import { CreateParams } from './interfaces';
+
+export default async function createProxy({ units, to, value, txParams = {} }: any): Promise<any> {
```
spalladino:
```suggestion
export default async function transfer({ units, to, value, txParams = {} }: any): Promise<any> {
```

packages/cli/src/models/network/TransactionController.ts
```diff
@@ -0,0 +1,18 @@
...
+  public txParams: any;
+  constructor(txParams) {
+    this.txParams = txParams;
+  }
+
+  public async transfer(units: string, to: string, value: string): Promise<void> {
+    log.info(`Sending ${value} wei to ${to}...`);
+    const receipt = await Transactions.sendRawTransaction(to, { value }, this.txParams);
+    log.info(`Funds successfully sent!. Receipt:`);
+    stdout(receipt);
```
spalladino: Let's output just the tx hash instead of the full receipt

packages/cli/src/commands/transfer.ts
```diff
@@ -0,0 +1,61 @@
...
+
+const name: string = 'transfer';
+const signature: string = name;
+const description: string = 'send funds in ether to the specified address';
+
+const register: (program: any) => any = (program) => program
+  .command(signature, undefined, { noHelp: true })
+  .usage('[options]')
+  .description(description)
+  .option('--units [units]', 'specify units.')
```
spalladino:
```suggestion
  .option('--units <units>', 'specify units.')
```

I'd also consider renaming to `unit` instead of `units`, and indicate which values are valid ('eth', 'wei', etc).

packages/cli/src/commands/transfer.ts
```diff
@@ -0,0 +1,61 @@
...
+    ...networksList('network', 'list'),
+    from: {
+      type: 'list',
+      message: 'Choose the account to send transactions from',
+      choices: accounts.map((account, index) => ({
+        name: `(${index}) ${account}`,
+        value: account
+      }))
+    },
+    units: {
```
spalladino: WDYT about removing this question and just defaulting to ETH? The user can still use a different unit if using the option explicitly.

packages/cli/src/commands/transfer.ts
```diff
@@ -0,0 +1,61 @@
...
+  .withNonInteractiveOption()
+  .action(action);
+
+async function action(options: any): Promise<void> {
+  const { network: networkInOpts, units, to, value, from, interactive } = options;
+  const promptedNetwork = await promptIfNeeded({ opts: { network: networkInOpts }, props: getCommandProps() }, interactive);
+  const { txParams } = await ConfigVariablesInitializer.initNetworkConfiguration(promptedNetwork, true);
+  const accounts = await ZWeb3.accounts();
+  const opts = { units, to, from, value };
+  console.log(opts);
```
spalladino:
```suggestion
```

packages/cli/src/commands/transfer.ts
jcarpanelli: I agree. This way we can also let the user choose between any valid unit.

jcarpanelli:
Ready for review!
I refactored and finished implementing the command. I also removed the `unit` type prompt question. Now, if the user do not specify any unit type, `ether` will be the default, and if the user specifies an non-existent unit, an error will be thrown. Note that we're allowing the user to specify any unit that is [here](https://web3js.readthedocs.io/en/1.0/web3-utils.html#id74). 
Updated demo:

image

packages/cli/src/commands/transfer.ts
```diff
@@ -0,0 +1,63 @@
...
+const signature: string = name;
+const description: string = 'send funds to a given address';
+
+const register: (program: any) => any = (program) => program
+  .command(signature, undefined, { noHelp: true })
+  .usage('[options]')
+  .description(description)
+  .option('--to <to>', 'specify recipient address')
+  .option('--value <value>', 'the amount of ether units to be transferred')
+  .option('--unit <unit>', 'unit name. If none is given, \'ether\' will be used. To see supported units, go to: https://web3js.readthedocs.io/en/1.0/web3-utils.html#id85')
```
spalladino: Let's avoid tying ourselves to web3 documentation. I'd just list here the most common ones (wei, kwei, gwei, milli, ether).

packages/cli/src/commands/transfer.ts
```diff
@@ -0,0 +1,63 @@
...
+        value: account
+      }))
+    },
+    to: {
+      type: 'input',
+      message: 'Enter a receiver account'
+    },
+    value: {
+      type: 'input',
+      message: 'Enter an amount to transfer',
```
spalladino: Can we include the `unit` as part of the message, so we remind the user right here?

packages/cli/src/models/network/TransactionController.ts
```diff
@@ -0,0 +1,23 @@
...
+export default class TransactionController {
+  public txParams: any;
+
+  constructor(txParams) {
+    this.txParams = txParams;
+  }
+
+  public async transfer(to: string, amount: string, unit: string): Promise<void> {
+    if (!isValidUnit(unit)) {
+      throw Error(`Invalid specified unit ${unit}. To see supported units, visit: https://web3js.readthedocs.io/en/1.0/web3-utils.html#id74`);
```
spalladino: Same here, let's just list the most common ones

packages/cli/test/commands/share.js
```diff
@@ -85,33 +87,7 @@ exports.stubCommands = function () {
...
-    this.update.restore()
-    this.verify.restore()
-    this.setAdmin.restore()
-    this.errorHandler.restore()
-    this.initializer.restore()
-    this.compiler.restore()
-    this.getZosversion.restore()
-    this.packageFile.restore()
-    this.dependency.restore()
+    sinon.restore()
```
spalladino: :clap:

packages/cli/test/scripts/transfer.test.js
```diff
@@ -0,0 +1,51 @@
...
+
+    context('when no amount value is specified', function() {
+      it('throws an error', async function() {
+        await transfer({ to: receiver, unit: 'ether' }).should.be.rejectedWith('An amount to be transferred must be specified');
+      });
+    });
+
+    context('when specifying an invalid unit', function() {
+      it('throws an error', async function() {
+        await transfer({ to: receiver, value: '10', unit: 'palla' }).should.be.rejectedWith(/Invalid specified unit palla/);
```
spalladino: :smile:

packages/cli/src/prompts/prompt.ts
```diff
@@ -24,6 +24,7 @@ interface InquirerQuestion {
   default?: any;
   choices?: ChoicesT;
   when?: ((answers: { [key: string]: any }) => boolean);
+  transformer?: ((value: string, answers: { [key: string]: any }) => string);
```
spalladino: What's the difference between this one and normalize?

jcarpanelli: `normalize` is a custom function defined by us to specify the format of the answer, while `transformer` is an `inquirer.js` function that  returns a transformed value to display in the console. The latter only impacts what is shown while editing and it doesn't modify the answer. You can check the gif above, and see how the `ether`, `gwei`, etc is shown while typing an amount value.

packages/cli/src/commands/transfer.ts
jcarpanelli: We will be reminding them the unit while they type the amount value (with the `transformer` function)

packages/cli/test/commands/share.js
ylv-io: The moment you feel stupid!

packages/cli/test/scripts/transfer.test.js
```diff
@@ -0,0 +1,51 @@
...
+    });
+
+    context('when sending a valid amount of ether', function() {
+      it('transfers funds', async function() {
+        (await ZWeb3.getBalance(sender)).should.eq(100e18.toString());
+        (await ZWeb3.getBalance(receiver)).should.eq(100e18.toString());
+
+        await transfer({ from: sender, to: receiver, value: '10', unit: 'ether' });
+
+        (await ZWeb3.getBalance(sender)).should.not.eq(110e18.toString());
```
ylv-io: I assume you are not using `90e18` because of gas cost? I think we should either calculate gas cost or use epsilon if it is not possible, like `|90e18 - value| < epsilon`. WDYT?

spalladino:
> Why `ZWeb3.getBalance` returns `string` when `eth.getBalance` returns `BigNumber`?

Note that, in web3 beta 37, getBalance returns a string and not a BN. However, I agree that we should move into BN - the question is which one. Given the last contributions by george on ethers, I think it's time to revisit that.

packages/cli/test/scripts/transfer.test.js
jcarpanelli: Indeed, that's the reason!  I also noticed I'm asserting against `110e18` instead of `100e18` because of your review! I think we are fine leaving it as `100e18` for now.

PR #834, opened 2019-05-02
Balance command

### LINKED ISSUE #828: Command for querying balance of an address
```
$ zos balance $ADDR
> 0.12 ETH
```

### LINKED ISSUE #829: Command for querying balance in tokens of an address
Related to #828 and #825 

```
$ zos balance $ADDR --erc20 $TOKENADDR/$TOKENNAME
> 100 ZEP
```

jcarpanelli:
Fixes #828, fixes #829.

Started implementing both interactive command and script to get the balance of an account on a specified network, or to get the balance in tokens if an ERC20 address is specified.

Example usage:
balance-gif

Still have to add some tests, and also decide what to do with the ERC20 `abi` that I included in a new `utils/constants.ts`, which I believe it's mandatory to have in the project to query both token balance and symbol.

~_Message for the reviewer: I branched from `feature/transfer-command` and squashed all the commits implemented on it for making the review easier. You should review all commits after `f3f0b76` (included). After merging https://github.com/zeppelinos/zos/pull/823, we can remove that squashed commit_.~

### DIFF
--- a/packages/cli/src/commands/balance.ts
+++ b/packages/cli/src/commands/balance.ts
@@ -1,0 +1,40 @@
+import balance from '../scripts/balance';
+import { promptIfNeeded, networksList, InquirerQuestions } from '../prompts/prompt';
+import ConfigVariablesInitializer from '../models/initializer/ConfigVariablesInitializer';
+
+const name: string = 'balance';
+const signature: string = `${name} [address]`;
+const description: string = 'query the balance of the specified account';
+
+const register: (program: any) => any = (program) => program
+  .command(signature, undefined, { noHelp: true })
+  .usage('[options]')
+  .description(description)
+  .option('--erc20 <contractAddress>', 'specify an ERC20 token address')
+  .withNetworkOptions()
+  .withNonInteractiveOption()
+  .action(action);
+
+async function action(accountAddress: string, options: any): Promise<void> {
+  const { network: networkInOpts, erc20: contractAddress, interactive } = options;
+  const args = { accountAddress };
+  const opts = { network: networkInOpts };
+  const props = getCommandProps();
+  const promptedConfig = await promptIfNeeded({ args, opts, props }, interactive);
+
+  await ConfigVariablesInitializer.initNetworkConfiguration(promptedConfig);
+  await balance({ accountAddress: promptedConfig.accountAddress, contractAddress });
+
+  if (!options.dontExitProcess && process.env.NODE_ENV !== 'test') process.exit(0);
+}
+
+function getCommandProps(): InquirerQuestions {
+  return {
+    ...networksList('network', 'list'),
+    accountAddress: {
+      type: 'input',
+      message: 'Enter an account address',
+    },
+  };
+}
+export default { name, signature, description, register, action };
--- a/packages/cli/src/commands/balance.ts
+++ b/packages/cli/src/commands/balance.ts
@@ -11,7 +11,7 @@ const register: (program: any) => any = (program) => program
   .command(signature, undefined, { noHelp: true })
   .usage('[options]')
   .description(description)
-  .option('--erc20 <contractAddress>', 'specify an ERC20 token address')
+  .option('--erc20 <contractAddress>', 'specify an ERC20 token address or leave empty to query ETH balance')
   .withNetworkOptions()
   .withNonInteractiveOption()
   .action(action);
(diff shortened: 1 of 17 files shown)


packages/lib/src/artifacts/ZWeb3.ts
```diff
@@ -50,6 +50,10 @@ export default class ZWeb3 {
     return Web3.utils.toWei(value, unit);
   }
 
+  public static fromWei(value: string, unit: any): string {
```
spalladino: I'd consider removing this and other utils functions from here. They are in zweb3 only because they are implemented by web3, but they are actually unrelated to the connection to the network. We can tackle this later, of course.

packages/cli/src/commands/balance.ts
```diff
@@ -0,0 +1,40 @@
...
+
+const name: string = 'balance';
+const signature: string = `${name} [address]`;
+const description: string = 'query the balance of the specified account';
+
+const register: (program: any) => any = (program) => program
+  .command(signature, undefined, { noHelp: true })
+  .usage('[options]')
+  .description(description)
+  .option('--erc20 <contractAddress>', 'specify an ERC20 token address')
```
spalladino: I'd add something along the lines of "or leave empty to query ETH balance"

jcarpanelli:
Ready for another review :racehorse:. I implemented the suggestions above, added tests for `units.ts`, and for the balance command and script.
I also used `bignumber.js` (which is already installed in the `cli` package) to do some simple operations, but as we have been discussing, we should consider moving to `BN.js`, which is used by `web3@1.0.0`.

packages/cli/src/commands/balance.ts
```diff
@@ -0,0 +1,43 @@
...
+
+const name: string = 'balance';
+const signature: string = `${name} [address]`;
+const description: string = 'query the balance of the specified account';
+
+const register: (program: any) => any = (program) => program
+  .command(signature, undefined, { noHelp: true })
+  .usage('[options]')
+  .description(description)
+  .option('--erc20 <contractAddress>', 'specify an ERC20 token address or leave empty to query ETH balance')
```
spalladino:
```suggestion
  .option('--erc20 <contractAddress>', 'query the balance of an ERC20 token instead of ETH')
```
"Leave empty" seems to indicate that the value of the flag should be left empty, not that the user should not set the flag at all.

packages/cli/src/commands/balance.ts
```diff
@@ -0,0 +1,43 @@
...
+
+  if (!options.dontExitProcess && process.env.NODE_ENV !== 'test') process.exit(0);
+}
+
+function getCommandProps(): InquirerQuestions {
+  return {
+    ...networksList('network', 'list'),
+    accountAddress: {
+      type: 'input',
+      message: 'Enter an account address',
```
spalladino:
```suggestion
      message: 'Enter an address to query its balance',
```

packages/cli/src/models/network/TransactionController.ts
```diff
@@ -1,23 +1,57 @@
...
+    }
+  }
+
+  private async getTokenInfo(accountAddress: string, contractAddress: string): Promise<ERC20TokenInfo | never> {
+    let balance, tokenSymbol, tokenDecimals;
+    try {
+      const contract = ZWeb3.contract(ERC20_PARTIAL_ABI, contractAddress);
+      balance = await contract.methods.balanceOf(accountAddress).call();
+      tokenSymbol = await contract.methods.symbol().call();
+      tokenDecimals = await contract.methods.decimals().call();
```
spalladino: A `Promise.all` to query all simultaneously is usually better in these scenarios, as it cuts down the latency time to 1/3rd.

packages/cli/src/utils/units.ts
```diff
@@ -1,5 +1,22 @@
...
+import web3Utils from 'web3-utils';
 
 export function isValidUnit(unit: string): boolean {
-  return ZWeb3.getUnits().includes(unit.toLowerCase());
+  return Object.keys(web3Utils.unitMap).includes(unit.toLowerCase());
+}
+
+export function prettifyTokenAmount(amount: string, decimals?: string, symbol?: string): string {
+  const prettifiedAmount = decimals
+    ? new BN(amount).dividedBy(new BN(10).exponentiatedBy(decimals)).toFormat()
```
spalladino:
```suggestion
    ? new BN(amount).shiftedBy(-decimals).toFormat()
```
See http://mikemcl.github.io/bignumber.js/#shift

PR #838, opened 2019-05-06
Remove unused solidity coverage

spalladino:
It was causing a critical security vulnerability upon npm audit.


spalladino:
We need to merge this back into master later.

PR #839, opened 2019-05-06
Interactivity for `set-admin` command

ylv-io:
The PR introduce interactivity support for `set-admin` command.

Flow =>
* Select a network from the network list
* For which proxies would you like to transfer ownership?
  * All proxies
  * Choose by name
  * Choose by address
* Enter an address of a new upgradeability admin
* Please double check your address and type the last 4 characters of the new admin address. If you provide a wrong address, you will lose control over your contracts.


spalladino:
Looking good so far! It'd be good for @jcarpanelli to review once ready, and also to try and abstract somewhere the shared code between set-admin and update.

packages/cli/src/commands/set-admin.ts
```diff
@@ -1,11 +1,35 @@
...
+  networksList,
+  promptForNetwork,
+  proxiesList,
+  proxyInfo,
+  InquirerQuestions
+} from '../prompts/prompt';
 import { hasToMigrateProject } from '../prompts/migrations';
 import ConfigVariablesInitializer from '../models/initializer/ConfigVariablesInitializer';
 
+interface PropsParams {
```
jcarpanelli: Given that now we have some very similar interfaces shared between the `update` and the `set-admin` command, it might be a good idea to move them to a shared file `interfaces.ts` inside `src/commands` and use them the same way we do with the scripts interfaces. WDYT?

packages/cli/src/commands/set-admin.ts
```diff
@@ -19,46 +43,111 @@ const register: (program: any) => any = (program) => program
   .description(description)
   .option('-f, --force', 'bypass a manual check')
   .withNetworkOptions()
+  .withNonInteractiveOption()
```
jcarpanelli: :sunglasses:

packages/cli/src/commands/set-admin.ts
```diff
@@ -19,46 +43,111 @@ const register: (program: any) => any = (program) => program
...
   .action(action);
 
-async function action(contractFullNameOrAddress: string, newAdmin: string, options: any): Promise<void | never> {
-  const { network, txParams } = await ConfigVariablesInitializer.initNetworkConfiguration(options);
+async function action(contractReference: string, newAdmin: string, options: any): Promise<void | never> {
+  const { force, network: networkInOpts, interactive  } = options;
```
jcarpanelli:
```suggestion
  const { force, network: networkInOpts, interactive } = options;
```

packages/cli/src/commands/set-admin.ts
```diff
@@ -19,46 +43,111 @@ const register: (program: any) => any = (program) => program
...
-  // we assume if newAdmin is empty it was specified as first argument
-  if (!newAdmin) {
-    newAdmin = contractFullNameOrAddress;
-    contractFullNameOrAddress = '';
-  }
-
-  if (!options.force) {
+  // has to be a standalone question from interactivity
+  // because it is security related and can't be disabled with interactivity set to false
+  if (!force) {
```
jcarpanelli: I don't agree with forcing interactivity when the user uses the `--no-interactive` flag, even if it's a security validation. Take into account that there might be users that may want to use these commands as we do in our integration tests, and prompting a question will stall those tests.
Let's prompt the `force` flag only if the `--no-interactive` flag is not specified, and check its value later in the `set-admin` script file. (the force flag could be setted in the `Please double check your address and type the last 4 characters...` question)

packages/cli/src/commands/set-admin.ts
```diff
@@ -19,46 +43,111 @@ const register: (program: any) => any = (program) => program
...
     const answers = await inquirer.prompt([
```
jcarpanelli: Same as above. Let's try only to prompt the user when they don't specify the `--no-interactive` flag. That being saind, I believe we can implement this question as all the other ones (i.e., setting the question's properties in the `setCommandProps()` function).

packages/cli/src/commands/set-admin.ts
```diff
@@ -19,46 +43,111 @@ const register: (program: any) => any = (program) => program
...
       {
         name: 'address',
         type: 'string',
         message: 'Please double check your address and type the last 4 characters of the new admin address. If you provide a wrong address, you will lose control over your contracts.',
```
jcarpanelli: I find this message (and how and when it's prompted) somewhat confusing. Maybe just shortening it to a `Please, provide the last 4 characters of the specified address:` or something like that.

packages/cli/src/commands/set-admin.ts
```diff
@@ -19,46 +43,111 @@ const register: (program: any) => any = (program) => program
...
       }
     ]);
-    if (answers.address !== newAdmin.slice(-4)) {
+    if (answers.address.toLowerCase() !== pickedNewAdmin.slice(-4).toLowerCase()) {
```
jcarpanelli:
```suggestion
    if (pickedNewAdmin && answers.address.toLowerCase() !== pickedNewAdmin.slice(-4).toLowerCase()) {
```
I believe that if the user do not specify the new admin address, and is using the `--no-interactive` flag, `zos` will error something like: `cannot read property slice of undefined`, as `newAdmin` will be `undefined`. You will get rid of this error if we only prompt this question when not specifying the `--no-interactive` though.

packages/cli/src/commands/set-admin.ts
```diff
@@ -19,46 +43,111 @@ const register: (program: any) => any = (program) => program
...
+  // has to check if a new admin address has balance or wallet
+  // if not display yet another warning
+  const balance = await ZWeb3.getBalance(pickedNewAdmin);
+  const code = await ZWeb3.getCode(pickedNewAdmin);
+  if (!force && (balance === 0x0.toString()) && (code === '0x')) {
+    const answers = await inquirer.prompt([
+      {
+        name: 'certain',
+        type: 'confirm',
+        message: 'The new admin address doesn\'t have any funds and not a wallet. Are you sure?',
```
jcarpanelli:
```suggestion
        message: 'The new admin address has no funds nor wallet. Are you sure you want to continue?',
```

packages/cli/src/commands/set-admin.ts
```diff
@@ -19,46 +43,111 @@ const register: (program: any) => any = (program) => program
...
 
-  if (contractFullNameOrAddress && contractFullNameOrAddress.startsWith('0x')) {
-    proxyAddress = contractFullNameOrAddress;
-  } else if (contractFullNameOrAddress) {
-    ({ contract: contractAlias, package: packageName } = fromContractFullName(contractFullNameOrAddress));
+  // has to check if a new admin address has balance or wallet
+  // if not display yet another warning
+  const balance = await ZWeb3.getBalance(pickedNewAdmin);
+  const code = await ZWeb3.getCode(pickedNewAdmin);
+  if (!force && (balance === 0x0.toString()) && (code === '0x')) {
```
jcarpanelli: Can we use the `promptIfNeeded`  and the `setCommandProps` here as well?

packages/cli/src/commands/update.ts
```diff
@@ -1,12 +1,19 @@
 import pickBy from 'lodash.pickby';
 
 import update from '../scripts/update';
-import Session from '../models/network/Session';
-import { parseInit } from '../utils/input';
-import { fromContractFullName } from '../utils/naming';
+import { parseContractReference } from '../utils/contract';
```
jcarpanelli: :heart: Thanks for this!

packages/cli/src/prompts/prompt.ts
```diff
@@ -48,7 +49,7 @@ export let DEFAULT_INTERACTIVE_STATUS = true;
 /*
  * This function will parse and wrap both arguments and options into inquirer questions, where
  * the 'arguments' are the parameters sent right after the command name, e.g., * zos create Foo
- * (Foo is the argument) and the optionn are the parameters sent right after a flag * e.g.,
+ * (Foo is the argument) and the options are the parameters sent right after a flag * e.g.,
```
jcarpanelli: Nice catch!

packages/cli/src/commands/update.ts
```diff
@@ -49,7 +50,7 @@ const register: (program: any) => any = (program) => program
 
 async function action(contractReference: string, options: any): Promise<void> {
   const { force, interactive, all } = options;
-  const networkOpts = await promptForNetwork(options);
+  const networkOpts = await promptForNetwork(options, () => getCommandProps());
```
jcarpanelli: Why this change?

ylv-io: Because `promptForNetwork` has been moved to `prompt.ts` and it needs access to local `getCommandProps` function.

packages/cli/src/commands/set-admin.ts
ylv-io: Just run a test and it worked. Have you considered this code?
```
if (!newAdmin) {
    newAdmin = contractReference;
    contractReference = '';
  }
```

packages/cli/src/commands/set-admin.ts
ylv-io: What exactly is confusing? This is important message. Mishandling `set-admin` command may result in a loss of millions of dollars and reputation. `Please, provide the last 4 characters of the specified address` doesn't put required significance on it.

packages/cli/src/commands/set-admin.ts
ylv-io: It is not related to `interactivity` as a comment says. It is a security feature thus it can't be disabled with `interactivity` flag. Also there is no `getCommandProps` function I assume you mean `getCommandProps` function.

packages/cli/src/commands/set-admin.ts
ylv-io: `force` flag exactly for integration tests and automation. Interactivity is not a security.

packages/cli/src/commands/set-admin.ts
ylv-io: We need linting for such things.

packages/cli/src/commands/set-admin.ts
ylv-io: Yes, good idea.

packages/cli/src/commands/set-admin.ts
jcarpanelli: I meant the `getCommandProps`

packages/cli/src/commands/set-admin.ts
jcarpanelli: Agree!

packages/cli/src/commands/set-admin.ts
ylv-io: Sure

packages/cli/src/commands/set-admin.ts
jcarpanelli: Yeap, I've seen it. But if the user runs `zos set-admin --no-interactive --network development` It will throw that ugly error instead of gently erroring with a "Please provide a contract address/admin address/whatever`.

packages/cli/src/commands/set-admin.ts
jcarpanelli: I think that prompt messages should be as simple (and short) as possible, but you are right, in this case it will be better to be more explicit about the consequences.  

What is confusing for me is that the prompt message is not finishing with a reference to what the user should answer. WDYT about rephrasing it to something like:
`Warning! If you provide a wrong address, you will lose control over your contracts. Please double check your address and type the last 4 characters of the new admin address:`

packages/cli/src/commands/set-admin.ts
jcarpanelli: My point is that we can still have the security feature without prompting a question to the user and just fail if no `--force` flag is provided.

packages/cli/src/commands/set-admin.ts
ylv-io: So, true. Fixed it.

packages/cli/src/commands/set-admin.ts
ylv-io: I like idea failing with  `--no-interactive` && `force` not set.

PR #840, opened 2019-05-07
Fix set-admin owner check

ylv-io:
* The PR adds `set-admin` owner check in order to provide more meaningful messaging, and prevents sending failing transactions.
* It also introduce `TxParams` interface across `lib` and `cli` project for consistent typing. 
* Refactors `ConfigVariablesInitializer` to ES6 class.

### DIFF
--- a/packages/lib/src/artifacts/ZWeb3.ts
+++ b/packages/lib/src/artifacts/ZWeb3.ts
@@ -18,6 +18,11 @@ const NETWORKS = {
   42: 'kovan'
 };
 
+export interface TxParams {
+  from?: string;
+  gas?: number;
+}
+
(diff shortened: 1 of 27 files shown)


packages/lib/src/artifacts/ZWeb3.ts
```diff
@@ -18,6 +18,11 @@ const NETWORKS = {
   42: 'kovan'
 };
 
+export interface TxParams {
```
spalladino: Could you double check whether `txParams` can include gasPrice as well? Clearly we are not using them from our code (or the compiler would have complained), but it would be useful to know if we are passing gasPrice onto web3 at any point.

PR #848, opened 2019-05-09
`send-tx` command

jcarpanelli:
Fixes https://github.com/zeppelinos/zos/issues/826.

This PR implements the `send-tx` command to call a function that is not pure nor viewable.
I decided to list all the proxies (not only the owned ones), as the instance might have functions without any call restrictions, but this can be changed if needed.

~Also, I still have to implement the `--value` and `--gas` flags, but it should be pretty straightforward to do (and will do it soon :smile:).~ => implemented.

Demo:
send-tx


ylv-io:
This is dope. Never seen such UX in Blockchain.

packages/cli/src/commands/interfaces.ts
```diff
@@ -1,28 +1,38 @@
...
+
+export interface SendTxSelectionParams extends SharedSelectionParams {}
+
+export interface UpdateSelectionParams extends SharedSelectionParams {
+  all: boolean;
 }
 
-export interface SetAdminSelectionParams {
-  address?: string;
+export interface SendTxPropsParams extends SharedPropsParams, MethodParams {
```
ylv-io: Glad that you have refactored these classes to base parents.

packages/cli/src/models/network/TransactionController.ts
```diff
@@ -13,9 +17,18 @@ interface ERC20TokenInfo {
 
 export default class TransactionController {
   public txParams: TxParams;
+  public packageFile: ZosPackageFile;
+  public networkFile: ZosNetworkFile;
 
-  constructor(txParams?: any) {
+  constructor(txParams?: any, network?: string, networkFile?: ZosNetworkFile) {
```
ylv-io: txParams should have type TxParams. I've recently added it. Let's keep using it everywhere!

packages/cli/src/models/network/TransactionController.ts
```diff
@@ -1,8 +1,12 @@
-import { Transactions, Logger, ZWeb3, TxParams } from 'zos-lib';
+import { Transactions, Logger, ZWeb3, TxParams, ABI } from 'zos-lib';
 import { isValidUnit, prettifyTokenAmount, toWei, fromWei } from '../../utils/units';
 import { ERC20_PARTIAL_ABI } from '../../utils/constants';
 import { allPromisesOrError } from '../../utils/async';
+import LocalController from '../local/LocalController';
+import ZosPackageFile from '../files/ZosPackageFile';
+import ZosNetworkFile from '../files/ZosNetworkFile';
 
+const { buildCallData, callDescription }  = ABI;
```
ylv-io:
```suggestion
const { buildCallData, callDescription } = ABI;
```

packages/cli/src/models/network/TransactionController.ts
```diff
@@ -1,8 +1,12 @@
-import { Transactions, Logger, ZWeb3, TxParams } from 'zos-lib';
+import { Transactions, Logger, ZWeb3, TxParams, ABI } from 'zos-lib';
 import { isValidUnit, prettifyTokenAmount, toWei, fromWei } from '../../utils/units';
 import { ERC20_PARTIAL_ABI } from '../../utils/constants';
 import { allPromisesOrError } from '../../utils/async';
+import LocalController from '../local/LocalController';
```
ylv-io: I am pretty certain controllers should not use other controllers.  Idea behind Controllers concept is usually a layer which communicates in our case between scripts layers and models layer. I suggest we stick to this pattern. Thoughts @jcarpanelli @spalladino?

packages/cli/src/scripts/send-tx.ts
```diff
@@ -0,0 +1,11 @@
+import TransactionController from '../models/network/TransactionController';
+import { SendTxParams } from './interfaces';
+
+export default async function sendTx({ proxyAddress, methodName, methodArgs, value, gas, network, txParams, networkFile }: SendTxParams): Promise<void | never> {
```
ylv-io: Why proxyAddress not checked? What if it is undefined?

packages/cli/src/commands/send-tx.ts
```diff
@@ -0,0 +1,85 @@
...
+import { SendTxPropsParams, SendTxSelectionParams } from './interfaces';
+import promptForMethod from '../prompts/method-params';
+
+const name: string = 'send-tx';
+const signature: string = name;
+const description: string = 'Send a transaction to the specified contract instance. Provide the [address], method to call and its arguments if needed';
+
+const register: (program: any) => any = (program) => program
+  .command(signature, undefined, { noHelp: true })
+  .usage('[options]')
```
ylv-io: Why we specify a contract address and a method as optional? It seems to me they are not optional.

ylv-io:
Just read the code. Stellar job @jcarpanelli! I left a couple small comments but generally speaking it is ready to be merged 🚀 🚢 !

packages/cli/src/models/network/TransactionController.ts
jcarpanelli: As you said in the past, we should definitely add linter rules for these :joy:

packages/cli/src/models/network/TransactionController.ts
jcarpanelli: I completely agree with you. As a matter of fact, it was a pain for me to import the `LocalController` into this new `TransactionController` :cry:. That being said, it was in our tech debt plans to modularize both `NetworkController` and `LocalController` into different controllers and models as apropiate, and `getContractClass` is definitely a function that will live in a model. Btw, if you check `LocalController#L132` you will see that we have commented where this function should be defined (in a future cli `Contract` model)

packages/cli/src/scripts/send-tx.ts
jcarpanelli: Good catch :+1:

packages/cli/src/commands/send-tx.ts
jcarpanelli: Good question.  My first answer would be because _some_ of the defined flags are indeed optional, but definitely that usage description is not accurate. We might check which is the `commander.js` or the general cli convention in these cases.

packages/cli/src/commands/send-tx.ts
```diff
@@ -0,0 +1,85 @@
...
+  const opts = { proxy: proxyAddress };
+  const props = getCommandProps({ network });
+  const { proxy: promptedProxy } = await promptIfNeeded({ opts, props }, interactive);
+
+  return promptedProxy;
+}
+
+function getCommandProps({ network, contractFullName, methodName, methodArgs }: SendTxPropsParams = {}): InquirerQuestions {
+  const methods = methodsList(contractFullName);
+  const args = argsList(contractFullName, methodName)
```
spalladino: If this construct (or others here) are repeated across several commands maybe we can abstract them somewhere? Can be done on a separate PR, of course.

packages/cli/src/models/network/TransactionController.ts
spalladino: https://github.com/zeppelinos/zos/issues/849

packages/cli/src/models/network/TransactionController.ts
spalladino: At the risk of being too much of a :policeman:, I'd suggest making that specific refactor _now_ (ie on this PR). I fear that if we allow this single reference between controllers, this will get out of hand real soon, and since this controller is using a single method from LocalController, it should be easy to refactor out to a separate model.

To be clear: I'm only saying to move `getContractClass` somewhere else to remove the need for this dependency, not to refactor the entire controllers.

packages/cli/src/models/network/TransactionController.ts
```diff
@@ -39,6 +52,21 @@ export default class TransactionController {
...
+    }
+
+    const { package: packageName, contract: contractName } = this.networkFile.getProxy(proxyAddress);
+    const localController = new LocalController(this.packageFile);
+    const contract = localController.getContractClass(packageName, contractName).at(proxyAddress);
+    const { method } = buildCallData(contract, methodName, methodArgs);
+
+    log.info(`Calling ${methodName} with: ${callDescription(method, methodArgs)}`);
+    const { transactionHash } = await Transactions.sendTransaction(contract.methods[methodName], methodArgs, this.txParams);
+    log.info(`Function successfully called! Transaction hash: ${transactionHash}`);
```
spalladino: We should properly handle the case where the tx reverts, showing something nice to the user, and include the transaction hash as well.

spalladino: Also, how difficult would it be to list the events emitted...?

spalladino:
```suggestion
    log.info(`Transaction successful: ${transactionHash}`);
```

packages/cli/src/models/network/TransactionController.ts
jcarpanelli: No more to say! :muscle:

packages/cli/src/commands/send-tx.ts
jcarpanelli: Agree! Will do it on a separate PR.

packages/cli/src/models/network/TransactionController.ts
jcarpanelli: Should be super easy. How should we list them? do you have something in mind?

PR #850, opened 2019-05-10
Minimal proxies

spalladino:
Adds PR for [EIP 1167 minimal](http://eips.ethereum.org/EIPS/eip-1167) non-upgradeable proxies. A user may now use the `--minimal` flag on the `create` command, which spawns a minimal proxy instead of a regular one. Under the hood, it spawns a `ProxyFactory` which is used to create and initialize the non-upgradeable proxies in a single tx.

### DIFF
--- a/packages/cli/src/models/network/NetworkController.ts
+++ b/packages/cli/src/models/network/NetworkController.ts
@@ -38,7 +38,8 @@ import {
   TxParams,
   ProxyAdmin,
   SimpleProject,
-  AppProxyMigrator
+  AppProxyMigrator,
+  MinimalProxy
 } from 'zos-lib';
 import { isMigratableZosversion } from '../files/ZosVersion';
 import { allPromisesOrError } from '../../utils/async';
@@ -52,6 +53,7 @@ import LocalController from '../local/LocalController';
 import ZosNetworkFile, { ProxyInterface } from '../files/ZosNetworkFile';
 import ZosPackageFile from '../files/ZosPackageFile';
 import { ZOS_VERSION } from '../files/ZosVersion';
+import { ProxyType } from '../../scripts/interfaces';
 
 const log = new Logger('NetworkController');
 type Project = ProxyAdminProject | AppProject;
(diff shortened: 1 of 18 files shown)


packages/cli/src/models/network/NetworkController.ts
```diff
@@ -544,22 +546,44 @@ export default class NetworkController {
...
+        proxy = await Proxy.at(instance.address);
+        return { proxy, instance };
+
+      case ProxyType.Minimal:
+        if (salt) {
+          throw new Error(`Cannot create a minimal proxy with a precomputed address, use an Upgradeable proxy instead.`);
+        }
+        instance = await this.project.createMinimalProxy(contract, createArgs);
+        proxy = await MinimalProxy.at(instance.address);
+        return { proxy, instance };
```
ylv-io: Please move to the bottom of the function and remove duplication.

packages/lib/src/project/AppProject.ts
```diff
@@ -212,6 +212,18 @@ class BaseAppProject extends BasePackageProject {
     return contract.at(proxy.address);
   }
 
+  public async createMinimalProxy(contract, contractInterface: ContractInterface = {}): Promise<Contract> {
```
ylv-io:
```suggestion
  public async createMinimalProxy(contract: Contract, contractInterface: ContractInterface = {}): Promise<Contract> {
```

packages/lib/src/proxy/MinimalProxy.ts
```diff
@@ -0,0 +1,19 @@
...
+    return new this(address);
+  }
+
+  constructor(address: string) {
+    this.address = toAddress(address);
+  }
+
+  public async implementation(): Promise<string> {
+    const code = await ZWeb3.getCode(this.address);
+    return `0x${code.slice(22, 62)}`;
```
ylv-io: Add comment here and link on EIP which explains what 22 and 62 are.

PR #853, opened 2019-05-13
`call` command

jcarpanelli:
Fixes https://github.com/zeppelinos/zos/issues/827
Implemented `zos call` to do a non-transactional call to a contract method. Usage:

call-cmd


packages/cli/src/prompts/prompt.ts
```diff
@@ -170,13 +170,13 @@ export function argsList(contractFullName: string, methodIdentifier: string, opt
...
 
-function contractMethods(contractFullName: string, options: MethodOptions, packageFile: ZosPackageFile): any[] {
+function contractMethods(contractFullName: string, constant: boolean = false, packageFile: ZosPackageFile): any[] {
   const { contract: contractAlias, package: packageName } = fromContractFullName(contractFullName);
   const contractManager = new ContractManager(packageFile);
   if (!contractManager.hasContract(packageName, contractAlias)) return [];
   const contract = contractManager.getContractClass(packageName, contractAlias);
 
-  return contractMethodsFromAst(contract, options);
+  return constant ? contractMethodsFromAbi(contract, constant) : contractMethodsFromAst(contract, constant);
```
spalladino: :+1:

packages/cli/src/models/network/TransactionController.ts
```diff
@@ -55,18 +56,27 @@ export default class TransactionController {
...
+      let result = await contract.methods[methodName](...methodArgs).call({ ...this.txParams });
+      if (typeof result === 'object' && Object.keys(result).length !== 0) {
+        result = `(${Object.values(result).join(', ')})`;
+      }
 
-    const { package: packageName, contract: contractName } = this.networkFile.getProxy(proxyAddress);
-    const contractManager = new ContractManager(this.packageFile);
-    const contract = contractManager.getContractClass(packageName, contractName).at(proxyAddress);
-    const { method } = buildCallData(contract, methodName, methodArgs);
+      isUndefined(result) || typeof result === 'object'
```
spalladino: Be careful with the typeof object check: if the result is an array, typeof an array is object, so the result will not be shown. Same applies to bignumbers (web3.js returns strings in the version 37, but have rolled back to bns in a more recent one). Also, I'm not sure what happens when returning structs.

packages/cli/src/models/network/TransactionController.ts
```diff
@@ -55,18 +56,27 @@ export default class TransactionController {
...
+        ? log.info(`Method ${methodName} successfully called.`)
+        : log.info(`Call returned: ${result}`);
```
spalladino: What do you think about, instead of logging the result here, return the value from the `controller` to the `script`? And have the script output the result to `stdout`, so it can be piped if needed? It would be similar to how `create` works in outputting the deployment address.

packages/cli/src/commands/call.ts
```diff
@@ -0,0 +1,83 @@
...
+async function promptForProxy(proxyAddress: string, network: string, options: any): Promise<SendTxSelectionParams> {
+  const { interactive } = options;
+  const opts = { proxy: proxyAddress };
+  const props = getCommandProps({ network });
+  const { proxy: promptedProxy } = await promptIfNeeded({ opts, props }, interactive);
+
+  return promptedProxy;
+}
+
+function getCommandProps({ network, contractFullName, methodName, methodArgs }: SendTxPropsParams = {}): InquirerQuestions {
```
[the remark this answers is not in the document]
jcarpanelli: Agree! I already have a good approach in mind to refactor this :slightly_smiling_face:

packages/cli/src/prompts/prompt.ts
jcarpanelli: Y has some doubts regarding this change. I loved that it was super easy to build the array of functions using the ABI, and I was wondering whether we should wrap the `contractMethodsFromAst` call inside a `try-catch`block  and use the `contractMethodsFromAbi` if the former fails (The `No AST nodes with id X found` came to my mind)

packages/cli/src/models/network/TransactionController.ts
jcarpanelli: We can do both!

jcarpanelli:
I added one more check to cover the case in which a `web3` method call returns `null` (it seems that for JS, `typeof null === object` is true :-1: )

PR #866, opened 2019-05-14
Store proxy admin address on failed create

### LINKED ISSUE #860: Deployed ProxyAdmin is not stored if contract deployment fails
If the deployment of the first contract in a network fails, then the ProxyAdmin address is not stored in the network json file, and it is re-deployed again afterwards. We should save that address regardless of whether the deployment of the contract itself succeeded. Same for ProxyFactory.

```
~/Projects/zeppelinos/examples/test-2.3$ npx zos create
? Choose a contract Sample
? Select a network from the network list ropsten
? Do you want to run a function after creating the instance? Yes
? Select a method [Initializable] initialize(_value: uint256)
? _value: 10
Deploying new ProxyAdmin...
Deployed ProxyAdmin at 0xafE242D34086fE9B1045A63F97Dc8594287E8F3F
Creating proxy to logic contract 0x7a3181d81623D698ca4e8FAf8d46F095DE9FE760 and initializing by calling initialize with: 
 - _value (uint256): "10"
The contract code couldn't be stored, please check your gas limit.
~/Projects/zeppelinos/examples/test-2.3$ npx zos create
? Choose a contract Sample
? Select a network from the network list ropsten
? Do you want to run a function after creating the instance? Yes
? Select a method [Initializable] initialize(_value: uint256)
? _value: 11
Deploying new ProxyAdmin...
Deployed ProxyAdmin at 0xA9289c673F98Ec9b703A0a35920f913a607ba45a
```

Screenshot from 2019-05-13 16-55-19

spalladino:
Fixes #860

Needs to be cherry-picked to release/2.3

### DIFF
--- a/packages/cli/contracts/mocks/ImplV1.sol
+++ b/packages/cli/contracts/mocks/ImplV1.sol
@@ -21,6 +21,11 @@ contract ImplV1 {
   function initializeNumbers(uint256[] memory _numbers) public {
     numbers = _numbers;
   }
+  
+  function initializeThatFails(uint256 _value) public {
+    require(false, "Fail");
+    value = _value;
+  }
 
   function say() public pure returns (string memory) {
     return "V1";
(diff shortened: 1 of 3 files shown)


spalladino:
Agree on the assessment Igor, and we can look into it when we break down the massive Controller into smaller chunks. I also have a few other ideas in mind, like killing the `scripts` altogether, but we can discuss them later.

spalladino:
Moved to 2.3 in 7bd0915d67ece824023a1ecffcd19bdb3472c3f0

PR #878, opened 2019-05-16
Run push before creating an instance

### LINKED ISSUE #862: Automatically push contract if it changed locally
When running `zos create` on a contract that changed (ie source modified and recompiled), the command fails with a "Contract Sample has changed locally since the last deploy, consider running 'zos push'." error. Instead, automatically run `zos push` on behalf of the user (if possible, only on the contract being created).

Screenshot from 2019-05-13 17-16-04

To reproduce:
- Write a contract, and compile with `truffle compile`
- Create a new instance via `zos create`
- Modify the contract, and recompile with `truffle compile`
- Create a new instance via `zos create` on the same network

jcarpanelli:
Same as https://github.com/zeppelinos/zos/pull/876 (fixes #862) but for master branch.

### DIFF
--- a/packages/cli/src/bin/options.ts
+++ b/packages/cli/src/bin/options.ts
@@ -16,12 +16,17 @@ program.Command.prototype.withNetworkOptions = function(): Command {
 program.Command.prototype.withPushOptions = function(): Command {
   return this
     .option('--push [network]', 'push all changes to the specified network')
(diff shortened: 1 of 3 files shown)


jcarpanelli:
Yeah, 100% agree that we should add tests for this. Not sure which is the best approach though :/.

PR #881, opened 2019-05-17
Fix zos.json error message

### LINKED ISSUE #873: Broken `zos.json` results to unacceptable output
If `zos.json` is broken output for almost any command will be `Unexpected end of JSON input` which doesn't give user any idea what is wrong. We should change it to something like 'zos.json file is broken, please fix it' or something like that.

image

ylv-io:
Fix #873

### DIFF
--- a/packages/cli/src/models/files/ZosPackageFile.ts
+++ b/packages/cli/src/models/files/ZosPackageFile.ts
@@ -27,7 +27,13 @@ export default class ZosPackageFile {
 
   constructor(fileName: string = 'zos.json') {
     this.fileName = fileName;
-    this.data = fs.parseJsonIfExists(this.fileName) || { zosversion: ZOS_VERSION };
+    try {
+      this.data = fs.parseJsonIfExists(this.fileName) || { zosversion: ZOS_VERSION };
+      // if we failed to read and parse zos.json
+    } catch(e) {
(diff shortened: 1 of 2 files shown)


ylv-io:
Absolutely. Totally missed it!

packages/cli/src/models/files/ZosPackageFile.ts
```diff
@@ -27,7 +27,13 @@ export default class ZosPackageFile {
 
   constructor(fileName: string = 'zos.json') {
     this.fileName = fileName;
-    this.data = fs.parseJsonIfExists(this.fileName) || { zosversion: ZOS_VERSION };
+    try {
+      this.data = fs.parseJsonIfExists(this.fileName) || { zosversion: ZOS_VERSION };
+      // if we failed to read and parse zos.json
+    } catch(e) {
+      e.message = `Failed to read and parse ${fileName} file. Please make sure ${fileName} is present and valid. Details: ${e.message}.`;
```
spalladino:
```suggestion
      e.message = `Failed to parse ${fileName} file. Please make sure that ${fileName} is a valid JSON file. Details: ${e.message}.`;
```

spalladino: Also, and given that we are allowing people to run zos from other folders, could we resolve `fileName` to the full path when outputting the error?

ylv-io:
I'll add zos.network.json to this PR. Full path for `fileName` is good idea.

PR #886, opened 2019-05-17
Fix prompt#contractsList with source fromLocal

### LINKED ISSUE #884: Interactive remove command fails with "Cannot read property MyContractName of undefined"
Having a `zos.json` with at least one contract and running `zos remove` yields the following error in 2.3:
```
Cannot read property 'ImplV1' of undefined
```

This is probably related to this line, which is checking a non-existing `this.contracts` instead of `localPackageFile.contracts`.

https://github.com/zeppelinos/zos/blob/2a97b2432b07d5e50066c2232dfd648e552495d6/packages/cli/src/prompts/prompt.ts#L123

spalladino:
Fixes bug on `source === 'fromLocal'` branch, and adds tests.

Fixes #884

### DIFF
--- a/packages/cli/test/prompts/prompt.test.js
+++ b/packages/cli/test/prompts/prompt.test.js
@@ -97,20 +97,45 @@ describe('prompt', function() {
     describe('#contractsList', function() {
       beforeEach('set stub and initialize', function() {
         sinon.stub(Truffle, 'getContractNames').returns(['Foo', 'Bar']);
+        sinon.stub(ZosPackageFile.prototype, 'dependencies').get(() => ({'mock-stdlib': '1.1.0'}));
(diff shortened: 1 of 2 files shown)


packages/cli/test/prompts/prompt.test.js
```diff
@@ -97,21 +97,46 @@ describe('prompt', function() {
...
+        const contracts = contractsList('keyName', 'Im a message', 'listy', 'fromBuildDir');
 
         contracts.should.be.an('object');
         contracts.keyName.should.be.an('object').that.has.all.keys('type', 'message', 'choices');
         contracts.keyName.type.should.eq('listy');
         contracts.keyName.message.should.eq('Im a message');
         contracts.keyName.choices.should.include.members(['Foo', 'Bar']);
       });
+
+      it('returns an object with correct keys and values from local', function() {
```
jcarpanelli: :raised_hands:

Issue #874, opened 2019-05-16
Set CLI interactivity via an environment variable


spalladino:
Allow changing the value of `DEFAULT_INTERACTIVE_STATUS` in `prompt.ts` via an environment variable, such as `ZOS_NON_INTERACTIVE`.

Also, if that environment variable is not present, check whether [`DEBIAN_FRONTEND`](https://askubuntu.com/questions/972516/debian-frontend-environment-variable) is set to `noninteractive`. If so, then also default to non-interactivity.

spalladino:
Note that the test setup file sets `DEFAULT_INTERACTIVE_STATUS` to false when the tests load. Take this into account when writing the tests for this feature.

https://github.com/zeppelinos/zos/blob/5bdb946ee3373abc35b8cfc8672d3e430a8d3f00/packages/cli/test/setup.js#L13

spalladino:
Fixed in #887

PR #887, opened 2019-05-17
Add interactivity env

ylv-io:
Implements #874;
Would help to implement tools like Solidity Hot Loader which have to run all commands in non interactive mode.
You can set `zos` in non interactive mode by setting either `ZOS_NON_INTERACTIVE` env variable or `DEBIAN_FRONTEND` to `noninteractive`.

@jcarpanelli can you take a look at line 59 and 70 in prompt.test.js? These two test have the same code but different names. Pretty sure that is not what intended.

### DIFF
--- a/packages/cli/src/commands/publish.ts
+++ b/packages/cli/src/commands/publish.ts
@@ -17,13 +17,13 @@ const register: (program: any) => any = (program) => program
   .action(action);
 
 async function action(options: any): Promise<void> {
-  const { network: networkInArgs } = options;
+  const { network: networkInArgs, interactive } = options;
   const { network: networkInSession } = Session.getOptions();
   const defaults = { network: Session.getNetwork() };
   const opts = { network: networkInSession || networkInArgs };
   const props = getCommandProps();
 
-  const promptedOpts = await promptIfNeeded({ opts, defaults, props });
+  const promptedOpts = await promptIfNeeded({ opts, defaults, props }, interactive);
   const { network, txParams } = await ConfigVariablesInitializer.initNetworkConfiguration(promptedOpts);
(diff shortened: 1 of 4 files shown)


packages/cli/test/setup.js
```diff
@@ -4,13 +4,11 @@ import { ZWeb3, Contracts } from 'zos-lib'
 import Dependency from '../src/models/dependency/Dependency'
 import ZosPackageFile from '../src/models/files/ZosPackageFile'
 import ZosNetworkFile from '../src/models/files/ZosNetworkFile'
-const prompt = require('../src/prompts/prompt');
 
 useTestZosPackageFile()
 doNotInstallStdlib()
 ZWeb3.initialize(web3.currentProvider)
 setArtifactDefaults()
-prompt.DEFAULT_INTERACTIVE_STATUS = false
```
spalladino: Isn't this required to prevent the tests in `test/commands` from being interactive?

ylv-io: I think no. Two reason:
* DEFAULT_INTERACTIVE_STATUS was never used because commander.js always set `interactive` flag regardless of will you provide it or not
* I think all our commands tests doesn't ever trigger interactivity.

packages/cli/src/commands/publish.ts
```diff
@@ -17,13 +17,13 @@ const register: (program: any) => any = (program) => program
   .action(action);
 
 async function action(options: any): Promise<void> {
-  const { network: networkInArgs } = options;
+  const { network: networkInArgs, interactive } = options;
```
jcarpanelli: Good catch!

PR #889, opened 2019-05-17
Remove useless question in `zos link` prompt

jcarpanelli:
Related to https://github.com/zeppelinos/zos/issues/872.

### DIFF
--- a/packages/cli/src/commands/link.ts
+++ b/packages/cli/src/commands/link.ts
@@ -20,14 +20,13 @@ const register: (program: any) => any = (program) => program
 
 async function action(dependencies: string[], options: any): Promise<void> {
   const { install, forceInstall, interactive } = options;
-  const installDependencies = forceInstall || (install && interactive ? undefined : install);
+  const installDependencies = install || forceInstall;
   const args = { dependencies };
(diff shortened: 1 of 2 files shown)


packages/cli/src/commands/link.ts
```diff
@@ -20,14 +20,13 @@ const register: (program: any) => any = (program) => program
 
 async function action(dependencies: string[], options: any): Promise<void> {
   const { install, forceInstall, interactive } = options;
-  const installDependencies = forceInstall || (install && interactive ? undefined : install);
+  const installDependencies = install || forceInstall;
```
spalladino: Where does `forceInstall` come from?

jcarpanelli: From the `create` command.  If a user runs a `zos create openzeppelin-eth/SomeContract` without having `openzeppelin-eth` linked, zos will install and link it for them.

spalladino:
To add more context for future reference: we decided to remove this question since it did not add much value, and it caused the command to always prompt (unless `--no-interactive` was set) even if all arguments were set, which would have broken backwards compatibility for non-interactive scripts relying on `zos link`.

PR #902, opened 2019-05-22
Pick contract functions always from ABI instead of using AST.

### LINKED ISSUE #900: Interactive `zos create/update` lists all contracts methods from AST.
When running `zos create` or `zos update` in interactive mode, `zos` will not only show a list of the current contract methods, but also the methods of those contracts that are part of its inheritance hierarchy.
A simple way to solve this would be to take the contract methods from the ABI instead of taking them from the AST, and then look for matches in the AST to take the modifiers.

jcarpanelli:
Fixes #900. Now we are picking contract functions from the `ABI` and we are only using the `AST` to pick functions modifiers.

### DIFF
--- a/packages/lib/src/artifacts/Contract.ts
+++ b/packages/lib/src/artifacts/Contract.ts
@@ -128,13 +128,18 @@ export function createContract(schema: any): Contract {
 
 export function contractMethodsFromAbi(instance: Contract, constant: ContractMethodMutability = ContractMethodMutability.NotConstant): any[] {
   const isConstant = constant === ContractMethodMutability.Constant;
+  const contractAst = new ContractAST(instance, null, { nodesFilter: ['ContractDefinition'] });
+  const methodsFromAst = contractAst.getMethods();
 
   return instance.schema.abi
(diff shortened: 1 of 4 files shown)


packages/lib/src/artifacts/Contract.ts
```diff
@@ -128,13 +128,18 @@ export function createContract(schema: any): Contract {
 
 export function contractMethodsFromAbi(instance: Contract, constant: ContractMethodMutability = ContractMethodMutability.NotConstant): any[] {
   const isConstant = constant === ContractMethodMutability.Constant;
+  const contractAst = new ContractAST(instance, null, { nodesFilter: ['ContractDefinition'] });
+  const methodsFromAst = contractAst.getMethods();
```
spalladino: In which order does `getMethods` return them, in case you have overridden methods? For instance, if you have the following scenario:

```solidity
contract Base is Initializable {
  function foo() initializer { }
}

contract Child is Base {
  function foo() { }
}
```

Then `contractMethodsFromAbi` should return `foo` **without** the initializable flag. But if `getMethods` return from less to most derived (which I think it may be the case), then `find` will hit the first occurrence, and return that it does have an initializer.

PR #904, opened 2019-05-22
Add a new contract list type for getting non-added contracts

jcarpanelli:
Fixes https://github.com/zeppelinos/zos/issues/899. I added a new contract list type (`notAdded`) and also renamed the already defined types.

### DIFF
--- a/packages/cli/src/prompts/prompt.ts
+++ b/packages/cli/src/prompts/prompt.ts
@@ -128,7 +128,7 @@ export function contractsList(name: string, message: string, type: string, sourc
     return inquirerQuestion(name, message, type, contractsFromBuild);
   // get contracts from zos.json file
   } else if (source === 'notAdded') {
-    const contracts = difference(contractsFromBuild, contractsFromLocal.map(a => a.value));
+    const contracts = difference(contractsFromBuild, contractsFromLocal.map(({ value }) => value));
     return  inquirerQuestion(name, message, type, contracts);
   } else if(source === 'added') {
     return inquirerQuestion(name, message, type, contractsFromLocal);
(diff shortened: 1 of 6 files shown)


packages/cli/test/prompts/prompt.test.js
```diff
@@ -119,46 +118,61 @@ describe('prompt', function() {
 
     describe('#contractsList', function() {
       beforeEach('set stub and initialize', function() {
-        sinon.stub(Truffle, 'getContractNames').returns(['Foo', 'Bar']);
-        sinon.stub(ZosPackageFile.prototype, 'dependencies').get(() => ({'mock-stdlib': '1.1.0'}));
+        sinon.stub(Truffle, 'getContractNames').returns(['Foo', 'Bar', 'Buz']);
```
spalladino: It's `Baz`, not `Buz`!

packages/cli/src/prompts/prompt.ts
```diff
@@ -114,20 +115,22 @@ export function proxiesList(pickProxyBy: string, network: string, packageFile?:
...
+    .keys(localPackageFile.contracts)
+    .map(alias => ({ name: localPackageFile.contracts[alias], alias }))
+    .map(({ name: contractName, alias }) => {
+      const label = contractName === alias ? alias : `${alias}[${contractName}]`;
+      return { name: label, value: alias };
+    });
 
   // get contracts from `build/contracts`
-  if (!source || source === 'fromBuildDir') {
+  if (!source || source === 'built') {
```
spalladino: Is this option used at all? If not, can we remove it?

spalladino:
Looks good, merging! Left one minor comment which can be addressed later.

packages/cli/src/prompts/prompt.ts
jcarpanelli: Mmh, I believe we are not using it! My bad!

PR #909, opened 2019-05-24
Remove web3-provider-engine dependency

### LINKED ISSUE #907: Remove babel from CLI
The CLI depends on web3-provider-engine, which requires json-rpc-engine, which [bundles babel as dependency](https://github.com/MetaMask/json-rpc-engine/issues/10). 

We are currently using web3-provider-engine only to manually inject a NonceTracker to an hdwallet-provider if present, due to [this issue](https://github.com/trufflesuite/truffle-hdwallet-provider/issues/65). See https://github.com/zeppelinos/zos/issues/586 for a recent occurrence of that problem in `zos`.

See if we can work this around by including only what we strictly need from provider-engine, to avoid importing json-rpc-engine at all. Another option would be to check if the issue in hdwallet-provider is fixed, and just suggest our users to upgrade if we detect they are using an old version.

jcarpanelli:
Fixes #907. `core-js` is not part of our `package-lock.json` file, so I expect we won't see any `post-install` message from it anymore.
I also:
- Added a function in the `Truffle.ts` file (`_checkHdWalletProviderVersion`) to gently warn the user if they are using an old version of `truffle-hdwallet-provider`, 
- Removed the `_setNonceTrackerIfNeeded` function, as it was doing nothing in versions greater than `1.0.0` of `truffle-hdwallet-provider`.


packages/cli/src/models/initializer/truffle/Truffle.ts
```diff
@@ -92,23 +97,17 @@ const Truffle = {
...
-      });
+  async _checkHdWalletProviderVersion(provider: any, path: string = process.cwd()): Promise<void> {
+    if (provider.constructor.name !== 'HDWalletProvider') return;
+    const packagesList = await npm.list(path);
+    const hdwalletProviderPackage = packagesList
+      .find(packageNameAndVersion => packageNameAndVersion.match(/truffle-hdwallet-provider/));
+    if (hdwalletProviderPackage) {
+      const [name, version] = hdwalletProviderPackage.split('@');
+      if (version && semver.lt(version, '1.0.0')) {
+        log.warn(`WARNING: Version ${version} of truffle-hdwallet-provider might cause errors when sending transactions. Consider upgrading its version to '1.0.0' or higher`);
```
spalladino:
```suggestion
        log.warn(`Version ${version} of truffle-hdwallet-provider might fail when deploying multiple contracts. Consider upgrading it to version '1.0.0' or higher.`);
```

packages/cli/src/models/initializer/truffle/Truffle.ts
```diff
@@ -92,23 +97,17 @@ const Truffle = {
...
-        if (aProvider.constructor.name === 'ProviderSubprovider') {
-          nonceTracker.setEngine(engine);
-          engine._providers.splice(index, 0, nonceTracker);
-        }
-      });
+  async _checkHdWalletProviderVersion(provider: any, path: string = process.cwd()): Promise<void> {
+    if (provider.constructor.name !== 'HDWalletProvider') return;
+    const packagesList = await npm.list(path);
+    const hdwalletProviderPackage = packagesList
+      .find(packageNameAndVersion => packageNameAndVersion.match(/truffle-hdwallet-provider/));
```
spalladino:
```suggestion
      .find(packageNameAndVersion => packageNameAndVersion.match(/^truffle-hdwallet-provider@/));
```
Not sure if this regex works, but I guess the CI will tell!

packages/cli/src/models/initializer/truffle/Truffle.ts
```diff
@@ -92,23 +97,17 @@ const Truffle = {
...
-          engine._providers.splice(index, 0, nonceTracker);
-        }
-      });
+  async _checkHdWalletProviderVersion(provider: any, path: string = process.cwd()): Promise<void> {
+    if (provider.constructor.name !== 'HDWalletProvider') return;
+    const packagesList = await npm.list(path);
+    const hdwalletProviderPackage = packagesList
+      .find(packageNameAndVersion => packageNameAndVersion.match(/^truffle-hdwallet-provider@/));
+    if (hdwalletProviderPackage) {
+      const [name, version] = hdwalletProviderPackage.split('@');
```
spalladino: What happens if the package belongs to an organization, such as `@resolver-engine/imports-fs`? Wouldn't the split return an empty string for `name`?

spalladino: Resolved offline by @jcarpanelli - since we are matching on a specific package, doesn't matter

PR #911, opened 2019-05-27
Merge 2.3 into master

spalladino:


packages/docs/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "zos-docs",
-  "version": "2.3.0-alpha.1",
+  "version": "2.3.0-rc.3",
```
jcarpanelli:
```suggestion
  "version": "2.3.0",
```
we should also run a `lerna bootstrap` or a local `npm install` inside the `docs` package to update the `package-lock.json` Would you like to include it in this PR?

tests/cli/test/package.json
```diff
@@ -1,7 +1,7 @@
 {
   "name": "tests-cli-test",
   "private": true,
-  "version": "2.3.0-alpha.1",
+  "version": "2.3.0-rc.3",
```
jcarpanelli:
```suggestion
  "version": "2.3.0",
```

packages/docs/package.json
spalladino: I'll update manually for now, and look for a more robust solution later.

PR #913, opened 2019-05-29
Add ZepKit integration tests

ylv-io:
Adding ZepKit integration tests would help us to verify that our kits can be installed and work fine with ZeppelinOS.


tests/zepkit/test/package.json
```diff
@@ -0,0 +1,19 @@
+{
+  "name": "tests-kit-test",
+  "private": true,
+  "version": "2.3.0",
+  "description": "Integration tests for ZepKit and an unpack command",
+  "author": "santiago@zeppelin.solutions",
```
spalladino:
```suggestion
  "author": "igor@zeppelin.solutions",
```
He who kickstarts the project gets to author it :-)

tests/zepkit/test/package.json
```diff
@@ -0,0 +1,19 @@
...
+  "license": "MIT",
+  "scripts": {
+    "test": "scripts/test.sh"
+  },
+  "devDependencies": {
+    "chai": "^4.1.2",
+    "lodash": "^4.17.0",
+    "mocha": "5.2.0",
+    "truffle": "^5.0.19",
+    "truffle-hdwallet-provider": "0.0.6"
```
spalladino: WDYT about moving to a more recent truffle-hdwallet-provider? Actually, is this used in this suite?

tests/zepkit/test/src/integration.test.js
```diff
@@ -0,0 +1,55 @@
...
+  run('cd .. && rm -rf workdir && mkdir workdir');
+}
+
+const network = getNetwork();
+
+function runIntegrationTest({ kit }, action) {
+  before('cleaning up project folder', cleanup);
+
+  it('unpack kit', function () {
+    run(`npx zos unpack ${kit}`);
```
spalladino: Big question here: what `zos` is being used here? We need to make sure the current version from CLI is being used, and *not* the one published to npm.

tests/cli/workdir/package-lock.json
```diff
@@ -1,9 +1,97 @@
 {
 	"name": "tests-cli-workdir",
-	"version": "2.3.0-alpha.1",
+	"version": "2.3.0",
 	"lockfileVersion": 1,
 	"requires": true,
 	"dependencies": {
+		"@resolver-engine/core": {
```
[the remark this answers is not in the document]
ylv-io: No idea. I haven't paid an attention. Should I investigate?

tests/zepkit/test/package.json
ylv-io: Not used at all. I've remove it.

tests/zepkit/test/src/integration.test.js
ylv-io: You got me on this one 😅  NPX would pull down latest version from npm. Let me figure it out. There is even a bigger problem that Hot Loader will pull either local or global version. It would be perfect to trick npx to use dev version. I wonder is it possible?

tests/cli/workdir/package-lock.json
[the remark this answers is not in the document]
spalladino: At least, if we haven't changed the `package.json`, I'd avoid committing these changes here

tests/zepkit/test/src/integration.test.js
spalladino: It should be. If there is a `package.json` on the workdir, that points to the same `zos` version as in the current project, then `lerna bootstrap` **should** turn that dependency into a symlink to the local package folder.

At least, that's how the current `test/cli` is supposed to be working.

(as a side note, I'm considering removing the need for lerna bootstrap and replace it with just file references, which seems to be the latest recommended way of using lerna; but let's stick with this method for now)

tests/zepkit/test/src/integration.test.js
```diff
@@ -18,9 +18,15 @@ function runIntegrationTest({ kit }, action) {
...
 
   it('unpack kit', function () {
-    run(`npx zos unpack ${kit}`);
+    // have to replace zos with local version so we test current build
+    run(`node ../../../packages/cli/lib/bin/zos-cli.js unpack ${kit}`);
   })
 
+  // have to replace zos with local version so we test current build
+  it('replace zos with local version', function () {
+    run('~/zos/node_modules/.bin/lerna bootstrap --scope=tests-cli-zepkit --scope="zos*"');
```
spalladino: `~/zos` will only work where the project is checked out directly on the home folder (for instance, it would fail on my machine). I'd change this for `npx lerna`.

ylv-io: `npx lerna` fails on circleci. I'll figure it out. No worries.

spalladino: An alternative is to skip the lerna bootstrap step, run `npm install`, and then use [`npm link`](https://docs.npmjs.com/cli/link) manually to use the local one. If I understand correctly, you should run `npm link` on `packages/cli` to symlink it to the global scope, and then `npm link zos` in the workdir to link-install it.

ylv-io: Yeah. I am aware of `npm link`. I'll go with lerna for now for the sake of simplicity and consistency.

spalladino: > npx lerna fails on circleci. I'll figure it out. No worries.

I wonder why it failed. Still, findUp is a good alternative!

PR #915, opened 2019-05-30
Remove ora and add `spinnies`

ylv-io:
We totally should use a masterpiece written by @jcarpanelli! 🚢 🚀


ylv-io:
Just have to fix a `package.json` conflict.

spalladino:
I had assigned Jota as a reviewer, only to note that the original commit was his :-P
Let's move forward with this!!

PR #918, opened 2019-05-30
Add ZosConfig file (networks.js)

jcarpanelli:
Fixes https://github.com/zeppelinos/zos/issues/800.

Adds a new `networks.js` file to the project for managing networks. It manages string providers using `web3`, and also works fine using `truffle-hdwallet-provider`.


packages/cli/src/models/config/ZosConfig.ts
```diff
@@ -0,0 +1,112 @@
...
+    const buildDir = `${root}/build/contracts`;
+
+    return { ...zosConfigFile, compilers, buildDir };
+  }
+
+  public getBuildDir() {
+    return this.config ? this.config.buildDir : `${process.cwd()}/build/contracts`;
+  }
+
+  // TODO: set types.
```
spalladino: Let's definitely add types here

packages/cli/networks.js
```diff
@@ -0,0 +1,20 @@
...
+      network_id: '*',
+      gas: 5000000
+    },
+    testing: {
+      host: 'localhost',
+      network_id: '4447',
+      port: 9555
+    },
+  },
+  mocha: {
```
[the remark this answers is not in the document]
jcarpanelli: you mean the whole file?

packages/cli/scripts/copy-files.sh
```diff
@@ -1,4 +1,4 @@
 #!/bin/bash
 
-# blueprint.truffle.js
-cp src/models/initializer/truffle/blueprint.truffle.js lib/models/initializer/truffle/blueprint.truffle.js
+# blueprint.networks.js
+cp src/models/config/blueprint.networks.js lib/models/config/blueprint.networks.js
```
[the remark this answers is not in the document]
jcarpanelli: yeah, thanks for the `networks.js` name idea ;)

packages/cli/src/models/config/ZosConfig.ts
```diff
@@ -0,0 +1,112 @@
...
+
+  private setArtifactDefaults(zosConfigFile: any, network: any) {
+    const defaults = ['gas', 'gasPrice', 'from'];
+    const configDefaults = omit(pick(zosConfigFile, defaults), isUndefined);
+    const networkDefaults = omit(pick(network, defaults), isUndefined);
+
+    return { ...configDefaults, ...networkDefaults };
+  }
+
+  private setDefaultCompilersProperties() {
```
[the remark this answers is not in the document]
jcarpanelli: come on, @jcarpanelli! getters get stuff and setters set stuff!!! :man_facepalming:

packages/cli/networks.js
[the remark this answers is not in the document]
spalladino: The mocha bit, since it's unrelated to network config

packages/cli/src/models/config/ZosConfig.ts
```diff
@@ -0,0 +1,112 @@
...
+    const config = this.getConfig(root);
+    const { networks } = config;
+    if (!networks[networkName]) throw Error(`Given network '${networkName}' is not defined in your networks.js file`);
+
+    const network = networks[networkName];
+    const provider = this.setProvider(networks[networkName]);
+    const artifactDefaults = this.setArtifactDefaults(config, networks[networkName]);
+
+    return {
+      ...config,
```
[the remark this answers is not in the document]
jcarpanelli: The original config does not depend on any network in particular, and only stores the networks object read in the `networks.js`, the compilers info (which you will probably move to the `zos.json` and the `buildDIr`.

The `loadNetworkConfig` function takes this config which is independent of a network and adds some other info to it, such as the provider, artifacts defaults and the selected network.

packages/cli/src/models/config/ZosConfig.ts
```diff
@@ -0,0 +1,112 @@
...
+    this.createContractsDir(root);
+    this.createZosConfigFile(root);
+  }
+
+  public exists(root: string = process.cwd()): boolean {
+    return FileSystem.exists(`${root}/networks.js`);
+  }
+
+  public getConfig(root: string = process.cwd()) {
+    if (this.config) return this.config;
```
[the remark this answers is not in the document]
jcarpanelli: Nope, good catch.

packages/cli/networks.js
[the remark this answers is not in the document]
jcarpanelli: I moved the file outside the root directory (i created a `test/mocks/config-files`).

jcarpanelli:
@spalladino, this should be ready for another review! Basically, I:
- Added tests for `ConfigManager` object (formerly `ConfigVariablesInitializer`), which didn't have any.
- Created new mock folder `test/mocks/config-files` for reading both `truffle.js` and `networks.js` files for testing.
- Add `network_id` validations. I'm not happy using snake_case here though, but I don't want to add a point of friction to users that are used to work with truffle.
- Added types for `ZosConfig`.
- Added a custom error when `ZWeb3` fails in the first network-related connection. I sort of copy pasted the it from other place (hehe.. :roll_eyes:). Let's see if we can come up with a better error message together :)

packages/cli/src/models/config/ConfigManager.ts
```diff
@@ -23,12 +23,21 @@ const ConfigManager = {
...
     Contracts.setArtifactsDefaults(artifactDefaults);
 
-    const txParams = { from: ZWeb3.toChecksumAddress(from || artifactDefaults.from || await ZWeb3.defaultAccount()) };
-    return { network: await ZWeb3.getNetworkName(), txParams };
+    try {
+      ZWeb3.initialize(provider);
+      const txParams = { from: ZWeb3.toChecksumAddress(from || artifactDefaults.from || await ZWeb3.defaultAccount()) };
+      return { network: await ZWeb3.getNetworkName(), txParams };
+    } catch(error) {
+      if (this.config.constructor.name === 'ZosConfig') {
```
spalladino: Is it possible that this gets called before `setBaseConfig`? If so, could `this.config` be null?

packages/lib/src/artifacts/ZWeb3.ts
```diff
@@ -140,8 +140,12 @@ export default class ZWeb3 {
     return ZWeb3.eth().net.getId();
   }
 
-  public static async getNetworkName(): Promise<string> {
+  public static async getNetworkName(providedNetworkId?: string | number): Promise<string | never> {
```
spalladino: Not sure I like doing this test here. We are abusing the fact that one of the first methods we call on the Web3 provider is `getNetworkName`. We should validate the network ID when the provider is set up, not in a random method.

packages/lib/src/artifacts/ZWeb3.ts
```diff
@@ -140,8 +140,12 @@ export default class ZWeb3 {
...
     const networkId = await ZWeb3.getNetwork();
+    if (providedNetworkId !== undefined && providedNetworkId !== '*' && Number(networkId) !== Number(providedNetworkId)) {
+      throw Error(`Invalid network id ${providedNetworkId}`);
```
spalladino: I'd rephrase to sth like "Unexpected network ID: requested ${provided} but connected to ${networkId}"

packages/cli/test/scripts/init.test.js
```diff
@@ -6,18 +6,27 @@ import { cleanup, cleanupfn } from '../helpers/cleanup'
...
 contract('init script', function() {
   const name = 'MyApp';
   const version = '0.3.0';
   const tmpDir = 'test/tmp';
 
-  before('create tmp dir', () => fs.createDir(tmpDir))
-  after('cleanup tmp dir', cleanupfn(tmpDir))
+  before('create tmp dir', function () {
+    fs.createDir(tmpDir)
+    this.zosConfigInitialize = ZosConfig.prototype.initialize;
```
spalladino: Why not sinon? (agree on leaving this like this though, no need to change)

packages/cli/src/models/config/ConfigManager.ts
```diff
@@ -17,24 +17,24 @@ const ConfigManager = {
...
     Contracts.setSyncTimeout(timeout * 1000);
     Contracts.setArtifactsDefaults(artifactDefaults);
 
     try {
       ZWeb3.initialize(provider);
       const txParams = { from: ZWeb3.toChecksumAddress(from || artifactDefaults.from || await ZWeb3.defaultAccount()) };
-      return { network: await ZWeb3.getNetworkName(), txParams };
+      return { network: await ZWeb3.getNetworkName(network_id), txParams };
     } catch(error) {
       // TODO: change error message
```
spalladino: Adding a reminder to do this TODO!

spalladino:
> Add network_id validations. I'm not happy using snake_case here though, but I don't want to add a point of friction to users that are used to work with truffle.

Why not support both? Should be easy to do!

> Added a custom error when ZWeb3 fails in the first network-related connection. Let's see if we can come up with a better error message together :)

I'd go with sth like "Could not connect to the ${networkName} Ethereum network via ${providerType} on ${url}. Please check your networks.js configuration file." If we don't have info on the provider type, then we can just drop the "via ${providerType}" part.

packages/cli/test/scripts/init.test.js
jcarpanelli: no strong reason, but I prefer to use sinon. Nice catch.

packages/lib/src/artifacts/ZWeb3.ts
jcarpanelli: I agree, but I wouldn't change the `ZWeb3#initialize` implementation. WDYT if I add a new function to `ZWeb3`, something like `checkValidNetworkId` which will throw if the id is invalid? It will be called jsut after `ZWeb3#initialize`

spalladino: Sounds good!

packages/cli/src/models/config/ConfigManager.ts
jcarpanelli: It shouldn't. `this.setBaseConfig` will always be called before this `if` statement, as it calls `this.initStaticConfiguration` which calls `this.setBaseConfig`.

spalladino: Still, could we handle it just so we don't risk failing in a `catch` block?

jcarpanelli:
I implemented your suggestions @spalladino! Also I rebased master into this branch, and transformed `ZosConfig` and `Truffle` (now `TruffleConfig` for consistency) into objects (there was no reason for having classes over objects in both cases). Last but not least, I reran the linter.

packages/cli/src/models/config/ZosConfig.ts
```diff
@@ -17,16 +17,25 @@ interface Config {
...
-interface Network {
+interface NetworkCamel<T> {
+  networkId: T;
+}
+
+interface NetworkSnake<T> {
+  network_id: T;
+}
+
+type NetworkId<T> = NetworkCamel<T> | NetworkSnake<T> | (NetworkCamel<T> & NetworkSnake<T>);
```
spalladino: Why is the `(NetworkCamel<T> & NetworkSnake<T>)` needed? Isn't that intersection empty?

spalladino: Also, I'd just embed optional network_id and networkId directly in the Network type instead of defining separate interfaces.

examples/cli-create2/index.js
```diff
@@ -1,10 +1,10 @@
-const { ConfigVariablesInitializer, files, scripts, stdout } = require('zos');
+const { ConfigManager, files, scripts, stdout } = require('zos');
```
spalladino: Could you add a note to the `lib/changelog`, indicating this breaking change, so we don't forget about it when we add this feature to the CLI changelog?

pcowgill:
@jcarpanelli @spalladino Is the idea that the new `networks.js` file can be used to configure everything that can be configured with `truffle-config.js`? Or will many projects have both files if they're using Truffle?

In particular, I'm wondering whether both options should be reflected [here in the docs](https://docs.openzeppelin.com/sdk/2.5/faq#how-do-i-use-es6-javascript-syntax-in-my-tests).

Thanks!

spalladino:
>  Is the idea that the new networks.js file can be used to configure everything that can be configured with truffle-config.js?

Nope, we want to push networks.js as a framework-independent configuration file (hence the generic name) for specifying just networks. Info on compiler or testing should not be included there.

> In particular, I'm wondering whether both options should be reflected here in the docs.

That is specific to running tests via truffle, since you cannot decide how to run mocha when running tests. We are still designing how to push framework-independent testing, but we would like to avoid hiding too much behind a custom test command, and rather provide some basic functions to handle it that the user can call.

pcowgill:
> Info on compiler or testing should not be included there.

Got it - great to know, thanks!

> That is specific to running tests via truffle, since you cannot decide how to run mocha when running tests. We are still designing how to push framework-independent testing, but we would like to avoid hiding too much behind a custom test command, and rather provide some basic functions to handle it that the user can call.

Makes sense - thanks for taking the time to provide this context!

spalladino:
> Makes sense - thanks for taking the time to provide this context!

On the contrary, thanks for taking the time for contributing here!!

PR #920, opened 2019-05-31
Add ESLint and Prettier configs; Apply linting;

ylv-io:
The PR introduce ESLint and Prettier support 🍾 

* Linting is set up on a pre commit hook so it is impossible to commit code without compiling and linting it
* There are base configs for both ESLint and Prettier
* Editor extension are available for prettier so a code will be formatted on every save

Only `cli` and `lib` projects are linted at that moment and warnings are ignored. We have to try this config first and make sure we are comfortable with it. Next steps are:
* Linting tests as well
* Linting other projects
* Fixing warnings

I belive that code consistency is more important than personal preferences yet I want to be absolutely sure that you guys – @spalladino and @jcarpanelli  are comfortable with these rules. Let me know if anything feels wrong and we'll figure it out.

PS As @spalladino requested semicolons are in.


ylv-io:
Absolutely we'll remove it!

PR #924, opened 2019-06-03
Updated Deploying doc to 2.3 workflow

donntrenton:
This commit updates "Deploying your first project" (deploying.md) to the 2.3 workflow including interactive the zos create command; removed redundant instructions; revised setup instructions.


packages/docs/docs/docs/deploying.md
```diff
@@ -26,49 +24,55 @@ Use `npm` to create a `package.json` file:
 npm init
 ```
 
-This command will ask you for details about the project. For this very basic
-guide you can just press enter to accept the default values of each field.
+This command will ask you for details about the project. For this getting started project, just press **Enter** to accept the default values for each field.
```
spalladino: We can replace this by just suggesting `npm init -y`

packages/docs/docs/docs/deploying.md
```diff
@@ -26,49 +24,55 @@ Use `npm` to create a `package.json` file:
...
 
-Then, install ZeppelinOS in your project folder by running:
+Then install ZeppelinOS in your project folder by running:
 
 ```console
-npm install zos
+npm install zos@2.3.0-rc.2
```
```suggestion
npm install zos@2.3.0
```

packages/docs/docs/docs/deploying.md
```diff
@@ -26,49 +24,55 @@ Use `npm` to create a `package.json` file:
...
+Run `npx zos --version` to output the current ZeppelinOS version.
 
-Now, we can initialize the ZeppelinOS project:
+Run `npx zos --help` to output a list of all ZeppelinOS commands available. At any time
+during this getting started tutorial, you can run `npx zos <command-name> --help` to get detailed information about that command and its arguments.
+
+The contract in this tutorial imports a contract from the ZeppelinOS library package, which includes contracts to help with the development of ZeppelinOS compatible smart contracts. Install the `zos-lib` package as follows:
+
+```console
+npm install zos-lib@2.3.0-rc.2
```
```suggestion
npm install zos-lib@2.3.0
```

packages/docs/docs/docs/deploying.md
```diff
@@ -26,49 +24,55 @@ Use `npm` to create a `package.json` file:
...
-during these guides, when we introduce a new command, run
-`npx zos <command-name> --help` to get more information about that command and
-its arguments.
+Run `npx zos --version` to output the current ZeppelinOS version.
 
-Now, we can initialize the ZeppelinOS project:
+Run `npx zos --help` to output a list of all ZeppelinOS commands available. At any time
+during this getting started tutorial, you can run `npx zos <command-name> --help` to get detailed information about that command and its arguments.
+
+The contract in this tutorial imports a contract from the ZeppelinOS library package, which includes contracts to help with the development of ZeppelinOS compatible smart contracts. Install the `zos-lib` package as follows:
```
```suggestion
The contract we will be working with in this tutorial imports a contract from the ZeppelinOS library package, which includes contracts to help with the development of ZeppelinOS compatible smart contracts. Install the `zos-lib` package as follows:
```

packages/docs/docs/docs/deploying.md
```diff
@@ -7,13 +7,11 @@ The following steps will get you started using ZeppelinOS.
...
 
-First, let's install [Node.js](http://nodejs.org/) and
-[npm](https://npmjs.com/). On their
-respective websites you will find specific instructions for your machine.
+First, install [Node.js](http://nodejs.org/) and [npm](https://npmjs.com/). On the respective websites you will find specific instructions for installation on your computer.
 
 ## Setting up your project
 
-We'll need to create a directory for our project and access it:
+Create a directory for your project and navigate into it:
```
spalladino: Why the change from first plural to second singular? I don't have a strong preference for either, but I'm curious about why the change.

packages/docs/docs/docs/deploying.md
```diff
@@ -26,49 +24,55 @@ Use `npm` to create a `package.json` file:
...
 ```console
-npm install zos
+npm install zos@2.3.0-rc.2
 ```
 
-Run `npx zos --help` to get a list of all the ZeppelinOS commands available. At any time
-during these guides, when we introduce a new command, run
-`npx zos <command-name> --help` to get more information about that command and
-its arguments.
+Run `npx zos --version` to output the current ZeppelinOS version.
```
spalladino: Should we add a CTA on this, such as verifying that it matches the installed one?

packages/docs/docs/docs/deploying.md
```diff
@@ -26,49 +24,55 @@ Use `npm` to create a `package.json` file:
...
+Run `npx zos --help` to output a list of all ZeppelinOS commands available. At any time
+during this getting started tutorial, you can run `npx zos <command-name> --help` to get detailed information about that command and its arguments.
+
+The contract in this tutorial imports a contract from the ZeppelinOS library package, which includes contracts to help with the development of ZeppelinOS compatible smart contracts. Install the `zos-lib` package as follows:
+
+```console
+npm install zos-lib@2.3.0-rc.2
+```
+
+For this tutorial, you import `Initializable.sol` so your contract can use the `initializer` modifier, and thus prevent the `initialize` function from being called more than once.
```
spalladino: Readers don't know what is an initializer at this point. We may need to explain this later, or add a more complete explanation here.

packages/docs/docs/docs/deploying.md
```diff
@@ -87,112 +91,73 @@ contract MyContract is Initializable {
...
-
-```console
-npm install zos-lib
-```
-
-`zos-lib` is the ZeppelinOS library, which includes contracts to help with the
-development of ZeppelinOS compatible smart contracts. In this case, we are importing
-`Initializable.sol` in order to use the `initializer` modifier, which will
-prevent the `initialize` function from be called more than once.
+Notice that the sample contract has an `initialize` function instead of the standard
```
spalladino: We should stress that it uses the `initializer` modifier

packages/docs/docs/docs/deploying.md
```diff
@@ -87,112 +91,73 @@ contract MyContract is Initializable {
...
-npx zos session --network local --from 0x1df62f291b2e969fb0849d99d9ce41e2f137006e --expires 3600
+$ npx zos create
 ```
 
-The `session` command starts a session to work with a desired network.
-In this case, we are telling it to work with the `local` network with the
-`--network` option, and also setting a default sender address for the
-transactions we will run with the `--from` option. Additionally, the
-`expires` flag allows us to indicate the session expiration time in seconds.
+Most of the ZeppelinOS commands, such as `create`, are interactive as of version 2.3 and will prompt you for the needed information. In this case, the `create` command will prompt you, and you can use **Enter** to select the defaults as follows:
```
spalladino: Can we rephrase to avoid the repetition of _will prompt you_, which doesn't add much?

packages/docs/docs/docs/deploying.md
```diff
@@ -87,112 +91,73 @@ contract MyContract is Initializable {
...
-You can read more about this file format in the
-[configuration files](configuration.md#zos-network-json) section.
+The interactive command does the following:
 
-If a previous `zos.dev-<network_id>.json` was already created, the `push`
-command will just update said file with your last changes. For example, if you
-change the source of `MyContract` and call `zos push` again,
-ZeppelinOS will deploy a new version of `MyContract`
-and replace the entry in the json file's `contracts["MyContract"].address`.
+* Invokes `create` to compile your contract.
```
```suggestion
* Uses truffle to compile your contract.
```
Note that the `create` operation just by itself does not compile.

packages/docs/docs/docs/deploying.md
```diff
@@ -87,112 +91,73 @@ contract MyContract is Initializable {
...
 
-An important thing
-to understand is that the contracts deployed by the `push` command are logic contracts and are not intended to be used directly, rather to be used by
-upgradeable instances, as we will see later in the
-[Upgrading your project](https://docs.zeppelinos.org/docs/upgrading.html) section.
+  > **Note**: The current version of ZeppelinOS relies on Truffle to compile contracts and
+  > determine the Solidity compiler version to use. If you want to use a different version,
+  > specify it in the Truffle configuration file. You can find more information in the
+  > Truffle documentation at:
+  > https://github.com/trufflesuite/truffle/releases/tag/v5.0.0-beta.0#specify-a-solcjs-version
```
spalladino: Can we replace the URL in plain text by a link, and point to https://www.truffleframework.com/docs/truffle/reference/configuration#compiler-configuration instead?

packages/docs/docs/docs/deploying.md
```diff
@@ -87,112 +91,73 @@ contract MyContract is Initializable {
...
 
-For example, you wouldn't interact with the "MyContract" that we just deployed. This contract is only intended for defining the behavior or logic of the contract. You can find the address of this logic contract in the `zos.dev-<network_id>.json` file, under `contracts.MyContract.address`.
+* Invokes `add` to add your contract to the project by registering it in the `zos.json` configuration file.
 
-So what _do_ you interact with? As we'll see in depth in the next guide, what you do is create an upgradeable instance for the logic contract, and interact with this instance instead. 
+* Invokes `push` to deploy the contract to the `local` network. If your project had contained other contracts, they would also be deployed.
```
spalladino: The reader doesn't know what push means in this context, or why is there a difference between push and create.

spalladino: We should also add another step which is actually creating the upgradeable instance (ie the proxy) and initializing it

packages/docs/docs/docs/deploying.md
```diff
@@ -87,112 +91,73 @@ contract MyContract is Initializable {
...
 
-```console
-npx zos create MyContract --init initialize --args 42,hitchhiker
-```
+### Pushing contracts
+
+The `push` command creates a file `zos.dev-<network_id>.json` containing all the information about the project in the specified network. The file includes the address of the deployed contract implementation in `contracts.<contract-name>.address`. For more details about this file format, see
```
spalladino: Does it make sense to go in-depth in explaining the `push` command here, since the user hasn't even needed to use it? I fear this last _Pushing contracts_ section would end up being confusing to the new reader.

packages/docs/docs/docs/deploying.md
donntrenton: Made this edit.

packages/docs/docs/docs/deploying.md
donntrenton: Made this edit.

packages/docs/docs/docs/deploying.md
donntrenton: Changed to `npm init -y` and removed the subsequent paragraph, which is no longer needed.

packages/docs/docs/docs/deploying.md
donntrenton: Changed to: "The contract that you will be working with in this tutorial..."

packages/docs/docs/docs/deploying.md
donntrenton: Sorry, I probably should have discussed this Slack. Using the first plural is fine. You see it more for general articles and blogs. I recommend going with second singular for a few reasons. Second singular is a style more preferred for company content, and it's more of a standard. I think addressing text to "you" is easier and more direct for giving instructions and guidance. But ultimately both are styles to choose from.

packages/docs/docs/docs/deploying.md
donntrenton: What do you mean by 'CTA'?

packages/docs/docs/docs/deploying.md
donntrenton: Yes, that did sound repetitive. Changed to read:
"Most of the ZeppelinOS commands, such as `create`, are interactive as of version 2.3 and will request information as needed. In this case, use **Enter** to select the defaults as follows;"

packages/docs/docs/docs/deploying.md
donntrenton: Made this change.

packages/docs/docs/docs/deploying.md
donntrenton: Good suggestion; the link is now done as follows:
"You can find more information in the Truffle [Compiler Configuration](https://www.truffleframework.com/docs/truffle/reference/configuration#compiler-configuration) documentation."

packages/docs/docs/docs/deploying.md
donntrenton: Re adding another step to create an upgradeable instance--is this not done in the "Upgrading" tutorial? Let's discuss it on Slack.

packages/docs/docs/docs/deploying.md
donntrenton: I see your point, we need to find a place where it makes sense to use and explain `push`. We could move the discussion of `push` to the "Upgrading" tutorial, since a `push` step is done there. Or are you thinking we should create another step to introduce `push` in this tutorial? We should discuss this on Slack also.

packages/docs/docs/docs/deploying.md
spalladino: Call-to-action

packages/docs/docs/docs/deploying.md
donntrenton: Re this discussion of `push`, I think it's too much detail for a quickstart, but we don't have a better place to put it yet. I suggest we leave it here for now. I did remove the "Pushing contracts" heading because it seemed like it meant there was another step for the reader, and there isn't.

packages/docs/docs/docs/deploying.md
donntrenton: Re this discussion of push, I think it's too much detail for a quickstart, but we don't have a better place to put it yet. I suggest we leave it here for now. When I go through "Upgrading" maybe I can resolve this. I did remove the "Pushing contracts" heading because it seemed like it meant there was another step for the reader, and there really isn't.

packages/docs/docs/docs/deploying.md
donntrenton: Sure, we could add "Make sure the current version matches the version you installed."

spalladino:
Closing in favor of #961

PR #928, opened 2019-06-04
Fix directory path replacement

jcarpanelli:
Fixes https://github.com/zeppelinos/zos/issues/927. The main problem was that the previous regexp was not precise enough to cover some edge (well, not precisely edge) cases  if having a project directory path such as: `foo/bar/bazcontracts`, or anything with a `contracts` at the end of the directory name.
Now, we are using `path#dirname` instead (thanks @ylv-io!!)
If approved, I will add this to the `release/2.3` branch.


ylv-io:
🚳 
[path.dirname](https://nodejs.org/api/path.html#path_path_dirname_path)
Or I am missing something?

jcarpanelli:
> path.dirname
> Or I am missing something?

I always forget about that one! thanks @ylv-io

PR #934, opened 2019-06-05
Handles swarm hash metadata for solidity 0.5.2

### LINKED ISSUE #844: Handle new CBOR metadata format
Solidity is now storing the compiler version along with the swarm hash at the end of the contract (see [here](https://github.com/ethereum/solidity/pull/4764/commits/aa4ee3a1559ebc0354926af962efb3fcc7dc15bd)). We need to make sure that zOS properly handles this format when a new solidity compiler version is released.

We can also try extracting the compiler version from the hash to automate the `zos verify` process.

Related to https://github.com/zeppelinos/zos/issues/821.

spalladino:
Since Solidity 0.5.2, the swarm hash is included along with the compiler version, CBOR encoded at the end of the runtime bytecode. This handles both metadata versions and removes the swarm hash from both of them.

Fixes #844

### DIFF
--- a/packages/lib/src/utils/Bytecode.ts
+++ b/packages/lib/src/utils/Bytecode.ts
@@ -46,7 +46,7 @@ export function tryRemoveMetadata(bytecode: string): string {
   const length = parseInt(rawLength, 16);
 
   // Bail on unreasonable values for length (meaning we read something else other than metadata length)
-  if (length > 200) return bytecode;
+  if (length * 2 > bytecode.length - 4) return bytecode;
 
   // Gather what we assume is the CBOR encoded metadata, and try to parse it
   const metadataStart = bytecode.length - length * 2 - 4;
--- a/packages/lib/src/utils/Bytecode.ts
+++ b/packages/lib/src/utils/Bytecode.ts
@@ -1,6 +1,10 @@
 import crypto from 'crypto';
 import { ZERO_ADDRESS } from './Addresses';
 import Contract from '../artifacts/Contract';
+import cbor from 'cbor';
+import Logger from '../utils/Logger';
+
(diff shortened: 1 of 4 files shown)


spalladino:
Please hold off any review, I noted there are a few more changes I should squeeze in (basically, removing the whole CBOR metadata instead of just the swarm hash).

packages/lib/src/utils/Bytecode.ts
```diff
@@ -31,11 +35,33 @@ export function hasUnlinkedVariables(bytecode: string): boolean {
...
+export function tryRemoveMetadata(bytecode: string): string {
+  // Bail on empty bytecode
+  if (!bytecode || bytecode.length <= 2) return bytecode;
+
+  // Gather length of CBOR metadata from the end of the file
+  const rawLength = bytecode.slice(bytecode.length - 4);
+  const length = parseInt(rawLength, 16);
+
+  // Bail on unreasonable values for length (meaning we read something else other than metadata length)
+  if (length > 200) return bytecode;
```
ylv-io: Why 200? I would either remove this check or replace it with `length > (bytecode.length - 4)`. The reason is:

> The CBOR mapping can also contain other keys, so it is better to fully decode the data instead of relying on it starting with 0xa265. For example, if any experimental features that affect code generation are used, the mapping will also contain "experimental": true.

CBOR mapping can be way bigger than 200.

ylv-io: And if we read something else we'll fail on decoding.

spalladino: > And if we read something else we'll fail on decoding.

Good point!

PR #936, opened 2019-06-07
Renaming design doc

spalladino:
Please comment on this PR to define the new naming structure.

### DIFF
--- a/design/renaming.md
+++ b/design/renaming.md
@@ -30,7 +30,9 @@ We need to define how we want to support the migration from old to new config fi
 
 | Current | New |
 |---|---|
-| **zos** init MyProject | **oz** init MyProject |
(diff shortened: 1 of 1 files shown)


design/renaming.md
```diff
@@ -0,0 +1,59 @@
...
+| Current | New |
+|---|---|
+| packages/cli | packages/cli |
+| packages/lib | packages/lib |
+
+## Commands
+
+| Current | New |
+|---|---|
+| **zos** init MyProject | **oz** init MyProject |
```
spalladino: - Support `zos` with a deprecation notice 
- Support both `oz` and `openzeppelin`

In all messages and docs, use `openzeppelin` instead of `oz`

PR #947, opened 2019-06-10
Handle relative and absolute paths in isLocalContract

### LINKED ISSUE #945: Error when running zos create without added contracts
On master, running `zos create` without any contracts on zos.json yields the error "At least one contract name must be provided to add". Most likely the cause is #940, that messed up the `isLocalContract` check.

spalladino:
Fixes #945

### DIFF
--- a/packages/cli/src/models/local/ContractManager.ts
+++ b/packages/cli/src/models/local/ContractManager.ts
@@ -2,6 +2,7 @@ import { Contracts, Contract, FileSystem } from 'zos-lib';
 import Dependency from '../dependency/Dependency';
 import ZosPackageFile from '../files/ZosPackageFile';
 import ConfigManager from '../config/ConfigManager';
+import path from 'path';
 
 export default class ContractManager {
   public packageFile: ZosPackageFile;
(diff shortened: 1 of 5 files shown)


packages/cli/src/models/local/ContractManager.ts
```diff
@@ -51,11 +53,13 @@ export default class ContractManager {
...
   private isLocalContract(
-    buildDir: string,
-    contract: { [key: string]: any },
+    contractsDir: string,
+    contract: { sourcePath: string },
+    root: string,
   ): boolean {
-    const projectDir = buildDir.replace('build/contracts', '');
-    return contract.sourcePath.indexOf(projectDir) === 0;
+    const cwd = root || process.cwd();
```
jcarpanelli: nit: you can define `process.cwd()` as the default value in the `isLocalContract` function definition

packages/cli/src/models/local/LocalController.ts
```diff
@@ -96,29 +97,8 @@ export default class LocalController {
...
-          );
-
-        if (isProjectContract && !isLibrary) {
-          const contractName = contractData.contractName;
-          this.add(contractName, contractName);
-        }
-      }
-    });
+    const manager = new ContractManager(this.packageFile);
+    manager.getContractNames().forEach(name => this.add(name, name));
```
jcarpanelli: Great! :raised_hands:

PR #957, opened 2019-06-11
Fix zepkit integration test

spalladino:
When running zepkit tutorial integration tests, compilation would fail due to truffle trying to compile OZ contracts (that require solidity 0.5.2) with solc 0.5.0. Upgrading the truffle version to 5.0.21, which sets the default solc to 0.5.8, fixes the issue.


tests/zepkit/test/package.json
```diff
@@ -12,6 +12,6 @@
     "chai": "^4.1.2",
     "lodash": "^4.17.0",
     "mocha": "5.2.0",
-    "truffle": "^5.0.19"
+    "truffle": "5.0.21"
```
jcarpanelli: Should we pin this version?

jcarpanelli: Oh, it's already pinned :joy:

PR #963, opened 2019-06-12
Handle compile commands when there are no sources

### LINKED ISSUE #962: Compile fails if there are no contracts
Running `zos compile` on an empty project with no contracts fails with error:
```
$ zos compile
Compiling contracts with solc 0.5.9 (commit.e560f70d)...
Using local solc compiler found
Compilation errors: 
No input sources specified.
```

Note that this also happens if the user attempts to run `zos create` with no contracts but an EVM package linked.

spalladino:
Fixes #962

### DIFF
--- a/packages/cli/src/models/compiler/solidity/SolidityProjectCompiler.ts
+++ b/packages/cli/src/models/compiler/solidity/SolidityProjectCompiler.ts
@@ -66,7 +66,12 @@ class SolidityProjectCompiler {
     await this._loadDependencies();
 
     if (this.contracts.length === 0) {
-      log.info('No contracts found to compile.');
+      Loggy.noSpin(
+        __filename,
+        'call',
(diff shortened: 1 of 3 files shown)


packages/cli/src/models/compiler/solidity/SolidityProjectCompiler.ts
```diff
@@ -66,6 +66,12 @@ class SolidityProjectCompiler {
   public async call(): Promise<void> {
     await this._loadSoliditySourcesFromDir();
     await this._loadDependencies();
+
+    if (this.contracts.length === 0) {
+      log.info('No contracts found to compile.');
```
jcarpanelli: This line will break the `master` branch. I will push a fix into this same branch if you don't mind :smiley:, and include it in `v2.4.0-rc.1`

PR #948, opened 2019-06-10
Enhance zos outputs

jcarpanelli:
This PR implements several improvements to `zos` outputs. Some of them are:
- Improvements in log messages texts.
- Hide unnecessary logs on non-verbose mode.
- Implement multispinners using [spinnies](https://github.com/jcarpanelli/spinnies), a super handy and fancy lib for managing multiple spinners in command-line interface programs :stuck_out_tongue:.
- Reimplemented the `verbose` mode, adding a timestamp plus the name of the function/method where the log is being printed

**Bonus track**:

I removed the `StatusChecker`, `StatusComparator` and `StatusFetcher` together with the `pull`, `fetch` and `status` scripts. I also added an error message in the `status` command notifying that it's deprecated.

**TODO**
- ~I removed all references to the old `Logger` methods (warn, info, error). I still have to remove all the unused imports, the `Logger.silent(true)` in some config files and also have to refactor a bit the `unpack` command to work directly with the new logger instead of using the `spinner` utils.~
- ~Still have to finish fixing the `CaptureLogs` class~
- ~Remove `stdouts` that we discussed offline.~

A mini demo of how its working (running `zos create` before pushing contracts + pushing an EVM-package locally)

Before:
before

After:
after

PR #969, opened 2019-06-13
Set logger in silent mode by default

jcarpanelli:
While preparing everything to release `v2.4.0-rc.1` I noticed that `zos` was not in silent mode in some examples tests after adding a `Loggy.silent(false)` before merging #948. This PR fixes that.

### DIFF
--- a/examples/creating-instances-from-solidity/index.js
+++ b/examples/creating-instances-from-solidity/index.js
@@ -15,7 +15,6 @@ const fs = require('fs');
 // Import ZeppelinOS `zos` and `zos-lib` dependencies.
 const zos = require('zos');
 const lib = require('zos-lib');
-lib.Loggy.silent(true);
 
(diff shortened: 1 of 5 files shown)


spalladino:
@jcarpanelli I'm not sure I follow, as this PR should not be necessary. `Loggy` should be silent by default, unless explicitly activated. And we should be activating it only on `packages/cli/src/bin/zos-cli.ts`. Am I missing anything?

jcarpanelli:
@spalladino It seems that the one that wasn't following was me. I rollbacked the changes implemented on the examples and set the `isSilent` value as `true` (and also set the the isVerbose as false)

Finally, I added a super small refactor that I forgot to implement in the `Verifier`  model (sorry for including this here!)

spalladino:
Cool, thanks Jota!!

PR #976, opened 2019-06-15
Improve parsing of interactive args

spalladino:
Parses arguments to call a method (such as in the call, send-tx, update, or create commands) according to the argument type. 

In order to accepts arrays with elements with and without quotes (so something simple like `foo,bar` would work, without requiring the user to input `["foo","bar"]`), I wrote a simple parser by hand. Please take a good look at it.

I'm also merging the PR onto `release/2.4` directly to publish a new RC, but I'd appreciate if either @jcarpanelli or @ylv-io could review this PR after merged in case something slipped through.

### DIFF
--- a/packages/cli/src/utils/input.ts
+++ b/packages/cli/src/utils/input.ts
@@ -1,40 +1,16 @@
 import BN from 'bignumber.js';
-import { encodeParams } from 'zos-lib';
+import flattenDeep from 'lodash.flattendeep';
+import { encodeParams, Loggy } from 'zos-lib';
 
+// TODO: Deprecate in favor of a combination of parseArg and parseArray
 export function parseArgs(args: string): string[] | never {
   if (typeof args !== 'string') throw Error(`Cannot parse ${typeof args}`);
 
   // The following is inspired from ethereum/remix (Copyright (c) 2016)
   // https://github.com/ethereum/remix/blob/89f34fc4f35eca0c386379550c300205d35bfae0/remix-lib/src/execution/txFormat.js#L51-L53
   try {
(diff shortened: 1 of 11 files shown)


packages/cli/src/utils/input.ts
```diff
@@ -49,6 +25,199 @@ export function parseArgs(args: string): string[] | never {
...
+  // Replace non quoted number by quoted number.
+  const MATCH_WORDS = new RegExp(
+    START_LOOKBEHIND + '([-]?\\w+)' + END_LOOKAHEAD,
+    'g',
+  );
+  args = args.replace(MATCH_WORDS, '"$2"');
+  return args;
+}
+
+export function parseArg(input: string | string[], type: string): any {
```
spalladino: We are missing tests for this function

packages/cli/src/utils/input.ts
```diff
@@ -49,6 +25,199 @@ export function parseArgs(args: string): string[] | never {
...
+  // Booleans: match against true, t, yes, 1, etc
+  else if (type === 'bool' && requireInputString(input)) {
+    const lowercaseInput = input.toLowerCase();
+    if (TRUE_VALUES.includes(lowercaseInput)) return true;
+    else if (FALSE_VALUES.includes(lowercaseInput)) return false;
+    else throw new Error(`Could not parse boolean value ${input}`);
+  }
+
+  // Arrays: recursively parse as needed
+  else if (type.match(ARRAY_TYPE_REGEX)) {
```
spalladino: This one should be the first case to try to match. Otherwise, `startsWith(uint)` will match `uint[]`.

PR #985, opened 2019-06-18
Fix Verifier when using custom config

jcarpanelli:
Fixes https://github.com/zeppelinos/zos/issues/944.

The verifier is not working though. It's probably related to how we are picking the default compiler options in `cli/src/commands/verify.ts`.

### DIFF
--- a/packages/lib/src/utils/Solidity.ts
+++ b/packages/lib/src/utils/Solidity.ts
@@ -1,6 +1,5 @@
 import flatten from 'truffle-flattener';
 
-// TS-TODO: contract might be typed with some sort of web3 typed lib.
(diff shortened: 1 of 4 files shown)


packages/lib/src/utils/Solidity.ts
```diff
@@ -1,6 +1,9 @@
 import flatten from 'truffle-flattener';
 
 // TS-TODO: contract might be typed with some sort of web3 typed lib.
```
spalladino:
```suggestion
```

spalladino: We can now kill that TODO ;-)

PR #988, opened 2019-06-18
Improve inputs and outputs text messages

jcarpanelli:
Fixes https://github.com/zeppelinos/zos/issues/960.
- Implemented Manu's suggestions in both inputs and outputs.
- Updated spinnies to `v0.4.1`, which handles spinners texts with line breaks properly
- Improve the message logged when creating an instance calling a function that does not receive arguments.

### DIFF
--- a/packages/cli/src/models/network/NetworkController.ts
+++ b/packages/cli/src/models/network/NetworkController.ts
@@ -300,7 +300,7 @@ export default class NetworkController {
         __filename,
         'uploadContract',
         `upload-contract${contract.schema.contractName}`,
-        `Validating and pushing logic contract ${contract.schema.contractName}`,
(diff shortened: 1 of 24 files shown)


packages/cli/src/models/network/NetworkController.ts
```diff
@@ -318,7 +318,7 @@ export default class NetworkController {
 
       Loggy.succeed(
         `upload-contract${contract.schema.contractName}`,
-        `Logic contract ${contract.schema.contractName} successfully pushed`,
+        `Logic contract ${contract.schema.contractName} successfully deployed`,
```
spalladino: Remove this mention to logic contract if we are removing it from the previous one as well

Issue #931, opened 2019-06-05
Restructure CLI configuration files


spalladino:
With the upcoming [rename](https://github.com/zeppelinos/zos/issues/925), we have the opportunity to rethink how/where the CLI configuration files are set. We have received feedback that **the CLI can be annoying by cluttering the project root with too many files**, such as:
* zos.json
* zos.kovan.json
* zos.mainnet.json
* zos.rinkeby.json
* zos.ropsten.json
* zos.dev-NNNN.json
* .zos.lock
* .zos.session

With this in mind, we can consider moving **all of the files above to a `.openzeppelin` folder**, similar to what `.circleci` does. We can do this along with the renaming defined in #936. The new files should also **handle `manifestVersion` instead of `zosversion`**.

The new `networks.js` file should still be placed the project root, since it's designed to be framework-independent.

We need to define the migration path for this. We can either:
1. Support both of them. If there is a zos.json, use the old structure. If not, use the new one.
2. Automatically move all files to the new structure whenever a command is run, as a migration.

crazyrabbitLTC:
This would simplify a couple of "house keeping" tasks that occasionally come up- the process of removing old zos.<network>.json files from development networks. If you consistently run the same project over and over, you end up with an ever expanding list of a json files in your main project directory.

Would it make any sense to keep build artifacts along side the zos.<network>.json file? The reason being that deploying different version of contracts to different networks is not currently possible as while there can be numerous zos.<network>.json files, there seems like we can only have one build directory. I guess this implies also needing some sort of versioning for the contracts themselves, but perhaps it's worthwhile to keep in mind for the future?

PR #1000, opened 2019-06-19
Rename ZosPackageFile, ZosNetworkFile, ZosVersion and zosversion

ylv-io:
How funny it is that PR to rename ZOS is PR number 1000. I'd say it is a good luck.


examples/cli-create2/index.js
```diff
@@ -47,7 +47,7 @@ async function useSignedCreate2(initValue, networkConfig) {
   const initData = Sample.methods.initialize(initValue).encodeABI();
 
   // We will use the project's proxy admin as upgradeability admin of this instance
-  const networkFile = new files.ZosPackageFile().networkFile(networkConfig.network);
+  const networkFile = new files.ProjectFile().networkFile(networkConfig.network);
```
jcarpanelli: the noun `project` is more accurate than `Package`. Loved it!

packages/cli/src/models/files/ProjectFile.ts
```diff
@@ -58,7 +58,7 @@ export default class ZosPackageFile {
       }.`;
       throw e;
     }
-    checkVersion(this.data.zosversion, this.fileName);
+    checkVersion(this.data.manifestVersion, this.fileName);
```
jcarpanelli: Are we still going to support reading the `zosversion` attribute after these changes for users that have already started their projects with `zos` versions below `2.4.0`?

ylv-io: I don't think we should support reading old `zosversion`. I belive we should migrate them to new `manifestVersion`. @spalladino What do you think?

ylv-io:
This is pretty big PR and I don't want to add more stuff on top of it. Shall we merge it into a master even it doesn't have migration strategy for old `zosversion`? @jcarpanelli @spalladino ?

packages/cli/src/models/files/ProjectFile.ts
spalladino: Hmm I guess we should support `zosversion` for `zos.json` files, and `manifestVersion` for `.openzeppelin/project.json` files.

spalladino:
> Shall we merge it into the master even it doesn't have migration strategy for old zosversion? @jcarpanelli @spalladino ?

I'm fine with that @ylv-io! Make sure you keep in mind that the migration strategy is pending. We can handle it together with #931.

PR #1002, opened 2019-06-19
Fix compiler logs

jcarpanelli:
As seen offline together with @spalladino, the compiler was printing some `undefined` messages on verbose mode. this PR fixes that.


packages/cli/src/models/compiler/solidity/CompilerProvider.ts
```diff
@@ -205,7 +205,7 @@ async function downloadCompiler(
   localFile: string,
 ): Promise<void> {
   const { version, keccak256: expectedHash, path: versionPath } = build;
-  Loggy.spin(
+  Loggy.onVerbose(
```
spalladino: Why this change? Downloading a compiler is a long-running operation, I think it makes sense to display a spinner for it.

PR #1003, opened 2019-06-19
Improve outputs and move minimal proxy creation warn message to command

jcarpanelli:
Fixes https://github.com/zeppelinos/zos/issues/989. Fixes https://github.com/zeppelinos/zos/issues/965.

Also: 
- Changed `update` references to `upgrade`
- Changed `update` command to `upgrade`, and added `update` as alias (I didn't migrate the filenames though)
- Sort function names list in prompt (initializables first, abusing of `lodash` `sortBy` functionality, which lists the strings with special characters first)
- Implement wording fixes reviewed with Manu

### DIFF
--- a/packages/cli/src/commands/update.ts
+++ b/packages/cli/src/commands/update.ts
@@ -18,15 +18,15 @@ import {
 } from '../prompts/prompt';
 import promptForMethodParams from '../prompts/method-params';
 
-const name = 'update';
+const name = 'upgrade';
 const signature = `${name} [alias-or-address]`;
 const description =
-  'update contract to a new logic. Provide the [alias] or [package]/[alias] you added your contract with, its [address], or use --all flag to update all contracts in your project.';
+  'upgrade contract to a new logic. Provide the [alias] or [package]/[alias] you added your contract with, its [address], or use --all flag to upgrade all contracts in your project.';
 
 const register: (program: any) => any = program =>
   program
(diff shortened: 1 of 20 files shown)


packages/cli/src/commands/update.ts
```diff
@@ -18,15 +18,15 @@ import {
 } from '../prompts/prompt';
 import promptForMethodParams from '../prompts/method-params';
 
-const name = 'update';
+const name = 'upgrade';
```
spalladino: About time we implemented this change!

packages/cli/src/prompts/prompt.ts
```diff
@@ -220,6 +221,8 @@ export function methodsList(
       return { name: label, value: { name, selector } };
     },
   );
+
+  return sortBy(methods, ['name']);
```
spalladino: A name starting with a dollar sign [is a valid identifier in Solidity](https://github.com/solidityj/solidity-antlr4/blob/edce5c44d2d681e32bb23034d822bf0fcf1cb43e/Solidity.g4#L419-L421), and `$ < *` lexicographically, so any functions starting with dollar would be listed _before_ the initializable ones.

```js
> ['*', '$', 'A', 'a'].sort()
[ '$', '*', 'A', 'a' ]
```

jcarpanelli: ups, good catch!

PR #1018, opened 2019-06-19
Validate method args in interactive prompt

### LINKED ISSUE #999: Passing array arg is to `zos create` is not obvious
image

I couldn't figure out how to pass array arg to zos create. I've also checked docs and there is no example there.

spalladino:
Use the arg type and inquirer validate callback to check that the user input is correct. If not, we show an example on how to input the value.

Screenshot from 2019-06-19 17-31-02

Fixes #999

### DIFF
--- a/packages/cli/src/prompts/method-params.ts
+++ b/packages/cli/src/prompts/method-params.ts
@@ -97,7 +97,7 @@ function getCommandProps(
               parseArg(input, arg.type);
               return true;
             } catch (err) {
-              return `Error parsing ${input} (${err.message}). Enter a valid ${
+              return `${err.message}. Enter a valid ${
                 arg.type
               } such as: ${getPlaceholder(arg.type)}.`;
             }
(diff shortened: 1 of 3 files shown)


packages/cli/src/prompts/method-params.ts
```diff
@@ -125,3 +135,30 @@ function getCommandProps(
...
   };
 }
+
+function getPlaceholder(type: string): string {
+  const ARRAY_TYPE_REGEX = /(.+)\[\d*\]$/; // matches array type identifiers like uint[] or byte[4]
+
+  if (type.match(ARRAY_TYPE_REGEX)) {
+    const arrayType = type.match(ARRAY_TYPE_REGEX)[1];
+    const itemPlaceholder = getPlaceholder(arrayType);
+    return `[${itemPlaceholder}, ${itemPlaceholder}]`;
```
jcarpanelli: Can we show the simplified version of array arguments? (I recall that you implemented the `parseArg` function to work fine without using `[]`, right?)

spalladino: Yeah, but that fails when you have nested arrays. This one works always, so I decided to go with it.

PR #1033, opened 2019-06-21
Handle relative imports from project root in compiler

### LINKED ISSUE #1024: Import paths relative to root folder not supported by compiler
Given the following files in the root folder:
```
contracts/Folder/C1.sol
contracts/Folder/C2.sol
```

Using the import `import "contracts/Folder/C2.sol"` in C1 works in truffle compiler but fails in zos compile. If this is a standard syntax, we need to support it.

spalladino:
Support the syntax `import "contracts/Folder/C2.sol"` in a contract `contracts/Folder/C1.sol`.

Fixes #1024

### DIFF
--- a/packages/cli/mocks/mock-project-with-root-imports/zos.json
+++ b/packages/cli/mocks/mock-project-with-root-imports/zos.json
@@ -1,0 +1,9 @@
+{
+  "name": "mock-project-with-root-imports",
+  "zosversion": "2.2",
+  "version": "1.1.0",
+  "contracts": {
+    "Greeter": "GreeterImpl"
+  },
+  "dependencies": {}
+}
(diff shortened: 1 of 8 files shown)


packages/cli/mocks/mock-project-with-root-imports/zos.json
```diff
@@ -0,0 +1,9 @@
+{
+  "name": "mock-project-with-root-imports",
+  "zosversion": "2.2",
```
jcarpanelli: We should remember to migrate these `zosversion` references before merging this commit into master, as https://github.com/zeppelinos/zos/pull/1000 has already been merged.

spalladino: Good catch!

PR #1036, opened 2019-06-21
Constants still work the normal way with zos

pcowgill:
Related to https://forum.zeppelin.solutions/t/is-it-ok-to-define-constants-in-upgradable-contracts-with-zeppelinos/892/2


spalladino:
Awesome @pcowgill, thanks so much for the contribution!!

Issue #1021, opened 2019-06-19
Events not being showed properly on `zos send-tx`


jcarpanelli:
In some cases, for instance when `zos send-tx`ing a function that creates a contract which emits an event, the `events.ts#describeEvents` function is not properly parsing the list of emitted events.

spalladino:
Related to #1104

PR #1037, opened 2019-06-21
Avoid showing external contract emitted events

jcarpanelli:
Related to #1021. Doesn't solve it, but Now we can remove the `bug` flag :stuck_out_tongue:

### DIFF
--- a/packages/cli/src/commands/set-admin.ts
+++ b/packages/cli/src/commands/set-admin.ts
@@ -147,7 +147,7 @@ function getCommandProps({
   return {
     ...networksList('network', 'list'),
     pickProxyBy: {
-      message: 'For which proxies would you like to transfer ownership?',
+      message: 'For which instances would you like to transfer ownership?',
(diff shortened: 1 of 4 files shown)


packages/cli/src/commands/set-admin.ts
```diff
@@ -151,7 +151,7 @@ function getCommandProps({
       type: 'list',
       choices: [
         {
-          name: 'All proxies',
+          name: 'All instances',
```
spalladino: We need to update the message above as well: `For which proxies would you like to transfer ownership?`

packages/cli/src/commands/update.ts
```diff
@@ -145,7 +145,7 @@ function getCommandProps({
       type: 'list',
       choices: [
         {
-          name: 'All proxies',
+          name: 'All instances',
```
spalladino: Same here! `Which proxies would you like to upgrade?`

PR #1040, opened 2019-06-24
 Rename Project, Network, Sesion and Lock files

ylv-io:
The PR renames config files according to:

Current | New
-- | --
zos.json | .openzeppelin/project.json
zos.rinkeby.json | .openzeppelin/rinkeby.json
.zos.lock | .openzeppelin/.lock
.zos.session | .openzeppelin/.session

While keeping backward compatibility to old config files. Every new config file will be created using new naming convention but old config won't be migrated and will be used as it is.
We are hopping to break legacy config file support during release of v3.0.


packages/cli/src/models/files/ProjectFile.ts
```diff
@@ -22,53 +24,72 @@ interface ConfigFileCompilerOptions {
...
+      manifestVersion: MANIFEST_VERSION,
+    } as any;
+    // if project file path is provided use only it
+    if (filePath) {
+      this.filePath = filePath;
+      this.data = this.tryToLoadProject(filePath);
+    } else {
+      // try to load legacy file first
+      // TODO: remove legacy project file support as a breaking change at 3.0
+      let data = this.tryToLoadProject(LEGACY_PROJECT_FILE_NAME);
```
spalladino: I'm thinking about whether we should prioritize the legacy or the new format. What would be a scenario where both files are present? For instance, I'd expect an EVM package to use both: zos.json for supporting old versions of zOS, and the new format for new versions. If so, it would make sense to try the new format first, right?

packages/cli/test/models/Dependency.test.js
```diff
@@ -29,7 +29,7 @@ contract('Dependency', function([_, from]) {
         it('throws error', function() {
           assertErrorMessage(
             () => Dependency.fromNameWithVersion('bildts-kcom'),
-            /Could not find a zos.json file/,
+            /Could not find a project.json file/,
```
spalladino: Should we change this error message to `an .openzeppelin/project.json file`?

tsconfig.json
```diff
@@ -14,7 +14,7 @@
     ]
   },
   "exclude": [
-     "node_modules",
+     "**/node_modules/*",
```
spalladino: Good catch!

packages/cli/test/models/Dependency.test.js
ylv-io: No opinion on that at all.

packages/cli/src/models/dependency/Dependency.ts
```diff
@@ -59,17 +63,21 @@ export default class Dependency {
 
   public static hasDependenciesForDeploy(network: string): boolean {
     const dependencies = ProjectFile.getLinkedDependencies() || [];
-    const networkDependencies =
-      NetworkFile.getDependencies(`zos.${network}.json`) || {};
```
jcarpanelli: We can remove the `getDependencies` static method at `NetworkFile.ts` as we are not using it anymore

packages/cli/src/models/files/ProjectFile.ts
```diff
@@ -22,53 +24,68 @@ interface ConfigFileCompilerOptions {
...
-      )}' file. Please make sure that ${fileName} is a valid JSON file. Details: ${
-        e.message
-      }.`;
-      throw e;
     }
-    checkVersion(this.data.manifestVersion, this.fileName);
+    return [];
+  }
+
+  public constructor(filePath: string = null) {
```
jcarpanelli:
```suggestion
  constructor(filePath: string = null) {
```

packages/cli/src/models/files/ProjectFile.ts
```diff
@@ -22,53 +24,68 @@ interface ConfigFileCompilerOptions {
...
+    const defaultData = {
+      manifestVersion: MANIFEST_VERSION,
+    } as any;
+    this.filePath = ProjectFile.getExistingFilePath('.', filePath);
```
jcarpanelli: What about changing `ProjectFile.getExistingFilePath('.', filePath);` with `ProjectFile.getExistingFilePath(process.cwd(), filePath);` or setting the default value of the `dir` argument as `process.cwd()` in the `getExistingFilePath` function definition?

packages/cli/src/models/files/NetworkFile.ts
```diff
@@ -96,30 +97,40 @@ export default class NetworkFile {
   }
 
   // TS-TODO: type for network parameter (and class member too).
-  public constructor(projectFile: ProjectFile, network: any, fileName: string) {
+  public constructor(
```
jcarpanelli:
```suggestion
  constructor(
```

packages/cli/src/bin/zos-cli.ts
```diff
@@ -16,12 +22,13 @@ if (!IGNORED_COMMANDS_IN_ROOT_DIR.includes(command)) {
...
-const LOCKFILE = '.zos.lock';
+
 try {
-  lockSync(LOCKFILE, { retries: 0 });
+  fs.ensureDirSync(OPEN_ZEPPELIN_FOLDER);
+  lockSync(LOCK_FILE_PATH, { retries: 0 });
 } catch (e) {
   console.error(
-    `Cannot run more than one instance of 'zos' at the same time.\nIf you are sure that no other instances are actually running, manually remove the file ${LOCKFILE} and try again.`,
+    `Cannot run more than one instance of 'openzeppelin' at the same time.\nIf you are sure that no other instances are actually running, manually remove the file ${LOCK_FILE_PATH} and try again.`,
```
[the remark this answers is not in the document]
ylv-io: No idea. I am fine with any.

packages/cli/src/models/files/ProjectFile.ts
ylv-io: Default value looks as a better path.

packages/cli/src/models/files/ProjectFile.ts
ylv-io: Having explicit constructor is a [typescript-eslint](https://github.com/typescript-eslint/typescript-eslint/blob/v1.9.0/packages/eslint-plugin/docs/rules/no-explicit-any.md).

jcarpanelli:
https://github.com/zeppelinos/zos/pull/1054 Already merged. To master it goes! :100:

PR #1048, opened 2019-06-25
Update linking.md

amystrayer:
I don't know why some notes were fully deleted then rewritten exactly. I didn't modify these.
In the last paragraph, it says "allowing us to spin up a new token". Is "spin up" normally used in this way in this subject? Or do you mean "create a new token"?

### DIFF
--- a/packages/docs/docs/docs/linking.md
+++ b/packages/docs/docs/docs/linking.md
@@ -3,11 +3,11 @@ id: linking
 title: Linking OpenZeppelin contracts
 ---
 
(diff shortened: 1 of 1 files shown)


spalladino:
> I don't know why some notes were fully deleted then rewritten exactly. I didn't modify these.

That's odd. No worry about them though.

> In the last paragraph, it says "allowing us to spin up a new token". Is "spin up" normally used in this way in this subject? Or do you mean "create a new token"?

Yeah! Spin up is mostly used with servers or processes, rather than contracts, but it still applies.

PR #1054, opened 2019-06-26
Merge release/2.4 fixes into master

jcarpanelli:
I also added fixed some residual references to the deprecated `zosversion`. 

*Edit*: this PR should be merged and not squashed.


ylv-io:
Why are we committing `examples/linking-contracts/build/`?

jcarpanelli:
> Why are we committing examples/linking-contracts/build/ ?

Good question. Is there any reason behind this @spalladino? Or should I add it to the `.gitignore`?

spalladino:
Good point. Not sure if the CI will work fine if we remove it though. If we remove it and it still works, I'm fine with it.

jcarpanelli:
Seems to work fine! To master it goes then!

PR #1063, opened 2019-07-01
Adds frictionless support for git hooks and linting

ylv-io:
* Use husky to run git hooks – no need to install anything.
* Use lint-staged to lint only changed files not entire project
* Compiles entire project using TypeScript on each commit

### DIFF
--- a/.githooks/pre-commit
+++ b/.githooks/pre-commit
@@ -1,2 +1,0 @@
-#!/bin/sh
-./node_modules/.bin/lerna run lint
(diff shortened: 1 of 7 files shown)


ylv-io:
Yes. No action is required. You might need to drop your .git/hooks/precommit hook because husky wouldn't overwrite existing hook but for fresh checkouts it would work out of the box.

ylv-io:
I'll test it myself after merge.

PR #1074, opened 2019-07-04
Rename: zos-lib contracts and zos command

jcarpanelli:
This PR implements:
- Renames in `zos-lib` contract references:
 
| Current | New |
|---|---|
| ZOSLibOwnable | OpenZeppelinUpgradesOwnable |
| ZOSLibECDSA | OpenZeppelinUpgradesECDSA |
| ZOSLibAddress | OpenZeppelinUpgradesAddress |

- Renames both implementation & admin slots in contracts and `zos-lib` contracts, giving support for those proxies already created with the old slot labels references.
- Rename `zos-cli.ts` to `oz-cli.ts`. Also changed references in `cli/package.json` files
- Add `openzeppelin` and `oz` as valid binaries, and added `zos` and `oz` as commander aliases.
- Rename `bin/zos-cli.js` references in zepkit integration tests and `first-project` sample.
- Add deprecation warning when using `zos` instead of `oz` or `openzeppelin`

~Tests are failing but they should be not very difficult to fix, as almost all are throwing the same error.~


packages/lib/test/contracts/ProxyAdmin.test.js
```diff
@@ -40,7 +38,7 @@ contract('ProxyAdmin', function(accounts) {
       this.ownable = this.proxyAdmin;
     });
 
-    shouldBehaveLikeOwnable(proxyAdminOwner, anotherAccount);
+    // shouldBehaveLikeOwnable(proxyAdminOwner, anotherAccount);
```
ylv-io: 👀  Why is that?

jcarpanelli: Ha! nice catch. I left it commented while debugging :cry:

packages/lib/test/contracts/upgradeability/AdminUpgradeabilityProxy.behaviour.js
```diff
@@ -53,6 +54,21 @@ export default function shouldBehaveLikeAdminUpgradeabilityProxy(
     this.proxyAddress = this.proxy.address;
   });
 
+  afterEach(function () {
+    sinon.restore();
+  });
+
+  it('uses the correct admin storage slot', async function() {
```
spalladino: Note that this test does not check that the Proxy class falls back to the deprecated label. I'd suggest adding a `LegacyProxy` contract in the mocks folder, and testing against it to ensure both the storage and admin are correctly retrieved.

packages/lib/test/contracts/upgradeability/UpgradeabilityProxy.behaviour.js
```diff
@@ -46,6 +54,16 @@ export default function shouldBehaveLikeUpgradeabilityProxy(
     it('has expected balance', async function() {
       (await ZWeb3.getBalance(this.proxy)).should.eq(balance.toString());
     });
+
+    it('uses the correct implementation storage slot', async function() {
```
spalladino: Same for this test

packages/lib/test/contracts/upgradeability/AdminUpgradeabilityProxy.behaviour.js
jcarpanelli: Yeap, I'm working on it. I created a `zos` folder inside the `mock` contracts folder and added the `Zos` prefix to the upgradeability contract names. Is that fine?

jcarpanelli:
@spalladino should be ready for another review.

packages/lib/test/contracts/upgradeability/storageSlot.test.js
```diff
@@ -0,0 +1,70 @@
...
+import sinon from 'sinon';
+import { toBN, toHex } from 'web3-utils';
+
+import Proxy from '../../../src/proxy/Proxy';
+import ZWeb3 from '../../../src/artifacts/ZWeb3';
+import Contracts from '../../../src/artifacts/Contracts';
+
+const DummyImplementation = Contracts.getFromLocal('DummyImplementation');
+
+export function shouldUseEIP1967StorageSlot(createProxy, accounts, labels, fnName) {
```
spalladino: I'd say that the tests in this file are already covered by the `shouldBehaveLikeProxy` tests run on both the new and the legacy implementations, so they should not be needed - but they definitely not hurt, so let's keep them.

PR #1101, opened 2019-07-16
Add support for deprecated manifest version (zosversion)

### LINKED ISSUE #1099: Lack of manifest tricks 2.5 release candidate in thinking we are using zos 1.x
I have `zos 2.4` installed globally and installed the 2.5 release candidate locally. The project was built using `zos 2.4`.

Screen Shot 2019-07-16 at 8 07 42 AM

When trying to compile with the 2.5 release candidate I get this error: 

```Manifest version identifier not found in /Users/dennison/Documents/Identity/zos.json. This means the project was built with an older version of openzeppelin (1.x), and needs to be upgraded. Please refer to the documentation at https://docs.zeppelinos.org for more info.``` 

It seems that the lack of manifest version is interpreted as being a `zos 1.x` project, which it is not. There also does not seem to be any information how to correct this.  The version jump from 1.x to 2.x was accompanied by a changelog that promised a migration guide would be created to help users move from 1.x to 2.x. Was this ever created? 
https://github.com/zeppelinos/zos/releases/tag/v2.0.0

### PR DESCRIPTION
**Pull request by jcarpanelli**
Fixes #1099.

### DIFF
--- a/packages/cli/test/models/ProjectFile.test.js
+++ b/packages/cli/test/models/ProjectFile.test.js
@@ -8,9 +8,7 @@ contract('ProjectFile', function() {
   describe('class methods', function() {
     describe('#getLinkedDependencies', function() {
       it('gets an array of dependencies', function() {
-        const dependencies = ProjectFile.getLinkedDependencies(
-          'test/mocks/packages/package-with-stdlib.zos.json',
-        );
+        const dependencies = ProjectFile.getLinkedDependencies('test/mocks/packages/package-with-stdlib.zos.json');
         dependencies.should.not.be.empty;
         dependencies.should.have.lengthOf(1);
       });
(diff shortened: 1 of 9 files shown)

### CODE REVIEW

File: packages/cli/test/models/ProjectFile.test.js [hunk 1]
```diff
@@ -24,22 +22,22 @@ contract('ProjectFile', function() {
...
-      ).to.throw(/Unrecognized manifest version identifier 3/);
+      expect(() => new ProjectFile('test/mocks/packages/package-unsupported-manifest-version.zos.json')).to.throw(
+        /Unrecognized manifest version identifier 3/,
+      );
     });
   });
+
+  it('supports deprecated manifest version', function() {
+    const file = new ProjectFile('test/mocks/packages/package-deprecated-manifest-version.zos.json');
+    file.data.zosversion.should.eq('2.2');
```
spalladino:
```suggestion
    file.manifestVersion.should.eq('2.2');
```

Shouldn't we be checking that the getter handles both formats?

spalladino: Actually, we'd need to add a getter to be able to do this (the network file has one, not the project file), so feel free to disregard this suggestion

spalladino:
Actually, I'm merging without the suggestion, I think it wasn't worth it. Again, great work!!

PR #1110, opened 2019-07-16
Use require.resolve to lookup contracts in deps

### LINKED ISSUE #1076: Doesn't work with yarn workspaces
I cannot use zos within a monorepo maintained with [yarn workspaces](https://yarnpkg.com/lang/en/docs/workspaces/). I'm getting the following error:

> Could not find a zos.json file for 'openzeppelin-eth'. Make sure it is provided by the npm package.

Workspaces is a feature from yarn that aggregates all common node modules at the root of your repo, but I presume zos is only looking for the relative folder, never above.

I also asked about an alternative solution on StackOverflow, but no one answered: https://stackoverflow.com/questions/56910467/when-using-yarn-workspaces-how-to-force-a-package-to-be-installed-in-the-relati

spalladino:
Change lookup in both Dependency and Contracts classes. Instead of looking for just node_modules in the current folder, it uses require.resolve (relative to the current workdir) to look for the dependency.
    
Fixes #1076

### DIFF
--- a/packages/lib/src/artifacts/Contracts.ts
+++ b/packages/lib/src/artifacts/Contracts.ts
@@ -12,6 +12,7 @@ export default class Contracts {
   private static timeout: number = Contracts.DEFAULT_SYNC_TIMEOUT;
   private static buildDir: string = Contracts.DEFAULT_BUILD_DIR;
   private static contractsDir: string = Contracts.DEFAULT_CONTRACTS_DIR;
+  private static projectRoot: string = null;
   private static artifactDefaults: any = {};
   private static defaultFromAddress: string;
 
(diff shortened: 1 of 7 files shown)


spalladino:
And here as well! :-)

packages/lib/src/artifacts/Contracts.ts
```diff
@@ -27,6 +28,10 @@ export default class Contracts {
     return path.resolve(Contracts.contractsDir || Contracts.DEFAULT_CONTRACTS_DIR);
   }
 
+  public static getProjectRoot(): string {
+    return path.resolve(this.projectRoot || process.cwd());
```
jcarpanelli: :+1: Nice one!

PR #1112, opened 2019-07-17
Handle incorrect keys to loggy

spalladino:
When calling Loggy.success with a non-existing key, the reference would not be found, and the call to path.basename in Loggy._log would fail (since its parameter was null). This commit:

- Catches any exceptions in _log, and displays an error (allowing to continue)
- Adds a testing mode to the logger, which is silent but executes logic, and throws an exception upon issues (only to be used in the CLI and lib tests)
- Fixes incorrect or missing keys in calls to the logger

_Note: we should cherry pick this one onto 2.4_

### DIFF
--- a/packages/cli/src/models/network/NetworkController.ts
+++ b/packages/cli/src/models/network/NetworkController.ts
@@ -233,7 +233,7 @@ export default class NetworkController {
     Loggy.spin(__filename, '_uploadSolidityLib', `upload-solidity-lib${libName}`, `Uploading ${libName} library`);
     const libInstance = await this.project.setImplementation(libClass, libName);
     this.networkFile.addSolidityLib(libName, libInstance);
-    Loggy.succeed(`upload-solidity-lob${libName}`, `${libName} library uploaded`);
+    Loggy.succeed(`upload-solidity-lib${libName}`, `${libName} library uploaded`);
   }
 
   // Contract model
--- a/packages/lib/test/src/utils/Logger.test.js
+++ b/packages/lib/test/src/utils/Logger.test.js
@@ -11,20 +11,20 @@ import {
   LogType,
 } from '../../../src/utils/Logger';
 
-describe('Loggger', function() {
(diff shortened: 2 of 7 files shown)


packages/cli/src/models/network/NetworkController.ts
```diff
@@ -233,7 +233,7 @@ export default class NetworkController {
     Loggy.spin(__filename, '_uploadSolidityLib', `upload-solidity-lib${libName}`, `Uploading ${libName} library`);
     const libInstance = await this.project.setImplementation(libClass, libName);
     this.networkFile.addSolidityLib(libName, libInstance);
-    Loggy.succeed(`upload-solidity-lob${libName}`, `${libName} library uploaded`);
+    Loggy.succeed(`upload-solidity-lib${libName}`, `${libName} library uploaded`);
```
jcarpanelli: Good catch!

packages/lib/test/src/utils/Logger.test.js
```diff
@@ -11,20 +11,20 @@ import {
   LogType,
 } from '../../../src/utils/Logger';
 
-describe('Loggger', function() {
-  afterEach('restore logger', function() {
-    Loggy.silent(true);
+describe('Logger', function() {
```
jcarpanelli: :+1:

packages/lib/test/src/utils/Logger.test.js
```diff
@@ -11,20 +11,20 @@ import {
...
     sinon.restore();
   });
 
   describe('default attributes', function() {
-    it('is silent and non-verbose by default', function() {
-      Loggy.isSilent.should.be.true;
-      Loggy.isVerbose.should.be.false;
-    });
-
     it(`has a 'logs' object initialized`, function() {
```
jcarpanelli: This test seems to be failing because `Loggy.logs` is being filled by some other test. I believe it's not even necessary to have it, as it's only testing a variable assignment rather than a behavior.

PR #1116, opened 2019-07-18
Fix verbose logging in base package project

spalladino:
Add missing parameters to the `onVerbose` call. We definitely need to add types to Logger.ts!

### DIFF
--- a/packages/lib/src/project/BasePackageProject.ts
+++ b/packages/lib/src/project/BasePackageProject.ts
@@ -58,12 +58,7 @@ export default abstract class BasePackageProject {
     );
     const directory: ImplementationDirectory = await this.getCurrentDirectory();
     await directory.setImplementation(contractName, implementation.address);
-    Loggy.onVerbose(
-      __filename,
-      'setImplementation',
-      `set-implementation-${contractName}`,
-      `Implementation set: ${implementation.address}`,
-    );
+    Loggy.succeed(`set-implementation-${contractName}`);
     return implementation;
   }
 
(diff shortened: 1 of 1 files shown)


packages/lib/src/project/BasePackageProject.ts
```diff
@@ -74,7 +76,12 @@ export default abstract class BasePackageProject {
     );
     const directory: ImplementationDirectory = await this.getCurrentDirectory();
     await directory.unsetImplementation(contractName);
-    Loggy.onVerbose(`unset-implementation-${contractName}`);
+    Loggy.onVerbose(
```
jcarpanelli: Shouldn't we keep this one as it was, but change the `onVerbose` to `succeed`?
It's not a big deal though, as this will be logged only on verbose mode, but will be consistent with other logs throughout the project.

packages/lib/src/project/BasePackageProject.ts
```diff
@@ -59,6 +59,8 @@ export default abstract class BasePackageProject {
     const directory: ImplementationDirectory = await this.getCurrentDirectory();
     await directory.setImplementation(contractName, implementation.address);
     Loggy.onVerbose(
+      __filename,
```
jcarpanelli: Good catch! :muscle:

packages/lib/src/project/BasePackageProject.ts
spalladino: Your call! Does Logger support calling a `success` on a `onVerbose`? If so, I'll adjust this.

PR #1119, opened 2019-07-18
Check outdated dependencies to push

spalladino:


packages/cli/test/models/Dependency.test.js
```diff
@@ -99,6 +99,16 @@ contract('Dependency', function([_, from]) {
           Dependency.hasDependenciesForDeploy('test').should.be.false;
         });
       });
+
+      context('when there are dependencies to update', function() {
+        it('returns true', function() {
+          Dependency.hasDependenciesForDeploy(
```
jcarpanelli: This is always returning false, as  the `dependency` variable in `Dependency.ts#L61` always equals `[]`.

Lemme fix that one for you :slightly_smiling_face:

PR #1123, opened 2019-07-18
Remove repeated flag -f from push and set-admin

### LINKED ISSUE #1052: On the CLI push docs, -f can't be both --force and --from
In the docs for the CLI `push` command, `-f` can't be shorthand for both `--force` and `--from`

https://docs.zeppelinos.org/docs/cli_push.html

spalladino:
Flag -f was already being used for --from, but was overwritten by both push and set-admin as force. Now --from is the only accepted meaning.

Fixes #1052


spalladino:
> push and set-admin were the only two commands where the --force flag was also defined as -f, right?

Yep!

PR #1129, opened 2019-07-19
Fix readme in docs and set correct readme in `packages` directory

### LINKED ISSUE #1126: Github shows the `packages/docs/README.md` as the `packages` directory readme
We should fix it so it does not show any readme (or show the readme in the root directory).
Also, the `docs` package still has references to `ZeppelinOS`.

jcarpanelli:
Fixes #1126.

I also renamed all the zos/ZeppelinOS references in the `docs/README.md` and added @frangio as a maintainer. We'll have to change the build and deploy instructions though.


packages/docs/README.md
```diff
@@ -39,9 +39,9 @@ This will serve the wesite locally in your computer at `http://localhost:3000/`.
 
 ## API Reference generation
```
frangio: This description is not entirely accurate, but if it's okay I can update it next week.

PR #1130, opened 2019-07-19
Implement `accounts` command

### LINKED ISSUE #967: Add a command to query current accounts
Add a `zos accounts` commands that simply outputs the result of `web3.eth.getAccounts()` for a given network.

jcarpanelli:
I had this in my `git stash` for a while... Just committed it so you can take care of it while I'm on vacations :slightly_smiling_face:  
Fixes #967?

*Note:* I wouldn't merge this feature for `v2.5.0`.

### DIFF
--- a/packages/cli/src/commands/accounts.ts
+++ b/packages/cli/src/commands/accounts.ts
@@ -1,0 +1,31 @@
+import accounts from '../scripts/accounts';
+import { networksList, promptForNetwork, InquirerQuestions } from '../prompts/prompt';
+import ConfigManager from '../models/config/ConfigManager';
+
+const name = 'accounts';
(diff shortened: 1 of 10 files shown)


spalladino:
> Pretty will break on Infura and other account-less providers.

Are you sure about this @ylv-io? I understand that offline-signing providers (such as the HDWallet one) capture the `eth_getAccounts` call and return the ones for which they have the keys. I'd expect this to work also on those cases.

ylv-io:
Nice comment @spalladino!
I just checked `blueprint.network.js`and it doesn't contain `HDWallet` or other providers. I guess that is a separate issue so feel free to merge the PR.

PR #1138, opened 2019-07-20
Handle packages from organizations on link

### LINKED ISSUE #1137: Error "Could not find dependency ." when installing @openzeppelin/contracts-ethereum-package
Using 2.5, on a fresh project:
```
$ oz link @openzeppelin/contracts-ethereum-package
✓ Dependency @openzeppelin/contracts-ethereum-package installed
Could not find dependency .
```

Note that installing `openzeppelin-eth` works fine.

spalladino:
Fixes #1137

### DIFF
--- a/packages/cli/src/models/dependency/Dependency.ts
+++ b/packages/cli/src/models/dependency/Dependency.ts
@@ -31,7 +31,7 @@ export default class Dependency {
   private _projectFile: ProjectFile;
 
   public static fromNameWithVersion(nameAndVersion: string): Dependency {
-    const [name, version] = nameAndVersion.split('@');
+    const [name, version] = this.splitNameAndVersion(nameAndVersion);
     return new this(name, version);
   }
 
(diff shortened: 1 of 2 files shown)


packages/cli/src/models/dependency/Dependency.ts
```diff
@@ -74,6 +74,15 @@ export default class Dependency {
     return this.fromNameWithVersion(nameAndVersion);
   }
 
+  public static splitNameAndVersion(nameAndVersion: string): [string, string] {
```
ylv-io:
```suggestion
    public static splitNameAndVersion(nameAndVersion: string): [string, string] {
    const isOrg = nameAndVersion.startsWith('@');
    const [name, version] = (isOrg ? nameAndVersion.slice(1) : nameAndVersion).split('@');
    return [isOrg ? `@${name}` : name, version];
  }
```

spalladino: Hmm why the extra indentation? Looks odd.

ylv-io: A glitch during copy-paste. Github actually doesn't allow to make multiple line suggestions yet.

PR #1146, opened 2019-07-22
Get dependency version with caret

spalladino:
Link a dependency version using the caret instead of exact version, and avoid relying on unix specific bash commands.

This allows linking a dependency like `@openzeppelin/contracts-ethereum-package` on version `^2.2.0`. This installs version 2.2.1, but allows using deployments made under version 2.2.0.


spalladino:
> My only question why we ignore patches?

Usually a bump in the patch does **not** imply a change in the deployed logic contracts. Using the caret allows us to use an older logic contract version with the new package.

PR #1148, opened 2019-07-22
Validate init params

### LINKED ISSUE #1142: Enhancement: Enforce Semantic Versioning when running oz init.
This is a minor enhancement, but when running `oz init` one of the prompts is for semantic version number. It might be nice to be somehow forced to use semantic versioning. Currently nothing stops us from using a string or any character combination as our version.

spalladino:
Validates that name is not empty, and that version is semver compatible. Only on interactive command.

Fixes #1142

### DIFF
--- a/packages/cli/src/commands/init.ts
+++ b/packages/cli/src/commands/init.ts
@@ -1,8 +1,10 @@
 import push from './push';
 import init from '../scripts/init';
+import semver from 'semver';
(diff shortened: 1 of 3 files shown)


spalladino:
> Side question – should we prefer 0.1.0 over 1.0.0 as default version?

I used npm's default to be honest

PR #1150, opened 2019-07-23
Initializable: moving initialized and initializing flags to memory slots

Amxx:
Similarly to the implementation, which is stored at position 0x360894a13ba1a3210667c828492db98dca3e2076cc3735a920a3ca505d382bbc to avoid conflict with the memory space of the master contract, it would be interesting to store the initialization flags away fro the straight forward memory space.

This would allow proxies and masters to be "initialization aware" without having to share a common header and risking conflicts.

An example of usage would be in [Kitsune Wallet's proxy base](https://github.com/Amxx/KitsuneWallet-ERC1836/blob/master/kitsune-contracts/contracts/proxy/BaseKitsuneProxy.sol) which has to `unlock()` (line 28) before re-initializating (during a migration). This unlock would be as simple as `_setInitialized(false)` with the proposed implementation.

### DIFF
--- a/packages/lib/contracts/Initializable.sol
+++ b/packages/lib/contracts/Initializable.sol
@@ -16,31 +16,77 @@ pragma solidity >=0.4.24 <0.6.0;
 contract Initializable {
 
   /**
-   * @dev Indicates that the contract has been initialized.
+   * @dev Storage slot with the address of the current 'initialized' flag.
+   * This is the keccak-256 hash of "zos.initializable.initialized" subtracted by 1
    */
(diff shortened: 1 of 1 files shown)


spalladino:
Hey @Amxx, thanks for the contribution! A change like this has been on the back of my mind for quite some time already, and I think it would certainly be interesting. 

However, since it is a breaking change, I want to tackle this along with other features related to initialization (such as automating the generation of initializers from constructors). We will probably be focusing on this on September/October. Stay tuned!

PR #1152, opened 2019-07-23
Migrate to web3 1.2.0

spalladino:
Let's give the new stable web3 a shot!


spalladino:
@ylv-io what do you think? Should we merge?

ylv-io:
Should be fine I guess given the [discussion](https://github.com/ethereum/web3.js/pull/2905) 👍

PR #1154, opened 2019-07-24
Merge docs setup updates

frangio:
While preparing the documentation for the new site, I did some configuration and content changes in the `docs/2.5` branch, which we should port to `master`. This is a different branch that points at the same commit because I don't want `docs/2.5` to show up as "merged". It's a long-lived branch that should be kept in the repository, since it's [used by the docs site generator](https://github.com/OpenZeppelin/docs.openzeppelin.com/blob/26ef2da3e9a88e0e026db9f57ba1b4026018c2b5/playbook.yml#L14-L16) to fetch the content.

From now own, `docs/2.5` should be used only to cherry-pick changes from `master` that should be reflected immediately on the site. This cherry-picking will be automated by a bot at some point.

(Please use a squash merge, there a lot of tiny commits here.)


spalladino:
> This is a different branch that points at the same commit because I don't want docs/2.5 to show up as "merged".

FWIW, we often merge release branches into master, even if they are to be long-lived, so I wouldn't worry too much about it.

PR #1155, opened 2019-07-24
Add public-deploy doc to index

spalladino:


packages/docs/modules/ROOT/pages/public-deploy.adoc
```diff
@@ -1,7 +1,7 @@
 [[deploy-your-contracts-to-a-public-network]]
 = Deploy your contracts to a public network
 
-This tutorial link:first[builds on the first tutorial] where we created a new OpenZeppelin SDK project, and created a simple `Counter` contract in a local development network. We will now see how to deploy this contract to a public network.  
+This guide link:first[builds on the first tutorial] where we created a new OpenZeppelin SDK project, and created a simple `Counter` contract in a local development network. We will now see how to deploy this contract to a public network.
```
frangio: For internal links we should use Antora crossreferences, as they are more reliable than relative links.

```suggestion
This guide xref:first.adoc[builds on the first tutorial] where we created a new OpenZeppelin SDK project, and created a simple `Counter` contract in a local development network. We will now see how to deploy this contract to a public network.  
```

frangio:
Somewhat unrelated, I noticed that you named this branch `docs/add-public-deploy`. We haven't set this up yet, but in the future the docs site would take all branches that match the pattern `docs/*` and put them in the site. So this branch would've caused a problem for that.

Perhaps we should use a different pattern for the docs site, that is less likely to be used like this? Somethinhg like `release-docs/*`.

spalladino:
> Perhaps we should use a different pattern for the docs site, that is less likely to be accidentally used for other things? Something like `release-docs/*`.

Agree!

PR #1156, opened 2019-07-24
Add revert message to reverts in ProxyFactory

0age:
Revert messages can be retrieved from the return buffer and passed along in a few locations.


spalladino:
Interesting, thanks for the contribution @0age! We'll add a few tests to try this out and merge.

PR #1166, opened 2019-07-28
Migrate encodeCall.ts to web3 per comment

CallMeGwei:
There was a comment at the top of helper encodeCall.ts about migrating to web3. I went ahead and migrated it for my own project. Thought I would submit it here as well.

Web3 handles encoding parameters just a little differently than what was being used before. Web3 wraps around if given a negative for type uint rather than throwing. It also appears to convert empty buffers to nulls rather than keeping them as empty buffers. I had to tweak the test file accordingly.

I also added a helper function in the test to convert from the web3 return value (which is an object) to what was expected in the asserts (which was an array).


spalladino:
Thanks for the contribution, @CallMeGwei! We'll review this in a few days :-)

PR #1192, opened 2019-08-15
Bump web3 version

spalladino:
Bumps web3 version to 1.2.1. This version has better node 12 support, and removes the problems with the .git folder from frozeman's websocket package. However, it introduces another issue: gulp being installed as a dependency. This has already been reported and may be fixed on a subsequent release.


ylv-io:
What `first-project`? Should we bump its web3 dep as well?

spalladino:
Good catch! I had missed it as it was tagged with a caret

PR #1198, opened 2019-08-19
Update public-deploy.adoc to use `openzeppelin accounts`

abcoathup:
Update guide to use `openzeppelin accounts` to get the configured account for the network.

Previous version of guide required importing mnemonic into MetaMask to get account to fund from a faucet.


abcoathup:
Hi @jcarpanelli 

`public-deploy.adoc` isn't in the production documentation yet.  It still needs to be merged (assume by @frangio) 

Also `openzeppelin accounts` is only in 2.5.3 which is currently `@openzeppelin/cli@next` rather than `@latest`.  So may need to wait until 2.5.3 is @latest.

PR #1205, opened 2019-08-24
Fixed method name in react code

Flash-Git:


Flash-Git:
> Hey @Flash-Git! I believe this change is wrong. The methods variable stores the Counter contract instance functions (initialize, increaseCounter, decreaseCounter) and there's no decrease function. Could you please re-check it?

I agree but I didn't touch the increase/decrease functions or their calls, I believe that this only adds brackets.

I'm happy to make another pull to handle that as well if you like.

Flash-Git:
I'll have a look and report back, thanks.

Flash-Git:
Although, I believe it would be better to refactor the code further and simply call `decrease()` the same way that `increase()` is called on the above button.

jcarpanelli:
You are right @Flash-Git! I was checking the code in [the gsn kit](https://github.com/OpenZeppelin/starter-kit-gsn) which is very similar to this one :sweat_smile:. It seems that this particular section on the documentation has some copy-pasted code directly from that kit, but in this particular guide that the `decrease` function is not necessary at all as we are not explaining how to upgrade a contract, and that topic is covered in other sections of the docsite. I'd then suggest to remove the `decrease` references from both the javascript and JSX code. WDYT? Let me know if you want me to tackle that, otherwise, feel free to change it so we can merge this PR :muscle: 

Good catch!

Flash-Git:
Thanks @jcarpanelli, I'm happy to make those changes in a minute!

packages/docs/modules/ROOT/pages/gsn-dapp.adoc
```diff
@@ -235,18 +228,9 @@ function App() {
             <div>{count}</div>
           </div>
           <div>Counter Actions</div>
-          <div>
-            <button onClick={() => increase(1)} size="small">
+            <button onClick={() => increase()} size="small">
```
jcarpanelli: Nice catch :muscle:!

Issue #1203, opened 2019-08-23
Invalid JSON output in compiler artifacts with multiple contracts with the same name


spalladino:
Given a project with more than one contract with the same name (see [here](https://github.com/crazyrabbitLTC/ETHBERLIN/tree/93592ea49f4f908ac64eda9a49e6095a3b31d064) for an example), the compiler generates invalid JSON files for them.

```
  "compiler": {
    "name": "solc",
    "version": "0.5.3+commit.10d17f24.Emscripten.clang",
    "optimizer": {},
    "evmVersion": "constantinople"
  }
}
nople"
  }
}
```

This is probably because we have more than one stream writing to the same file. To fix this, we should check the list of compilerOutput and remove any duplicate contract names before writing.

https://github.com/OpenZeppelin/openzeppelin-sdk/blob/45778202249e725a368fdb6a6600ccd49e1c66bb/packages/cli/src/models/compiler/solidity/SolidityProjectCompiler.ts#L189-L194

However, having duplicated contract names can yield to other errors, such as invalid storage layout calculations (since we rely on the AST of the build artifacts for that) or even unexpected contracts getting created (since the user may expect one of the bytecodes to be deployed, but we actually pick the other). We should also rise a big warning if there are indeed duplicated contracts.

melnikaite:
Could you please update link with example?

melnikaite:
I found 2 serious issues

First of all flat list of artefacts, in case we have Counter.sol and subdir/Counter.sol in build/contracts we will have artefact only for subdir/Counter.sol, so we are missing artefact for Counter.sol

And second issue you actually faced with is having multiple contracts in the same .sol file, e.g.
```
pragma solidity 0.5.4;

contract Counter {
    uint256 public value;

    function increase() public {
        value++;
    }

    function decrease() public {
        value--;
    }
}

contract X {}
```

For now the only workaround I see is keeping each contract in separate file and use unique contract names. However you still can use subfolders for imported contracts.

nventuro:
After encountering this myself, I'd say that erroring out early is the way to go: many parts of the SDK assume contract names are unique (e.g. when deploying via `create`), so we should notify the user as soon as we detect this and abort.

abcoathup:
Also reported in the community forum:
https://forum.openzeppelin.com/t/error-unexpected-string-in-json-at-position-23357/1614

I reproduced (and went to create an issue, and then recalled having seen this issue, oops):

[...]

PR #1213, opened 2019-08-30
Support contracts with the same name #1203

melnikaite:


jcarpanelli:
Hey @melnikaite, thanks for tackling this one! Let me know when it's ready for review :+1:.

melnikaite:
Hi @jcarpanelli 

I manage to successfully run `oz push`, but stuck with `oz create`, because it try to install `contract/name` as npm dependency, but obviously there is no such npm package and I didn't find the place where sdk check if it should run npm install

spalladino:
Hey @melnikaite! Thanks for the contribution :-)
Is this the function you are looking for regarding installing an npm dependency?

https://github.com/OpenZeppelin/openzeppelin-sdk/blob/5140c692f62a9858ee6580f2521197a938819a4b/packages/cli/src/models/dependency/Dependency.ts#L69-L74

melnikaite:
Yes, but in fact my issue was in `fromContractFullName`

It seems functionality works fine for all commands, but currently I don't have time to fix all the tests

PR #1220, opened 2019-09-03
Update interacting.adoc to use `oz accounts`

abcoathup:
Suggested by @jcarpanelli: https://github.com/OpenZeppelin/openzeppelin-sdk/pull/1198#pullrequestreview-282705864

Note: `openzeppelin accounts` is only in 2.5.3 which is currently `@openzeppelin/cli@next` rather than `@latest`. So may need to wait until 2.5.3 is `@latest`


abcoathup:
I looked at removing the `getAccounts` Snippet but the purpose is to get users to test a simple Web3 connection so it felt important to keep it. 

Yay for 2.5.3 being latest!

PR #1234, opened 2019-09-16
Update public-deploy to include Rinkeby and Kovan

abcoathup:
gsn-dapp documentation: https://docs.openzeppelin.com/sdk/2.5/gsn-dapp#moving-to-testnet uses Rinkeby network, so added Rinkeby and Kovan (had issue with default gas limit for Goerli otherwise would have added as well)

### DIFF
--- a/packages/docs/modules/ROOT/pages/public-deploy.adoc
+++ b/packages/docs/modules/ROOT/pages/public-deploy.adoc
@@ -100,6 +100,14 @@ module.exports = {
       provider: () => new HDWalletProvider(process.env.DEV_MNEMONIC, "https://ropsten.infura.io/v3/" + infuraProjectId),
       networkId: 3,       // Ropsten's id
     },
+    rinkeby: {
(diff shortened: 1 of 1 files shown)


packages/docs/modules/ROOT/pages/public-deploy.adoc
```diff
@@ -100,6 +100,14 @@ module.exports = {
       provider: () => new HDWalletProvider(process.env.DEV_MNEMONIC, "https://ropsten.infura.io/v3/" + infuraProjectId),
       networkId: 3,       // Ropsten's id
     },
+    rinkeby: {
+      provider: () => new HDWalletProvider(process.env.DEV_MNEMONIC, "https://rinkeby.infura.io/v3/" + infuraProjectId),
```
jcarpanelli: we could use template literals here :nerd_face:, they are easier to read!

abcoathup: I’ll have to do another PR

PR #1239, opened 2019-09-20
abcoathup add network js doc

abcoathup:
First draft of Network.js documentation.  

Similar to README currently, so may need to be structured more like a guide with example code?

### DIFF
--- a/packages/docs/modules/ROOT/pages/network-js.adoc
+++ b/packages/docs/modules/ROOT/pages/network-js.adoc
@@ -31,7 +31,7 @@ npx create-react-app client
 Our dapp will display the current Ethereum network that we are connected to and which web3 provider is being used.
 
 In the code below, first we import `useWeb3` from the React implementation of *OpenZeppelin Network.js* (`@openzeppelin/network/react`).
-We then get a web3Context using `useWeb3` which will attempt to retrieve an injected web3 provider (e.g. MetaMask), otherwise it will fall back to a network connection, which in this case is a connection to mainnet via Infura.
+We then get a web3Context using `useWeb3`. This hook will attempt to retrieve an injected web3 provider (e.g. MetaMask), otherwise it will fall back to a network connection, which in this case is a connection to mainnet via Infura.
 The `networkId`, `networkName` and `providerName` are obtained from the web3Context and displayed in the dapp.
 
 On the `client/src/App.js` file, replace the placeholder code in `App.js` in our react project with the following code:
--- a/packages/docs/modules/ROOT/pages/network-js.adoc
+++ b/packages/docs/modules/ROOT/pages/network-js.adoc
@@ -106,12 +106,10 @@ In the `client/src/App.js` file, add an import for our `Web3Info` component:
 import Web3Info from './components/Web3Info.js';
 ----
 
-Then replace the `App` function in `App.js` with the following code:
+Then replace the contents of the `App` function in `App.js` with the following code:
 
 [source,js]
 ----
-function App() {
-
   const web3Context = useWeb3(`wss://mainnet.infura.io/ws/v3/${infuraProjectId}`);
 
   return (
@@ -122,7 +120,6 @@ function App() {
       </div>
     </div>
   );
-}
 ----
 
 Start our dapp again by running `npm start` from within the `client` directory. 
--- a/packages/docs/modules/ROOT/pages/network-js.adoc
+++ b/packages/docs/modules/ROOT/pages/network-js.adoc
@@ -128,14 +128,16 @@ Start our dapp again by running `npm start` from within the `client` directory.
 
 Try changing networks in MetaMask and *OpenZeppelin Network.js* will cause the component to update with the network ID and network name.  
 
-== Add request access to injected web3 to the dapp
-For the third iteration of our dapp we will add a mechanism request to access injected web3 and once the user has connected will display the address of the users account.
+== Add request access to account address to the dapp
+For the third iteration of our dapp we will add a mechanism to request access to the users address and once the user has connected will display the account address.  
 
-NOTE: Best practice is to wait to request injected web3 access until a user wants to perform an action that they can only do with injected web3.  See https://github.com/ethereum/EIPs/blob/master/EIPS/eip-1102.md[EIP-1102] for details.
+Injected web3 providers don't give the dapp access to a users address until the user has given permission.  This is to protect the privacy of users, otherwise websites could track users based on their account address.  See https://github.com/ethereum/EIPs/blob/master/EIPS/eip-1102.md[EIP-1102] for details.
+
+NOTE: Best practice is for a dapp to wait to request access to the users address until a user wants to perform an action that they can only do which requires this access.  
 Remember dapps can be configured to use multiple web3 providers such as a network provider if necessary.
 
-In the code below, we get `accounts` from the `web3Context` and if available the users account is displayed in the dapp.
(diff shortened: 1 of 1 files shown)


abcoathup:
I haven't got a pattern to handle when there is no Injected Web3 https://github.com/OpenZeppelin/openzeppelin-network.js/issues/16

Also I don't have a working non-react sample

abcoathup:
I split the dapp code into four (rather than three), so that I didn't have to start with a component and the code is fairly small, though not sure if it is worth the effort, so could combine the first and second iterations.

I changed the order slightly as had request access before account balance.

packages/docs/modules/ROOT/pages/network-js.adoc
```diff
@@ -0,0 +1,302 @@
...
+
+[source,console]
+----
+npx create-react-app client
+----
+
+Our dapp will display the current Ethereum network that we are connected to and which web3 provider is being used.
+
+In the code below, first we import `useWeb3` from the React implementation of *OpenZeppelin Network.js* (`@openzeppelin/network/react`).
+We then get a web3Context using `useWeb3` which will attempt to retrieve an injected web3 provider (e.g. MetaMask), otherwise it will fall back to a network connection, which in this case is a connection to mainnet via Infura.
```
frangio:
```suggestion
We then get a web3Context using `useWeb3`. This hook will attempt to retrieve an injected web3 provider (e.g. MetaMask), otherwise it will fall back to a network connection, which in this case is a connection to mainnet via Infura.
```

frangio: Is there a particular reason why you're not using code formatting for "web3Context"?

packages/docs/modules/ROOT/pages/network-js.adoc
```diff
@@ -0,0 +1,302 @@
...
+We can start our dapp by running `npm start` from within the `client` directory. 
+
+The dapp will display the Ethereum network ID and network name, either of mainnet from our network connection if no injected web3 provider such as MetaMask is available, otherwise it will display the current network as selected by the injected web3 provider, along with the web3 provider name.
+
+We use `useWeb3` which attempts to obtain an injected web3 provider first, before falling back to a network connection.  Alternatively we could use `useWeb3Injected` for an injected web3 provider or `useWeb3Network` for a network provider e.g. Infura or a private node.  Dapps can also be configured to use multiple providers if necessary.
+
+NOTE: If you use Infura in your own dapp, you will need to create an account with Infura and setup an Infura Project (with corresponding Infura Project ID).  See https://infura.io[Infura website] to create an account.
+
+== Add a component to the dapp
+For the second iteration of our dapp we will move the display of the current Ethereum network to a component and show how components are re-rendered when changes are made, such as the network.
```
frangio:
```suggestion
For the second iteration of our dapp we will move the display of the current Ethereum network to a component and see how components are re-rendered when changes are made, such as the network.
```

packages/docs/modules/ROOT/pages/network-js.adoc
```diff
@@ -0,0 +1,302 @@
...
+
+To install MetaMask as our injected web3 provider (if you don't have an injected web3 provider installed already), follow the instructions on the https://metamask.io[MetaMask website] to install the browser extension.
+
+In the code below, our `Web3Info` component expects a web3Context.  The `networkId`, `networkName` and `providerName` are obtained from the web3Context and displayed in the component.
+
+In the `client/src` directory create a `components` directory. In the `components` directory create a `Web3Info` directory for our `Web3Info` component.  Then create an `index.js` file in the `Web3Info` directory with the following code:
```
frangio: This is equivalent and much simpler.

```suggestion
In the `client/src` directory create a `components` directory. In the `components` directory create a `Web3Info.js` file  with the following code:
```

packages/docs/modules/ROOT/pages/network-js.adoc
```diff
@@ -0,0 +1,302 @@
...
+      <div>Network: {networkId ? `${networkId} – ${networkName}` : 'No connection'}</div>
+      <div>Provider: {providerName}</div>
+    </div>
+  );
+}
+----
+
+In our App we use our `Web3Info` component and provide a web3Context.
+
+On the `client/src/App.js` file, replace `App.js` in our react project with the following code:
```
frangio: Instead of replacing the entire file contents, I think it's possible to tell the user to "replace the contents of the `App` function" and have a shorter and more relevant snippet. What do you think?

packages/docs/modules/ROOT/pages/network-js.adoc
```diff
@@ -0,0 +1,302 @@
...
+----
+
+Start our dapp again by running `npm start` from within the `client` directory. 
+
+*OpenZeppelin Network.js* re-renders React components on changes such as account, network, and connection.
+
+Try changing networks in MetaMask and *OpenZeppelin Network.js* will cause the component to update with the network ID and network name.  
+
+== Add request access to injected web3 to the dapp
+For the third iteration of our dapp we will add a mechanism request to access injected web3 and once the user has connected will display the address of the users account.
```
frangio: I think it's worth explaining to people that wallets have this measure in place to prevent the leaking of account information without user consent. It'll be very helpful to give the motivation behind it.

packages/docs/modules/ROOT/pages/network-js.adoc
```diff
@@ -0,0 +1,302 @@
...
+  return (
+    <div>
+      <h3> {props.title} </h3>
+      <div>Network: {networkId ? `${networkId} – ${networkName}` : 'No connection'}</div>
+      <div>Your address: {accounts && accounts.length ? accounts[0] : 'Unknown'}</div>
+      <div>Provider: {providerName}</div>
+      {accounts && accounts.length ? (
+        <div>Accounts & Signing Status: Access Granted</div>
+      ) : !!networkId && providerName !== 'infura' ? (
+        <div>
```
frangio: I change this to a `<button>` element and remove the `buttonStyle` variable. I feel that it distracts from the main point.

packages/docs/modules/ROOT/pages/network-js.adoc
abcoathup: Shows how little I currently know about React components :smile:

packages/docs/modules/ROOT/pages/network-js.adoc
abcoathup: As web3Context was a variable name that I chose I didn't use code formatting (compared with names from the library), but will change.

packages/docs/modules/ROOT/pages/network-js.adoc
abcoathup: Agree.  Have updated

packages/docs/modules/ROOT/pages/network-js.adoc
abcoathup: I think the starter kit uses a Button component and I couldn't work out how to create a button easily so did button style (rather than using a `button` element).  Oops.  I really need to learn React.

It looks better now!  Thank you.

packages/docs/modules/ROOT/pages/network-js.adoc
abcoathup: Added explanation and moved link to EIP-1102.

Would appreciate your review on the wording.

abcoathup:
Updated based on review.  Thank you :pray:

PR #1240, opened 2019-09-20
Implement regular deploys in the CLI

jcarpanelli:
This PR (which is still work in progress) adds the ability to do *regular deploys*, that is to say, to create non upgradeable instances that don't rely in proxies to work.

### TODO:
- ~Rename Initializable to Upgradeable (and keep alias?)~
- Fix and add tests
- Add Regular to "ProxyType" (and rename to InstanceType)
- ~Verify that Upgradeable is not in the inheritance chain, otherwise fail~
- Check that Upgradeable is from openzeppelin/upgrades
- ~Remove initializable function prompt for regular deploys~
- Hide regular instances from upgrade interactive prompt
- ~Rename create to deploy (and keep alias)~
- Review all commands one by one and think about how this change affects each one


frangio:
Superceded by #1376!

PR #1252, opened 2019-10-17
Libraries can link to other libraries with public functions

asselstine:
I needed to deploy a library that linked to another library that had public functions.  I made changes to allow the deployment of any dependent libraries.

- Altered NetworkController#_getAllSolidityLibNames to recursively retrieve libraries and detect cycles
- Altered NetworkController#uploadSolidityLibs to push libraries in order so that dependencies to ensure linking addresses exist
- Altered NetworkController#_uploadSolidityLib to link libraries


spalladino:
Hey @asselstine, thanks for the contribution! We'll check why the CI is complaining (there is some error with truffle apparently). In the meantime, could we ask you to add a test for this, in order to merge it? Thanks!

asselstine:
@spalladino I didn't see any tests for the NetworkController so I thought I might get away with no tests ;) I'll make sure to add one.

I did notice that I had to add the CI flag when running locally; i.e. `CI=true npm test`.  It couldn't seem to compile the artifacts correctly otherwise.

asselstine:
@spalladino tests added!

spalladino:
Awesome, thanks so much Brendan! Will review during the week :-)

packages/cli/src/models/network/NetworkController.ts
```diff
@@ -313,11 +318,29 @@ export default class NetworkController {
...
+    let libNames = newLibNames;
+    while (newLibNames.length > 0) {
+      // Get new library dependencies
+      newLibNames = this._getSolidityLibNames(newLibNames);
+
+      // Prepend new libs to ensure they are deployed first
+      libNames = newLibNames.concat(libNames);
+
+      // Detect loops
+      if (uniq(libNames).length !== libNames.length) {
```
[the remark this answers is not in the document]
asselstine: Yes, you're absolutely right.

I realized that cycles are fine: what we need is a terminating condition for the recursion.

The recursion should terminate when no *new* libs have been added.  That means that every lib has been searched.

I'll adjust the algo when I get a chance.

asselstine:
@spalladino I've improved the algorithm and added a test to ensure that it orders dependencies correctly.  I think this is the one!

asselstine:
@ylv-io Good idea; it would be a good use of recursion.  I've added you as a collaborator to the fork, so feel free to push to this branch.

packages/cli/src/models/network/NetworkController.ts
```diff
@@ -313,11 +318,42 @@ export default class NetworkController {
...
+        break;
+      }
+      // otherwise
+      // look into the new libs
+      newLibNames = uniqNewLibNames
+      // update all libs
+      libNames = nextLibNames
+    }
+
+    return libNames;
```
ylv-io: May I suggest we use a recursive version? I find it easy to read and understand.
```
    let newLibNames = this._getSolidityLibNames(contractNames);
    // deploy children before parents
    // keep contracts unique
    return uniq([...(newLibNames.length ? this._getAllSolidityLibNames(newLibNames) : []), ...newLibNames]);
```
It does a full walk over a forest though.

packages/cli/test/models/network/NetworkController.test.js
```diff
@@ -0,0 +1,101 @@
...
+      const contractNames = ['A', 'B', 'C']
+
+      controllerMock.expects('_getSolidityLibNames').withArgs(de(contractNames)).returns(['E', 'F', 'D'])
+      controllerMock.expects('_getSolidityLibNames').withArgs(de(['E', 'F', 'D'])).returns(['D', 'G', 'H'])
+      controllerMock.expects('_getSolidityLibNames').withArgs(de(['G', 'H'])).returns(['D', 'I'])
+      controllerMock.expects('_getSolidityLibNames').withArgs(de(['I'])).returns([])
+
+      const result = controller._getAllSolidityLibNames(['A', 'B', 'C'])
+
+      assert.deepEqual(result, ['I', 'D', 'G', 'H', 'E', 'F'])
```
ylv-io: And the test has to be a bit different because of it does a full walk.
```
controllerMock
        .expects('_getSolidityLibNames')
        .withArgs(de(contractNames))
        .returns(['E', 'F', 'D']);
      controllerMock
        .expects('_getSolidityLibNames')
        .withArgs(de(['E', 'F', 'D']))
        .returns(['D', 'G', 'H']);
      controllerMock
        .expects('_getSolidityLibNames')
        .withArgs(de(['D', 'G', 'H']))
        .returns(['D', 'I']);
      controllerMock
        .expects('_getSolidityLibNames')
        .withArgs(de(['D', 'I']))
        .returns([]);

      const result = controller._getAllSolidityLibNames(['A', 'B', 'C']);

      assert.deepEqual(result, ['D', 'I', 'G', 'H', 'E', 'F']);
```

ylv-io:
@asselstine for some reason it didn't work. I can't push and not on the list of collaborators.

asselstine:
@ylv-io Github's "invitations" are overly complicated, in my opinion.  I've had this problem countless times.

Your invite to the fork is still pending, so I'm guessing you need to check the email that is attached to your Github profile.  There will be an invite email from the Github repo.  Once you click that you'll be able to push to the branch.

ylv-io:
@asselstine That is exactly the case. Indeed, this is a confusing system.

packages/cli/src/models/network/NetworkController.ts
```diff
@@ -287,6 +292,7 @@ export default class NetworkController {
   private async _unsetSolidityLibs(): Promise<void> {
     const { contractNames } = this.projectFile;
     const libNames = this._getAllSolidityLibNames(contractNames);
+    console.log(libNames);
```
asselstine: ✂️

packages/cli/src/models/network/NetworkController.ts
```diff
@@ -313,11 +319,17 @@ export default class NetworkController {
 
   // Contract model || SolidityLib model
   private _getAllSolidityLibNames(contractNames: string[]): string[] {
+    let newLibNames = this._getSolidityLibNames(contractNames);
+    // deploy children before parents
+    // we want to deploy each contract only once
+    return uniq([...(newLibNames.length ? this._getAllSolidityLibNames(newLibNames) : []), ...newLibNames]);
+  }
```
asselstine: Very succinct!

One thing though: there may be a deadlock here.  It's possible to compile code like so:

```solidity
pragma solidity ^0.5.10;

import "./TestLib2.sol";

library TestLib1 {
  function test() public pure returns (string memory) {
    return TestLib2.test();
  }
}
```

```solidity
pragma solidity ^0.5.10;

import "./TestLib1.sol";

library TestLib2 {
  function test() public pure returns (string memory) {
    return TestLib1.test();
  }
}
```

Here TestLib1 depends on TestLib2 which depends on TestLib1 etc.  Compiling it is possible, even if deploying is not.  Further, there may be arbitrary depths of cycles: imagine if TestLib1 -> TestLib2 -> TestLib3 -> TestLib1 -> TestLib2 ... and so on.

One answer to this could be that the function _getAllSolidityLibNames has a second parameter which is all current parent dependencies.  If you could remove all the parent dependencies then you can determine what lib hasn't been searched already.

```typescript
private _getAllSolidityLibNames(contractNames: string[], allDependencies: string[] = []): string[] {
  let newLibNames = this._getSolidityLibNames(contractNames);

  let newLibs = difference(newLibNames, allDependencies)
  let repeatLibs = intersection(newLibNames, allDependencies)
  let oldLibs = difference(allDependencies, newLibNames)

  // reorder so that dependencies are pulled up
  let reorderedOldLibs = repeatLibs.concat(oldLibs)

  // create a new list of all libs
  let allLibs = newLibs.concat(reorderedOldLibs)

  return uniq([...(newLibs.length ? this._getAllSolidityLibNames(newLibs, allLibs) : []), ...reorderedOldLibs]);
}
```

I think that should do it.

asselstine:
Come to think of it, cycle detection *is* needed for this function.  It should throw an error if there is a library dependency ordering such as:

Lib_1 requires Lib_2
Lib_2 requires Lib_3
Lib_3 requires Lib_1 

While the above can be compiled, it can't be deployed all at once so OZ SDK should throw an error here.

packages/cli/src/models/network/NetworkController.ts
ylv-io: Damn. I completely missed the fact that you can compile code which wouldn't deploy. Thanks for pointing out that!

ylv-io:
Hey @asselstine. I've added suggested changes. Thanks again for your amazing contribution!

packages/cli/test/models/network/NetworkController.test.js
```diff
@@ -42,16 +42,16 @@ contract('NetworkController', function() {
         .returns(['D', 'G', 'H']);
       controllerMock
         .expects('_getSolidityLibNames')
-        .withArgs(de(['D', 'G', 'H']))
+        .withArgs(de(['G', 'H']))
         .returns(['D', 'I']);
```
asselstine: Excellent!  It even tests for it now

asselstine:
I realize that we are still treating loop as legit DAGs; as per Santiago's original comment.

What we need is a topological sort.  

I'll have a look at integrating one.

asselstine:
@ylv-io Ok.   This should be it!

It now recursively builds up a graph which is then topographically sorted.

[x] Repeated deps
[x] cycle detection.

I think we're done.

ylv-io:
@asselstine. It is a pleasure to collaborate with you!
Using `TopologicalSort` is a great idea. Just reminds me that we should always stand of the shoulders of giants!
I've polished code a bit and If there will be any CI tests errors let me handle that!

packages/cli/package.json
```diff
@@ -119,6 +119,7 @@
     "solc-wrapper": "^0.5.8",
     "solidity-parser-antlr": "^0.4.2",
     "spinnies": "^0.3.0",
+    "topological-sort": "^0.3.0",
```
nventuro: I'd rather we used [`toposort`](https://www.npmjs.com/package/toposort), with millions of weekly downloads, than this [relatively unused library](www.npmjs.com/package/topological-sort).

packages/cli/src/models/network/NetworkController.ts
```diff
@@ -226,12 +227,16 @@ export default class NetworkController {
 
   // Contract model || SolidityLib model
   public async uploadSolidityLibs(libs: Contract[]): Promise<void> {
-    await allPromisesOrError(libs.map(lib => this._uploadSolidityLib(lib)));
+    // Libs may have dependencies, so deploy them in order
+    for (let i = 0; i < libs.length; i++) {
+      await this._uploadSolidityLib(libs[i]);
```
nventuro: Is it necessary to wait for each deployment to be mined before the next one happens? Execution order should be guaranteed by the sender account's nonce field, and this approach will make deployment much slower.

That said, such a deployment is a rare scenario, and this _is_ safer in that a failing transaction with halt the whole deployment process, as opposed to potentially causing in-flight transactions to fail.

nventuro: Nevermind, we need for the current library deployment to actually go through because future libraries may depend on it, and need its address to link to it.

packages/cli/src/models/network/NetworkController.ts
```diff
@@ -226,12 +227,16 @@ export default class NetworkController {
...
+    }
   }
 
   // Contract model || SolidityLib model
   private async _uploadSolidityLib(libClass: Contract): Promise<void> {
     const libName = libClass.schema.contractName;
+    await this._setSolidityLibs(libClass);
```
nventuro:
```suggestion
    await this._setSolidityLibs(libClass); // Libraries may depend on other libraries themselves
```

packages/cli/test/models/network/NetworkController.test.js
```diff
@@ -0,0 +1,191 @@
...
+      /**
+       *
+       *  A => D, E, F
+       *  B => C, E, F, G
+       *  C => D, G
+       *  D => []
+       *  E => D, Y
+       *  F => G
+       *  G => E
+       *  Y => []
```
nventuro:
```suggestion
       *  Y => Z
       *  Z => []
```

The dependency list was outdated from the actual test code

packages/cli/test/models/network/NetworkController.test.js
```diff
@@ -0,0 +1,191 @@
...
+       *
+       *  Expected ordering:
+       *
+       *  [Y, D, E, G, F, A, C, B]
```
nventuro:
```suggestion
       *  [D, Z, Y, E, G, F, A, C, B]
```

Same (note that Y now depends on Z)

packages/cli/package.json
nventuro: I won't block on this though, and I'm not sure how strict you often are about this sort of thing. @ylv-io?

ylv-io: Let's use `topsort`. It is a right thing to do.

packages/cli/src/models/network/NetworkController.ts
ylv-io: Yes. It is a dependency tree.

ylv-io:
I've replaced the `topological-sort` library with `toposort`. I also had to adjust tests which is not ideal but required because `toposort` outputs libs in the different order which is still valid.

packages/cli/src/models/network/NetworkController.ts
```diff
@@ -318,27 +319,24 @@ export default class NetworkController {
 
   // Contract model || SolidityLib model
   private _getAllSolidityLibNames(contractNames: string[]): string[] {
-    const graph = new TopologicalSort<string, string>(new Map<string, string>());
+    const graph: string[][] = [];
+    const nodes: string[] = [];
 
     contractNames.forEach(contractName => {
-      const deps = this._getContractDependencies(contractName);
-      deps.forEach(dependencyContractName => {
```
nventuro: Nice, I hadn't realized this loop was duplicated :+1:

nventuro:
The CI test run is done on the branch, not on the result of the merge: it turns out this PR broke tests on the CLI due to #1288.

nventuro:
Fixed in 7f641a797834143aa98cf342c5dafac5de4d30c4

asselstine:
Great collab @ylv-io!  I'm pleased to help and I'm excited to have this in the build.  Now I won't have to worry about our library structure.

ylv-io:
@nventuro this is a great example of how CI tests can be successful on a branch but break a master.

spalladino:
Hey @asselstine, this PR is now part of the [2.7 RC](https://forum.openzeppelin.com/t/openzeppelin-cli-2-7-rc-0-available/2133) if you want to try it!

asselstine:
@spalladino Fantastic!

PR #1266, opened 2019-10-25
Add week number to telemetry report

### LINKED ISSUE #1260: Add dates to telemetry reports
Add dates (but no time) to telemetry reports. Ensure they can also be sorted, so we can understand the journey for a specific user.

frangio:
Fixes #1260


packages/cli/src/telemetry/send-to-firebase.ts
```diff
@@ -21,6 +21,8 @@ interface Arguments {
 }
 
 process.once('message', async function({ uuid, commandData }: Arguments) {
+  const unixWeek = Math.floor(Date.now() / (1000 * 60 * 60 * 24 * 7));
```
spalladino: - How reliable is this calculation? How much offset would we end up with in a few months time?
- How easy will it be to visualize later? Should we aim to label these as `YYYY-WW` instead, for readability? Or is it easy to generate this from the `unixWeek` value from firebase?

spalladino:
I think this may be needing a merge due to the failing test in the CI

packages/cli/src/telemetry/send-to-firebase.ts
frangio: I'm not sure I understand the question about reliability or offset. :thinking: Do you mean because of leap seconds, or floating point errors, or...? Leap seconds should not affect this because they're accounted for in [Unix time](https://en.wikipedia.org/wiki/Unix_time). This should precisely represent any week.

For visualization we had to convert back to a datetime in the BigQuery SQL code: `TIMESTAMP_SECONDS(unixWeek * 7 * 24 * 60 * 60) as week`. This worked fine, but it may be simpler to do if we store them as `YYYY-WW` or even `YYYY-MM-DD`. It really depends on what will be easier to do in an SQL query. I would have to read more about it.

spalladino: Now that I think of it, using Unix weeks instead of calendar weeks would be more accurate for statistical purposes (since we don't have longer and shorter weeks), so let's go with that!

frangio:
:raised_hands: Thanks for reviewing!

PR #1271, opened 2019-10-29
Force compilation to fix tests

frangio:
Fixes issues reported by @abcoathup in #1241.

I'm not sure why the test script was only compiling in the CI before. I suspect it was because it doesn't work without `rm -rf`-ing the `build/contracts` directory first... And this wasn't necessary in the CI since it's a clean environment.

### DIFF
--- a/packages/cli/scripts/test.sh
+++ b/packages/cli/scripts/test.sh
@@ -32,7 +32,9 @@ fi
 
 if [ "$CI" = true ]; then
   node_modules/.bin/truffle version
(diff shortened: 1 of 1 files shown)


spalladino:
> I'm not sure why the test script was only compiling in the CI before

I think the reason was to make tests faster by avoiding the compilation step. But I prefer having to run `truffle test` manually (instead of `npm t`) rather than having tests fail for someone new to the project due to missing steps.

PR #1275, opened 2019-10-31
Upload user environment info in telemetry reports

### LINKED ISSUE #1274: Upload environment info in telemetry reports
Include platform, arch, version of CLI, and versions of relevant dependencies.

spalladino:
Fixes #1274

Screenshot from 2019-10-31 19-25-52

### DIFF
--- a/packages/cli/src/telemetry/index.ts
+++ b/packages/cli/src/telemetry/index.ts
@@ -6,6 +6,7 @@ import envPaths from 'env-paths';
 import mapValues from 'lodash.mapvalues';
 import inquirer from 'inquirer';
 import proc from 'child_process';
+import process from 'process';
 
 import { DISABLE_INTERACTIVITY } from '../prompts/prompt';
 import { Params } from '../scripts/interfaces';
@@ -24,6 +25,16 @@ export type CommandData = Params & {
   network?: string;
 };
 
+export type UserEnvironment = {
+  platform: string;
+  arch: string;
+  nodeVersion: string;
+  cliVersion: string;
+  upgradesVersion: string | null;
+  web3Version: string | null;
+  truffleVersion: string | null;
+}
+
 // We export an object with report and sendToFirebase so that we can stub them in tests.
 export default {
   async report(commandName: string, options: Params, interactive: boolean): Promise<void> {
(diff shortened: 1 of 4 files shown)


packages/cli/src/telemetry/index.ts
```diff
@@ -59,6 +71,9 @@ export default {
 };
 
 async function checkOptIn(interactive: boolean): Promise<GlobalTelemetryOptions | undefined> {
+  // disable via env var for local development
+  if (!!process.env.OPENZEPPELIN_DISABLE_TELEMETRY) return undefined;
```
frangio: This is causing telemetry to be disabled if the env variable is defined to a value other than the emptry string. Is that the behavior that you wanted?

I'm not a fan of this kind of string to boolean conversion, but there seems to be several of these throughout the codebase so I think I'm okay with keeping it.

packages/cli/src/telemetry/index.ts
```diff
@@ -95,6 +110,26 @@ async function checkOptIn(interactive: boolean): Promise<GlobalTelemetryOptions
...
+    truffleVersion: getDependencyVersion('truffle'),
+    web3Version: getDependencyVersion('web3'),
+  };
+}
+
+function getDependencyVersion(dep: string): string | null {
+  try {
+    return require(`${dep}/package.json`).version;
+  } catch {
+    return null;
```
frangio: Is this `null` rather than `undefined` for a specific reason? Curious.

spalladino: Not really. I typically use `null` instead of `undefined` for empty values.

packages/cli/src/telemetry/index.ts
spalladino: Still, being a variable used for development, I wouldn't worry too much

spalladino:
@frangio added!

packages/cli/src/telemetry/index.ts
frangio: Actually I think `undefined` is a much better choice. For example using the question mark in an interface member or function argument makes it potentially `undefined`, but not `null`.

Also default function arguments will be set when the argument is `undefined` and not when it's `null`.

Overall I think `undefined` should be used here. What do you think?

spalladino: Good points. Fixed!

PR #1277, opened 2019-11-04
Bump web3.js to 1.2.2

spalladino:


nventuro:
It was my understanding that the latest web3 version comes with typings that work properly - if that's the case, shouldn't we also drop the dependency on @types/web3?

spalladino:
@nventuro CI is finally green. Can I ask for a re-review?

.circleci/config.yml
```diff
@@ -97,13 +97,6 @@ jobs:
           workspace: examples/lib-simple
           run-yarn: true
 
-  example-lib-complex:
-    <<: *defaults
-    steps:
-      - test:
-          workspace: examples/lib-complex
```
nventuro: I noticed all of lib-complex, including the CI is setup, is gone. Is this related to the web3 upgrade?

packages/cli/package.json
```diff
@@ -122,8 +121,10 @@
     "truffle-config": "1.1.16",
     "underscore": "^1.9.1",
     "uuid": "^3.3.3",
-    "web3": "1.2.1",
-    "web3-utils": "1.2.0"
+    "web3": "1.2.2",
+    "web3-eth": "1.2.2",
```
nventuro: Are these being added so that they are pinned? They seem to be already pinned on the web3 package itself.

packages/lib/src/artifacts/Contract.ts
```diff
@@ -119,7 +94,7 @@ function _wrapContractInstance(schema: any, instance: Web3Contract): Contract {
   instance.at = function(address: string): Contract | never {
     if (!ZWeb3.isAddress(address)) throw new Error('Given address is not valid: ' + address);
     const newWeb3Instance = instance.clone();
-    newWeb3Instance._address = address;
+    newWeb3Instance['_address'] = address;
```
nventuro: Why does this change have to be made? Also, why is setting `options` not enough?

packages/lib/src/artifacts/ZWeb3.ts
```diff
@@ -23,6 +23,13 @@ export interface TxParams {
   gasPrice?: number | string;
 }
 
+// Patch typing for getStorageAt method
```
nventuro:
```suggestion
// Patch typing for getStorageAt method -- see https://github.com/ethereum/web3.js/pull/3180
```

tests/kits/test/src/integration.test.js
```diff
@@ -20,13 +19,14 @@ function runIntegrationTest({ kit }, action) {
...
 
   it('unpack kit', function() {
-    // have to replace oz with local version so we test current build
     run(`node ../../../packages/cli/lib/bin/oz-cli.js unpack ${kit}`);
   });
 
   // have to replace oz with local version so we test current build
   it('replace oz with local version', function() {
-    run(`${findUp.sync('node_modules/.bin/lerna')} bootstrap --scope=tests-cli-kits --scope="@openzeppelin/*"`);
+    const version = require('../package.json').version;
```
nventuro: Is this a leftover from the yarn PR?

.circleci/config.yml
```diff
@@ -25,14 +25,14 @@ commands:
     steps:
       - checkout
       - restore_cache:
-          key: dependency-cache-v4-{{ checksum "yarn.lock" }}
+          key: dependency-cache-v5-{{ checksum "yarn.lock" }}
```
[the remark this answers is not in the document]
spalladino: In case you want to reset the cache because "something-went-wrong", it's the only option you have in Circle.

.circleci/config.yml
spalladino: Yep, lib-complex was using outdated versions of everything - it even pulled `zos-lib` and `web3@beta.37`. I spent a few hours trying to upgrade it, and eventually decided to cut it.

packages/cli/package.json
spalladino: This is what yarn failed to properly hoist. On a clean install, I ended up with web3 1.2.2, web3-eth 1.2.2, but web3-eth-contract 1.2.1. We have a transitive dependency to web3 1.2.1, which is installed within the package that requires it, but that's not the case for web3-eth-contract.

packages/lib/src/artifacts/Contract.ts
spalladino: `_address` is not public in the Contract type, and its handled as a separate property

tests/kits/test/src/integration.test.js
spalladino: I would have expected this to be changed on that PR, but I did it here as I was reviewing problems with compilation at this step. I'm not sure it's necessary as part of the upgrade, but I think it's nice to have everything on the same toolchain.

packages/cli/package.json
nventuro: Ah, I see. It'd be great if we were able to document this here. Shame Crockford decided to disallow comments

PR #1288, opened 2019-11-21
Migrates SDK CLI tests to `test-environment`

ylv-io:
This makes all tests in `cli` run using `test-environment` with Mocha and Chai instead of `truffle test`. On my machine, a clean run using this setup takes 3min 38s, while `master` takes 5min 02s. There is a massive difference on the time until the first test starts executing.

This PR is similar to a PR by @nventuro [Migrate Lib tests](https://github.com/OpenZeppelin/openzeppelin-sdk/pull/1290).

Environment | Truffle Test | Test Environment | % Gain
-------------|-------------|------------------| ----------
Local              | 4:35             | 3:38                         | ~21
CircleCI         | [5:33](https://circleci.com/gh/OpenZeppelin/openzeppelin-sdk/10111)                        | [4:59](https://circleci.com/gh/OpenZeppelin/openzeppelin-sdk/10131?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link) | ~10

## Migration
* Mocha, Chain and @openzeppelin/test-environment dependencies added.
* Truffle `contract` statements replaced by `describe`.
* Tests fixed.

I had to fix some tests due to differences how `truffle test` and `test-env` operate but they are rather small and trivial.
Screen Shot 2019-11-25 at 11 27 23 PM

image

PS Commit history is a mess. Sorry for that but it is possible to review only file changes.


packages/cli/package.json
```diff
@@ -57,6 +58,7 @@
   "dependencies": {
     "@openzeppelin/resolver-engine-core": "^0.3.3",
     "@openzeppelin/resolver-engine-imports-fs": "^0.3.3",
+    "@openzeppelin/test-environment": "^0.1.0-rc.1",
```
nventuro: You should now be able to upgrade to 0.1.0

packages/cli/test/models/ContractManager.test.js
```diff
@@ -2,13 +2,15 @@
...
 
+import { expect } from 'chai';
+
 import sinon from 'sinon';
 import { FileSystem, Contracts } from '@openzeppelin/upgrades';
 import ContractManager from '../../src/models/local/ContractManager';
 import ProjectFile from '../../src/models/files/ProjectFile';
 import ConfigManager from '../../src/models/config/ConfigManager';
 
-contract('ContractManager', function([_, from]) {
```
nventuro: The linter didn't catch that `from` was unused? We should review its config

packages/cli/test/models/ContractManager.test.js
```diff
@@ -61,14 +63,16 @@ contract('ContractManager', function([_, from]) {
...
+              this.fooContractPath = `${this.testDir}/build/contracts/Foo.json`;
+              FileSystem.writeJson(this.fooContractPath, builtContract);
               this.projectFile = new ProjectFile(`${this.testDir}/zos.json`);
               this.contractManager = new ContractManager(this.projectFile);
               sinon.stub(ConfigManager, 'getBuildDir').returns(`${this.testDir}/build/contracts`);
               sinon.stub(Contracts, 'getLocalContractsDir').returns(`${this.testDir}/contracts`);
             });
 
             afterEach(function() {
+              FileSystem.remove(this.fooContractPath);
```
nventuro: Why is this needed? Is is related to `test-environment` or `mocha`?

packages/cli/test/models/Dependency.test.js
```diff
@@ -2,11 +2,16 @@
...
 
+import { expect } from 'chai';
+import { accounts } from '@openzeppelin/test-environment';
+
 import sinon from 'sinon';
 import npm from 'npm-programmatic';
 import Dependency from '../../src/models/dependency/Dependency';
 
-contract('Dependency', function([_, from]) {
+const [from] = accounts;
```
nventuro: I think we should move these inside the `describe` block, to keep a style similar to the one `truffle test` uses and make these easier to read. This is what we're doing [in the contracts repo](https://github.com/OpenZeppelin/openzeppelin-contracts/pull/2007#discussion_r349930794)

packages/cli/test/scripts/balance.test.js
```diff
@@ -2,16 +2,19 @@
...
 
 import balance from '../../src/scripts/balance';
 
 const ERC20 = Contracts.getFromLocal('ERC20Fake');
 const ERC20Detailed = Contracts.getFromLocal('ERC20FakeDetailed');
 
-contract('balance script', function(accounts) {
-  const [_, accountAddress] = accounts.map(utils.toChecksumAddress);
+describe('balance script', function() {
+  const [accountAddress] = accounts.map(utils.toChecksumAddress);
```
nventuro:
```suggestion
  const [accountAddress] = accounts;
```

`test-environment` exports checksummed accounts, so we can drop this

nventuro: Note that multiple test files follow this same pattern.

packages/cli/test/scripts/transfer.test.js
```diff
@@ -58,8 +62,8 @@ contract('transfer script', function(accounts) {
           unit: 'ether',
         });
 
-        (await ZWeb3.getBalance(sender)).should.not.eq((100e18).toString());
-        (await ZWeb3.getBalance(receiver)).should.eq((110e18).toString());
+        (await ZWeb3.getBalance(sender)).should.not.eq(new BN(senderBalance).minus(10e18).toString());
```
nventuro: Shouldn't this be `should.eq` (without the `not`)? The old test was checking simply that the balance had changed, but it looks like you're trying to assert that it was reduced by the correct amount/

It won't match exactly however because the transaction cost gas. A quick hack for this is to make the transaction have a gas price of 0, but that may not work on all clients.

packages/cli/package.json
```diff
@@ -7,8 +7,9 @@
   "license": "MIT",
   "scripts": {
     "copy-files": "./scripts/copy-files.sh",
-    "compile-ts": "rm -rf lib && tsc",
-    "prepare": "npm run compile-ts && npm run copy-files && chmod 755 ./lib/bin/oz-cli.js",
+    "compile:ts": "rm -rf lib && tsc",
```
spalladino: Let's keep `compile-ts` to have the same command name in both lib and cli, in order to avoid confusion. We can later add a PR in which we rename all scripts consistently.

packages/cli/package.json
```diff
@@ -7,8 +7,9 @@
...
+    "compile:contracts": "rm -rf build/contracts && truffle compile",
```
spalladino: Same here, let's stick to the same naming as lib.

packages/cli/test/scripts/balance.test.js
ylv-io: Nice. I was thinking we should do checksum at `test-env` but it turns out we already do it.

packages/cli/test/models/ContractManager.test.js
ylv-io: Not related to `test-env` or `mocha` but it was causing tests to fail because `Foo.json` would be left in the file system and subsequent Compiler tests would fail because `Foo.json` is not a real artifact. I have no idea why doesn't fail on CircleCI.

nventuro: Strange :man_shrugging: Removing it does seem like the way to go though

PR #1290, opened 2019-11-21
Migrate lib tests to test-environment

nventuro:
This makes all tests in `lib` run using `test-environment` (with mocha and chai) instead of `truffle test`. On my machine, a clean run using this setup takes 1min 47s, while `master` takes 2min 39s. There is a massive difference on the time until the first test starts executing.

edit: there also seems to be a speedup on Circle: from [4min 32s](https://app.circleci.com/jobs/github/OpenZeppelin/openzeppelin-sdk/9984) on `master` to [2 min 22s](https://app.circleci.com/jobs/github/OpenZeppelin/openzeppelin-sdk/10014) here.

Commits are split between migration (the first ones) and slight modifications of some tests that were lightly coupled to the truffle setup. The latter (which all start with 'Fix ...') are best reviewed individually.

### Migration
This was rather simple:
 * truflfe's `contract` was replaced by `describe`, and the accounts obtained via `test-environment`
 * missing imports of chai, lodash, etc. (which are made globally available by truffle) were added
Because all files import `setup.js`, it is not necessary to `require` `test-environment` in every single file: only those that access the exports (e.g. `accounts` or `provider`). We could have `setup` export `accounts`, `provider`, and chai's `assert`, `expect` and `should`, but I didn't want to be too disruptive.

The old test script was quite vanilla, so removing it was not an issue. Note that it did configure `ts-node/register` to use `tsconfig.test.json`, which we now do manually on the mocha invocation. Also, because we no longer have snapshots reverting blockchain state (automatically injected by truffle), I had to increase the initial ETH balance of all accounts, which wasn't enough to cover for the execution of the whole suite.

A benefit of `npm test` calling mocha directly is that we can pass arguments to it: `npm test -- --bail` will run mocha with the `bail` flag, causing tests to immediately abort on the first failure, which is useful when debugging.

Note that we still depend on truffle and have a `truffle-config.js` file because it is being used for compilation.

### Test fixes
These are quite self-explanatory when reviewed commit-by-commit. I took the liberty of introducing a dependency on `openzeppelin/test-helpers` for a test that dealt with chances in balance, which are tricky to handle with the tools currently on the suite. 

This also made me note that said tools are quite varied: in some places we are e.g. still using `bignumber.js` instead of `BN.js`, the big number library currently exported by web3 (and used by most tools, including the test-helpers). Not a big deal, but it'd be nice to eventually improve this.

### What test-environment brings to the project
Mostly speed, specially when using the `.only` directive. Because `lib` has its own contract loader, this feature of `test-enviroment` is not being used here, leading to some duplication and possibly confusion (e.g. the custom contract format has its own configuration, independent of that of the library).

Given that this custom contract object doesn't add much (and should be replaceable by [`contract-loader`](https://github.com/OpenZeppelin/contract-loader)), it might make sense to eventually get rid of it.


nventuro:
Because we're not using truffle contracts, we don't need the `--exit` flag on mocha. Removing it revealed that the Logger tests were leaving remnants on the event loop, preventing the suite from properly exiting. I patched them so that all spinners are properly disposed.

yarn.lock
```diff
@@ -1261,11 +1313,75 @@
   dependencies:
     defer-to-connect "^1.0.1"
 
+"@truffle/blockchain-utils@^0.0.15":
```
spalladino: It seems we are pulling in several oldish truffle packages. Can we avoid this somehow? It's not critical though, since this is only for development.

yarn.lock
```diff
@@ -12608,6 +13104,15 @@ wcwidth@^1.0.0:
   dependencies:
     defaults "^1.0.3"
 
+web3-bzz@1.0.0-beta.35:
```
spalladino: Can we trace where does this web3 beta35 come from? I'd like to avoid introducing these dependencies, even if it's for development.

yarn.lock
```diff
@@ -13143,6 +14129,15 @@ websocket@1.0.29:
     typedarray-to-buffer "^3.1.5"
     yaeti "^0.0.6"
 
+"websocket@git://github.com/frozeman/WebSocket-Node.git#browserifyCompatible":
```
spalladino: No chance in hell I'm approving this PR with this package in it :angry:

packages/lib/package.json
```diff
@@ -6,8 +6,7 @@
   "private": false,
   "license": "MIT",
   "scripts": {
-    "test": "./scripts/test.sh",
-    "test-contracts": "TS_NODE_PROJECT='tsconfig.test.json' npx mocha --exit --require ts-node/register --recursive test/contracts",
+    "test": "TS_NODE_PROJECT='tsconfig.test.json' npx mocha --exit --timeout 5000 --require ts-node/register --recursive test",
```
spalladino: Can we set the timeout explicitly only for the tests that we _know_ are slow, instead of for the entire suite?

nventuro: Turns out only a single test required it, and mocha provides a handy way of configuring it by calling `.timeout` on the return value of `it`:

```javascript
it('is slow', () => {
}).timeout(10000);
```

yarn.lock
nventuro: These are actually not old, and are dependencies of `@truffle/contract`, which `test-environment` depends on.

yarn.lock
nventuro: These came from our fork of `ganache-core` for coverage runs (https://github.com/OpenZeppelin/ganache-core-coverage), which the beta version of `test-environment` we were using depended on. I updated it to the actual release and these are now gone.

yarn.lock
nventuro: These also came from the `ganache-core-coverage` fork

nventuro:
I updated our dependencies and improved the mocha timeout, this is probably ready to be merged now :tada: 

Btw instead of using `npm-why` you can run `yarn why <package>`, which looks quite nice and is yarn-workspace aware.

nventuro:
CI was failing due to our jobs not picking up the cache created by yarn in the setup stage. @ylv-io I suspect the same thing is happening on https://github.com/OpenZeppelin/openzeppelin-sdk/pull/1288

The cause of this was Circle running an old yarn version, which on installation made some small changes to the lockfile. These changes do not happen on our development machines. The cache name would be computed as a checksum of this modified lockfile. Since the other jobs pull the original lockfile from the repository, they looked for a different cache name, which didn't exist.

I made a couple improvements to our CI setup to fix this:
 1. an `engines` entry in our package.json will prevent old yarn versions from interacting with the package (you may need to update your local yarn installation)
 2. I updated the docker image Circle is using to one that uses the latest yarn version
 3. I added a simple script (which I copied from [here](https://github.com/yarnpkg/yarn/issues/5840#issuecomment-467516207)) to check that installing the packages did not modify the lockfile

These three changes should prevent these sort of errors in the future, and fail loudly and clearly on configurations that would lead to them.

frangio:
@nventuro Did you try the `--frozen-lockfike` flag by itself? I'm curious if it wouldn't have fixed the issue.

nventuro:
>@nventuro Did you try the --frozen-lockfike flag by itself? I'm curious if it wouldn't have fixed the issue.

We did, but it doesn't work. There's a long standing open issue on the yarn repo: https://github.com/yarnpkg/yarn/issues/5840

PR #1291, opened 2019-11-22
Use solidity fuzzy import parser

spalladino:
Increases performance of nothing-to-compile runs by using a custom fuzzy parser that only extracts imports from the text, instead of the full antlr4 parser.

Running both versions in the contracts repository, this yields a 40% speed increase when having nothing to compile.

This commit also lazy-loads truffle-flattener, which was causing significant delays to the CLI startup.

### DIFF
--- a/packages/cli/src/utils/solidity.ts
+++ b/packages/cli/src/utils/solidity.ts
@@ -1,5 +1,6 @@
 import { eq as semverEq, parse as parseSemver } from 'semver';
 import { CompilerVersionOptions } from '../models/compiler/solidity/SolidityContractsCompiler';
+import { getMetadata } from '@openzeppelin/fuzzy-solidity-import-parser';
 
 export function getPragma(source: string): string {
   if (!source) return null;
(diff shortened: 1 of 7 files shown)


spalladino:
We should wait until https://github.com/frangio/fuzzy-import-parser/issues/2 is implemented before merging

packages/cli/src/utils/solidity.ts
```diff
@@ -35,15 +36,9 @@ export function compilerSettingsMatch(s1: CompilerVersionOptions, s2: CompilerVe
...
-  const importedFiles: string[] = [];
-  parser.visit(ast, {
-    ImportDirective: (node: { path: string }) => importedFiles.push(node.path),
-  });
-
-  return importedFiles;
+  try {
+    return getMetadata(source).imports;
+  } catch {
+    return [];
```
spalladino: I'd output a warning at least, just in case that the error is different

nventuro: Wouldn't that be a bug in the parser? It shouldn't crash when fed valid Solidity source

spalladino:
Indeed. That's why I want to be able to report it and not silently fail.

nventuro:
@spalladino Once CI passes, this should be ready to merge. I extracted the error handling out of `getImports` to make it more explicit, and added calls to `trimURL` whenever we report a problematic file location that may have been the result of buggy parsing.

spalladino:
@nventuro given I'm the author, I cannot approve the PR. Please approve it and merge, I think it's ok!

PR #1293, opened 2019-11-27
Use CLI when compiling lib contracts

nventuro:
We're currently using `truffle` to compile the contracts that are used as part of the `upgrades` package (in its `prepare` script): it'd be best to migrate to our own CLI compiler. However, because the compiler itself depends on lib, this introduces a circular dependency.

This PR works around that issue by bootstraping the build process with a global installation of a previous version of the CLI. The CircleCI docs are not very clear regarding how this should be done, so I'm opening this as a draft while I experiment on the PR's build.


frangio:
Superceded by #1294. We thought that requiring a globally installed OZ CLI made the development setup too cumbersome by not being self-contained.

PR #1296, opened 2019-11-28
Switch to identifying contracts by full path

frangio:
The build in https://github.com/OpenZeppelin/openzeppelin-sdk/commit/855c2ad6a0f43737dfeb6bf98a99c00448c88c31 failed due to corrupted JSON artifacts. This was caused by writing the artifacts of two separate contracts into the same file concurrently, which itself was caused by using the basename instead of the full path to the Solidity files.


nventuro:
How is this related to https://github.com/OpenZeppelin/openzeppelin-sdk/issues/1203? Will the second contract overwrite the first one?

frangio:
I'm not sure it's related and I don't think we've fixed the issue.

PR #1299, opened 2019-12-02
Disable interactivity if stdin is not a terminal

frangio:
It doesn't make sense to have interactivity if stdin is not a terminal. We can check this using `tty.isatty`.

### DIFF
--- a/packages/cli/src/prompts/prompt.ts
+++ b/packages/cli/src/prompts/prompt.ts
@@ -1,4 +1,3 @@
-import tty from 'tty';
 import uniqBy from 'lodash.uniqby';
 import flatten from 'lodash.flatten';
 import isEmpty from 'lodash.isempty';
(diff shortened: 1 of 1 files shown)


packages/cli/src/prompts/prompt.ts
```diff
@@ -53,10 +54,21 @@ interface MethodOptions {
 }
 
 export let DISABLE_INTERACTIVITY: boolean =
+  !isStdinATerminal() ||
```
jcarpanelli: I believe that `process.stdin` has a `isTTY` variable setted as `true`/`false` as appropriate, so you can avoid implementing the function in `L62` and just do:
```suggestion
  !process.stdin.isTTY ||
```

frangio: :heart_eyes: Nice!

frangio:
Pending: fix the failing tests, which are probably assuming that interactivity is on during testing.

nventuro:
The tests were magically fixed :man_shrugging:

frangio:
Huh. I triggered a rerun just in case.

PR #1304, opened 2019-12-04
Restrict version range of Upgrades allowed to be used with CLI

frangio:
The current version ranges allow an installation of `@openzeppelin/cli` to use a newer version of `@openzeppelin/upgrades`. Since we don't test these combinations, we feel it would be best to restrict the allowed version range.


spalladino:
Can you check how lerna behaves when we run `lerna version` to bump versions? I recall that there is a `--exact` option that you can supply so dependencies are matched by exact version identifiers and not carets, but not sure if there is an option to use tilde. Also, it'd be interesting to see if we can set that as a setting in the `lerna.json` as well.

frangio:
Good catch. Lerna replaces the `~` with a caret. :man_facepalming: The only alternative is using `--exact`.

PR #1312, opened 2019-12-11
Update faq.adoc for solc version in oz compile

abcoathup:
`openzeppelin compile --solc-version 5.X` should be `openzeppelin compile --solc-version 0.5.X`
Raised by a community member in Telegram.

I have specified the (current) latest version of solc 0.5.13.


nventuro:
Good catch, thanks!

0.5.14 [was just released](https://github.com/ethereum/solidity/releases), defaulting to Instanbul, so I'd recommend that one instead.

abcoathup:
Thanks.  
Updated examples to solc `0.5.14`.

PR #1333, opened 2019-12-13
Add tuple support in interactive prompts

### LINKED ISSUE #1330: Support tuples in interactive commands
Trying to encode a tuple in an interactive command currently fails. We should add support for tuples, which are handled via the ABIEncoderV2.

```
~/Projects/sdk/examples/foo$ oz call
? Pick a network development
? Pick an instance Greeter at 0xA57B8a5584442B467b4689F1144D269d096A3daF
? Select which function foo(f: tuple)
? f (tuple): (10,"foo")
Unknown argument tuple (skipping input validation)
invalid tuple value (coderType="tuple", value="(10,\"foo\")", version=4.0.39)
```

Sample contract:

```solidity
pragma experimental ABIEncoderV2;
contract Greeter {
  struct Foo { uint256 value; string text; }
  function foo(Foo memory f) public pure returns (Foo memory) {
    return f;
  }
}
```

spalladino:
Fixes #1330. Tuples can be entered as comma separated, or as comma separated enclosed in parenthesis.

Screenshot from 2019-12-13 17-26-22

### DIFF
--- a/packages/cli/src/prompts/method-params.ts
+++ b/packages/cli/src/prompts/method-params.ts
@@ -59,18 +59,22 @@ function getCommandProps(
     return {
       ...accum,
       [arg.name]: {
-        message: arg.name ? `${arg.name} (${arg.type}):` : `(${arg.type}):`,
+        message: arg.name ? `${arg.name} (${arg.type}):` : `(${arg.type}):`, // TODO: better description for tuples
         type: 'input',
         when: () => !methodArgs || !methodArgs[index],
-        validate: input => {
+        validate: (input: string) => {
           try {
-            parseArg(input, arg.type);
+            parseArg(input, arg);
             return true;
           } catch (err) {
-            return `${err.message}. Enter a valid ${arg.type} such as: ${getPlaceholder(arg)}.`;
+            const placeholder = getPlaceholder(arg);
+            const msg = placeholder
+              ? `Enter a valid ${arg.type} such as: ${getPlaceholder(arg)}`
+              : `Enter a valid ${arg.type}`;
+            return `${err.message}. ${msg}.`;
           }
         },
-        normalize: input => parseArg(input, arg.type),
+        normalize: input => parseArg(input, arg),
       },
     };
   }, {});
--- a/packages/cli/src/prompts/method-params.ts
+++ b/packages/cli/src/prompts/method-params.ts
@@ -3,7 +3,7 @@ import pickBy from 'lodash.pickby';
 import isUndefined from 'lodash.isundefined';
 import negate from 'lodash.negate';
 
-import { parseMethodParams, parseArg } from '../utils/input';
+import { parseMethodParams, parseArg, getPlaceholder } from '../utils/input';
 import { promptIfNeeded, argsList, methodsList, InquirerQuestions } from './prompt';
 
 type PropsFn = (any) => InquirerQuestions;
@@ -97,31 +97,4 @@ function getCommandProps(
     },
     ...args,
   };
-}
-
-function getPlaceholder(type: string): string {
-  const ARRAY_TYPE_REGEX = /(.+)\[\d*\]$/; // matches array type identifiers like uint[] or byte[4]
-
-  if (type.match(ARRAY_TYPE_REGEX)) {
-    const arrayType = type.match(ARRAY_TYPE_REGEX)[1];
-    const itemPlaceholder = getPlaceholder(arrayType);
-    return `[${itemPlaceholder}, ${itemPlaceholder}]`;
-  } else if (
-    type.startsWith('uint') ||
-    type.startsWith('int') ||
-    type.startsWith('fixed') ||
-    type.startsWith('ufixed')
-  ) {
-    return '42';
-  } else if (type === 'bool') {
-    return 'true';
-  } else if (type === 'bytes') {
-    return '0xabcdef';
-  } else if (type === 'address') {
-    return '0x1df62f291b2e969fb0849d99d9ce41e2f137006e';
-  } else if (type === 'string') {
-    return 'Hello world';
-  } else {
-    return null;
-  }
 }
--- a/packages/cli/src/prompts/method-params.ts
+++ b/packages/cli/src/prompts/method-params.ts
@@ -67,7 +67,7 @@ function getCommandProps(
             parseArg(input, arg.type);
             return true;
           } catch (err) {
-            return `${err.message}. Enter a valid ${arg.type} such as: ${getPlaceholder(arg.type)}.`;
+            return `${err.message}. Enter a valid ${arg.type} such as: ${getPlaceholder(arg)}.`;
           }
         },
         normalize: input => parseArg(input, arg.type),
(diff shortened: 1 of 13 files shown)


packages/cli/src/utils/input.ts
```diff
@@ -238,3 +250,35 @@ export function validateSalt(salt: string, required = false) {
     throw new Error(`Invalid salt ${salt}, must be an uint256 value.`);
   }
 }
+
+export function getPlaceholder(arg: Pick<MethodArg, 'type' | 'components'>): string {
```
frangio: I don't think "placeholder" is very accurate. What about `getSampleInput`?

packages/cli/src/utils/input.ts
```diff
@@ -46,18 +48,24 @@ function quoteArguments(args: string) {
   return args;
 }
 
-export function parseArg(input: string | string[], type: string): any {
+export function parseArg(input: string | string[], { type, components }: Pick<MethodArg, 'type' | 'components'>): any {
```
frangio: Why not use the entire `MethodArg` type here? I imagine that using `Pick` simplifies the tests? Is there another reason?

If we insist on not using the entire type, I think we should define a type alias in this file. It represents an important concept, and it's used twice in the file.

packages/cli/src/utils/input.ts
```diff
@@ -46,18 +48,24 @@ function quoteArguments(args: string) {
...
   const TRUE_VALUES = ['y', 'yes', 't', 'true', '1'];
   const FALSE_VALUES = ['n', 'no', 'f', 'false', '0'];
   const ARRAY_TYPE_REGEX = /(.+)\[\d*\]$/; // matches array type identifiers like uint[] or byte[4]
 
-  // TODO: Handle tuples in ABI specification
+  // Tuples: recursively parse
+  if (type === 'tuple') {
+    const inputs = typeof input === 'string' ? parseArray(stripParens(input), '(', ')') : input;
```
frangio: I found it weird that `parseArray` wouldn't handle the delimiters itself. What do you think about doing that?

Also, why do we allow the input to _not_ be surrounded by delimiters? In what cases is that used? Does that apply to tuples in the same way that it applies to arrays?

packages/cli/src/prompts/method-params.ts
```diff
@@ -59,18 +59,20 @@ function getCommandProps(
     return {
       ...accum,
       [arg.name]: {
-        message: arg.name ? `${arg.name} (${arg.type}):` : `(${arg.type}):`,
+        message: argLabel(arg),
```
frangio: Note that the previous string had a colon at the end, and `argLabel` doesn't (which is okay because in the other place where it's used there was no colon).
```suggestion
        message: argLabel(arg) + ':',
```

packages/cli/src/utils/input.ts
frangio: Out of curiosity, in what cases is `input` an array?

packages/cli/src/utils/input.ts
```diff
@@ -46,18 +48,24 @@ function quoteArguments(args: string) {
...
+    if (inputs.length !== components.length)
+      throw new Error(`Expected ${components.length} values but got ${input.length}`);
+    return zipWith(inputs, components, parseArg);
```
frangio: Hehe this is nice.

packages/cli/src/utils/input.ts
```diff
@@ -261,8 +274,10 @@ export function getPlaceholder(type: string): string {
     return '0x1df62f291b2e969fb0849d99d9ce41e2f137006e';
   } else if (type === 'string') {
     return 'Hello world';
+  } else if (type === 'tuple' && components) {
```
frangio: Are there cases when `components` is undefined?

packages/cli/src/utils/input.ts
frangio: The compiler isn't catching this because we're not running it in strict mode. It will be a bit of an undertaking to enable it, but I would highly recommend we do.

```suggestion
export function getPlaceholder(arg: Pick<MethodArg, 'type' | 'components'>): string | null {
```

packages/lib/src/utils/ABIs.ts
```diff
@@ -66,9 +66,17 @@ export function getABIFunction(contract: Contract, methodName: string, args: any
   }
 }
 
+export function getABIType(arg: any): string {
```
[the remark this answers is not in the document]
spalladino: Good point on the naming. Changed it to `getArgTypeLabel`.

packages/cli/src/utils/input.ts
spalladino: Like it! The original name was placeholder because the idea was to use it as a placeholder when presenting the question, but it was after that we noted that inquirer does not support placeholders (it only displays them when they are the default value).

packages/cli/src/utils/input.ts
spalladino: The problem was in the recursive call, since the types for an array (for example) are not named. I agree with giving it a proper name, just created `MethodArgType` for that.

packages/cli/src/utils/input.ts
spalladino: > I found it weird that parseArray wouldn't handle the delimiters itself. What do you think about doing that?

Good point, I think it could be refactored. That said, given that this routine is pretty isolated (and working!), I wouldn't risk changing it now.

> Also, why do we allow the input to not be surrounded by delimiters? In what cases is that used? Does that apply to tuples in the same way that it applies to arrays?

When the user is prompted to enter an array, they can just enter a comma separated list, without the brackets, for simplicity. Same applies for tuples.

packages/cli/src/prompts/method-params.ts
spalladino: Good catch!

packages/cli/src/utils/input.ts
spalladino: IIRC, `parseArray` does not parse just the top-level arrays, but it returns arrays of arrays (as many levels deep as needed). So, the `parseArg` recursive call in the case of arrays may be done with an array instead of a string.

And writing this I'm thinking that maybe arrays of tuples (or viceversa) may not work...

packages/cli/src/utils/input.ts
spalladino: There shouldn't if type is tuple, but given I found no docs or guarantees about it, I preferred to err on the safe side.

packages/cli/src/utils/input.ts
spalladino: Good one. And agree on the strict mode!

spalladino:
@frangio updated!

packages/lib/src/utils/ABIs.ts
```diff
@@ -137,5 +146,6 @@ export function callDescription(method: any, args: string[]): string {
 export default {
   buildCallData,
   getABIFunction,
+  getABIType: getArgTypeLabel,
```
frangio: Can't we remove this here? It didn't exist before this PR so there are no backwards compatibility issues.

spalladino: Yes, that was the auto-rename backfiring - sorry about it

spalladino:
@frangio updated!

PR #1335, opened 2019-12-13
Warn when importing vanilla contracts package

### LINKED ISSUE #1297: Warn when creating a contract importing from `@openzeppelin/contracts`
Warn the user when creating a contract importing from `@openzeppelin/contracts` with a reference to the documentation: https://docs.openzeppelin.com/sdk/2.5/linking#linking-the-contracts-ethereum-package 

(We may need to move the Note about using @openzeppelin/contracts-ethereum-package` so that it is visible when viewing the link)

So the warning could be something like the following:
```
oz create
...
Contract Test imports `@openzeppelin/contracts`. Change to `@openzeppelin/contracts-ethereum-package`. 
See https://docs.openzeppelin.com/sdk/2.5/linking#linking-the-contracts-ethereum-package.
```

Similar to the warning for an explicit constructor:
```
Contract Test has an explicit constructor. Change it to an initializer function. 
See https://docs.openzeppelin.com/sdk/2.5/writing_contracts.html#initializers.
```

spalladino:
Users should only install contracts from oz/contracts-eth-package, not oz/contracts. Warn about this when an import for the vanilla contracts package is found.

Fixes #1297

### DIFF
--- a/packages/cli/src/interface/ValidationLogger.ts
+++ b/packages/cli/src/interface/ValidationLogger.ts
@@ -11,11 +11,12 @@ import {
 } from '@openzeppelin/upgrades';
 import { ContractInterface } from '../models/files/NetworkFile';
 
-const DOCS_HOME = 'https://docs.openzeppelin.com/sdk/2.5';
+const DOCS_HOME = 'https://docs.openzeppelin.com/sdk/2.6';
 const DANGEROUS_OPERATIONS_LINK = `${DOCS_HOME}/writing-contracts#potentially-unsafe-operations`;
 const AVOID_INITIAL_VALUES_LINK = `${DOCS_HOME}/writing-contracts#avoid-initial-values-in-field-declarations`;
 const INITIALIZERS_LINK = `${DOCS_HOME}/writing-contracts#initializers`;
 const STORAGE_CHECKS_LINK = `${DOCS_HOME}/writing-contracts#modifying-your-contracts`;
+const VANILLA_CONTRACTS_LINK = `${DOCS_HOME}/linking#linking-the-contracts-ethereum-package`;
 
 export default class ValidationLogger {
   public contract: Contract;
@@ -39,6 +40,7 @@ export default class ValidationLogger {
       uninitializedBaseContracts,
       storageDiff,
       storageUncheckedVars,
+      importsVanillaContracts,
     } = validations;
 
     this.logHasConstructor(hasConstructor);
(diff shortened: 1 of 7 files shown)


packages/lib/src/validations/VanillaContracts.ts
```diff
@@ -0,0 +1,19 @@
...
+import ContractAST from '../utils/ContractAST';
+
+let VANILLA_CONTRACTS = '@openzeppelin/contracts/';
+
+export function importsVanillaContracts(contract: Contract, buildArtifacts?: any): string[] | null {
+  const ast = new ContractAST(contract, buildArtifacts, { nodesFilter: ['ContractDefinition'] });
+  const illegalImports = [...ast.getImports()]
+    .filter(i => i.startsWith(VANILLA_CONTRACTS))
+    .map(i => i.slice(VANILLA_CONTRACTS.length))
+    .map(i => i.replace(/^contracts\//, ''));
```
frangio: The `@openzeppelin/contracts` package doesn't have the `contracts` directory in it. This line shouldn't be necessary.

packages/lib/src/validations/VanillaContracts.ts
```diff
@@ -0,0 +1,19 @@
...
+
+  return illegalImports.length > 0 ? illegalImports : null;
```
frangio: Why `null` here? (I feel that we've had this conversation before already. 🤔) We should favor `undefined` because it's what's used by TypeScript for optional properties and arguments. This isn't affecting us currently because we're not using TypeScript in strict mode, but we should and will. For example, in this same PR you defined an optional property `importsVanillaContracts?: string[]`, to which you're assigning the result of this function. This wouldn't type in strict mode because of `null`.

packages/lib/src/validations/VanillaContracts.ts
```diff
@@ -0,0 +1,19 @@
+import Contract from '../artifacts/Contract';
+import ContractAST from '../utils/ContractAST';
+
+let VANILLA_CONTRACTS = '@openzeppelin/contracts/';
+
+export function importsVanillaContracts(contract: Contract, buildArtifacts?: any): string[] | null {
+  const ast = new ContractAST(contract, buildArtifacts, { nodesFilter: ['ContractDefinition'] });
```
frangio: Do we need `ContractDefinition` nodes for this?

packages/cli/src/interface/ValidationLogger.ts
```diff
@@ -48,6 +50,19 @@ export default class ValidationLogger {
...
+    this.logImportsVanillaContracts(importsVanillaContracts);
+  }
+
+  public logImportsVanillaContracts(vanillaContracts: string[] | null): void {
+    if (!vanillaContracts) return;
+    Loggy.noSpin.warn(
+      __filename,
+      'logImportsVanillaContracts',
+      `validation-imports-vanilla-contracts`,
+      `- Contract ${this.contractName} imports ${vanillaContracts.join(
```
frangio: A bit of a pedantic comment but I do feel it would be best to be accurate here and talk about files importing other files, rather than contracts importing files.

packages/lib/src/validations/VanillaContracts.ts
spalladino: I know, but since it doesn't hurt, I preferred to leave it here in case we change our minds (again) regarding the folder structure.

packages/lib/src/validations/VanillaContracts.ts
spalladino: > Why null here? (I feel that we've had this conversation before already. thinking)

Because I forgot about the conversation we had already. Thanks for pointing it out, I'll do my best to break the force of habit.

packages/lib/src/validations/VanillaContracts.ts
spalladino: Ouch, copy-pasta error.

packages/cli/src/interface/ValidationLogger.ts
spalladino: There's a problem with that: the validator works on a per-contract basis, not per file. This also means that, if the user has more than one contract defined on the same file that imports vanilla contracts, they'll get the warning twice. 

We could solve this by restructuring how validations work, but given that this is a temporary feature, I preferred to ship anyway. Curious to see what you think though.

spalladino:
@frangio updated!

packages/cli/src/interface/ValidationLogger.ts
frangio: Agree.

Issue #1227, opened 2019-09-11
oz transfer might not be using the gasPrice specified in the networks.js file


jcarpanelli:
Reported by @nventuro. `oz transfer` seems to be ignoring the `gasPrice` specified in the `networks.js` file as follows:

```
module.exports = {
  networks: {
    mainnet: {
      protocol: 'http',
      host: 'localhost',
      port: 8546,
      gas: 5000000,
      gasPrice: 5e9,
      networkId: '1',
    }
  },
};
```

and used `20gwei` instead as shown [here](https://etherscan.io/tx/0x7ffa03d4c8dae0df262cfebe25df7f565089e609566a7fc881fe6a06b12ef21d).

Issue #1231, opened 2019-09-12
Fix uses of pickBy


frangio:
`pickBy` is being used to remove properties with `undefined` or `null` values, but it's being used in a way that zeroes and empty strings are being removed as well.

Example (but there are many others):
https://github.com/OpenZeppelin/openzeppelin-sdk/blob/269f097b368aee7ddfa97cdbc6058a2fc0d746ad/packages/lib/src/utils/Logger.ts#L89

The reason is that the `iteratee` argument in `pickBy` has a default value of `lodash.identity`. The fix is to use `pickBy(..., isNil)` with `isNil = require('lodash.isnil')`.

nventuro:
`isNil` returns true for null-like values, so what we actually want is 

```
pickBy(..., x => !isNil(x))
```

PR #1345, opened 2019-12-16
Load gas and gasPrice from networkConfig

nventuro:
Fixes https://github.com/OpenZeppelin/openzeppelin-sdk/issues/1227

The change itself is rather simple: the only thing I'm unsure is whether we should test this as a unit test, integration, or not at all.

That said, I want to discuss my findings while looking into this issue to provide some context that may help inform future decisions.

`lib` defines a `TxParams` object that is passed around when sending transactions and serves as a default for these. This object is populated by `ConfigManager` from the network configuration file, and before this PR, it only includes `from`. 

Because of this, I was surprised to see that the correct values for `gas` are nonetheless used in all transactions: it turns out `lib/utils/Transactions` handles a null value for `gas` by loading it from `Contracts.getArtifactsDefault`, which acts as a sort of global store of configuration. Notice that it was `ConfigManager` who called `Contracts.setArtifactsDefaults`, duplicating these values.

The bug surfaced in `transfer` because this is the first instance of the SDK sending transactions that are not contract function calls. On those, because the gas price was never specified, web3 used the default one that is set when creating the contract objects.

There seems to be a tension here between global configuration that is accessed directly (either via the `Contracts.setDefaultArtifacts` or the implicit default values used in the web3 objects), and explicit configuration objects that are passed down the call stack. We may want to revisit this at some point.

From the description above, it should be clear that `gas` and `gasPrice` were always obtained from `getArtifactDefault`, when available. Because of this, I'm relatively confident that this PR will not break current behaviour.

### DIFF
--- a/packages/cli/src/models/config/ConfigManager.ts
+++ b/packages/cli/src/models/config/ConfigManager.ts
@@ -3,6 +3,8 @@ import TruffleConfig from './TruffleConfig';
 import Session from '../network/Session';
 import NetworkConfig from './NetworkConfig';
 
+import pick from 'lodash.pick';
+
 const ConfigManager = {
   config: undefined,
 
@@ -38,6 +40,7 @@ const ConfigManager = {
       await ZWeb3.checkNetworkId(network.networkId);
       const txParams = {
         from: ZWeb3.toChecksumAddress(from || artifactDefaults.from || (await ZWeb3.defaultAccount())),
+        ...pick(artifactDefaults, ['gas', 'gasPrice']),
       };
 
       return { network: await ZWeb3.getNetworkName(), txParams };
--- a/packages/cli/src/models/config/ConfigManager.ts
+++ b/packages/cli/src/models/config/ConfigManager.ts
@@ -4,6 +4,8 @@ import Session from '../network/Session';
 import NetworkConfig from './NetworkConfig';
 
 import pick from 'lodash.pick';
+import pickBy from 'lodash.pickby';
+import isNil from 'lodash.isnil';
 
 const ConfigManager = {
   config: undefined,
@@ -40,7 +42,7 @@ const ConfigManager = {
       await ZWeb3.checkNetworkId(network.networkId);
       const txParams = {
         from: ZWeb3.toChecksumAddress(from || artifactDefaults.from || (await ZWeb3.defaultAccount())),
-        ...pick(artifactDefaults, ['gas', 'gasPrice']),
+        ...pickBy(pick(artifactDefaults, ['gas', 'gasPrice']), x => !isNil(x)),
       };
 
       return { network: await ZWeb3.getNetworkName(), txParams };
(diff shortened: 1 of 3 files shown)


nventuro:
The explanation ended up being a bit too wall-of-texty. Sorry about that!

packages/cli/src/models/config/ConfigManager.ts
```diff
@@ -38,6 +40,7 @@ const ConfigManager = {
       await ZWeb3.checkNetworkId(network.networkId);
       const txParams = {
         from: ZWeb3.toChecksumAddress(from || artifactDefaults.from || (await ZWeb3.defaultAccount())),
+        ...pick(artifactDefaults, ['gas', 'gasPrice']),
```
spalladino: Can we change this so these keys are not set if their value is empty? I want to avoid a case where we merge `txParams` using an object splat, which would end up overriding an existing value with `undefined` if we're not careful.

spalladino: As an example, in the following code, if `txParams = { gasPrice: undefined }`, the default `10e9` value would get overwritten with `undefined` (which is not the desired behaviour).

```
function foo(txParams) {
  const txParamsDefaults = { gasPrice: 10e9 }
  const txParamsToUse = { ...txParamsDefaults, ...txParams }
  ...
}
```

spalladino:
> There seems to be a tension here between global configuration that is accessed directly (either via the Contracts.setDefaultArtifacts or the implicit default values used in the web3 objects), and explicit configuration objects that are passed down the call stack. We may want to revisit this at some point.

Agree. This is one of the things I want to fix when reviewing the programmatic API: no more global state.

packages/cli/src/models/config/ConfigManager.ts
nventuro: This is currently not an issue due to how these values are created (empty entries are removed), but I addressed this nonetheless in the style suggested in https://github.com/OpenZeppelin/openzeppelin-sdk/issues/1231.

I also created https://github.com/OpenZeppelin/openzeppelin-sdk/issues/1346 to track this general issue.

packages/cli/src/models/config/ConfigManager.ts
```diff
@@ -38,6 +42,7 @@ const ConfigManager = {
       await ZWeb3.checkNetworkId(network.networkId);
       const txParams = {
         from: ZWeb3.toChecksumAddress(from || artifactDefaults.from || (await ZWeb3.defaultAccount())),
+        ...pickBy(pick(artifactDefaults, ['gas', 'gasPrice']), x => !isNil(x)),
```
spalladino: Personally, I'd have added `negate` as well, and replaced `x => !isNil(x)` by `negate(isNil)`, but maybe it's because I've been haskelling too much in my free time.

nventuro: Ah, I didn't know about `negate`. I updated #1231 to recommend `omitBy(..., isNil)`, but forgot to reflect the change PR.

Issue #1315, opened 2019-12-11
Type and review Loggy's interface


spalladino:
Add typings for Loggy to avoid type errors, and review its interface (eg consider renaming noSpin to something clearer).

frangio:
Have implemented this in a branch. Will open a PR soon.

frangio:
One of the main things we should fix is the reference argument. It shouldn't be necessary to manually specify a string reference id, at least not in the way we use it, which is to later update the spinner state. We should automatically return a value that identifies the spinner that is to be stored in a variable (only if necessary!) that either is an object with methods to update the spinner, or that can be passed as an argument to functions. This avoids the potential for accidental id reuse, and removes the obligation to define a reference when one isn't needed.

I also think we should remove the filename and function name arguments, and instead use Node.js's (i.e., V8's) stack traces to obtain that automatically.

PR #1349, opened 2019-12-17
Add types to Loggy

frangio:
I changed the object into a class with static functions because that way `this` is typed correctly.

I have not reviewed the interface so I'm not marking #1315 as fixed.


frangio:
Actually, I think we should just change it into a module that exports functions.

I think it's currently an object because we stub the functions in the tests. But I think we should stub the output stream instead.

frangio:
We have decided to move forward with this even if we're going to change the API later, to have the types in the meantime.

PR #1351, opened 2019-12-18
Remove ScriptError and ErrorHandler

frangio:
The custom error type `ScriptError` was used to register a callback that would be called when the error was handled by `ErrorHandler`. We can achieve essentially the same effect by using `finally`. The only difference I see is that the callback was being ignored during tests, since `ErrorHandler.prototype.call` was stubbed out, but I don't see why we would want to do this, and I don't know if this consequence was intentional before.

It was also necessary to move the `registerErrorHandler` call to `oz-cli.js`, so that the error handler (which exits the process) is not attached during testing. This used to be stubbed out, but it's not possible to do so now.


frangio:
I can confirm that things happen and in that order!

PR #1366, opened 2019-12-24
Rework upgrade.js example

spalladino:
Drops truffle from the lib-simple example, using vanilla web3 to run the code, and test-env for the tests. Also removes babel, and renames the example from lib-simple to upgrades-library.


examples/upgrades-library/test/main.test.js
```diff
@@ -0,0 +1,15 @@
+const Web3 = require('web3');
+
+const provider = require('@openzeppelin/test-environment').provider;
```
nventuro:
```suggestion
const { provider } = require('@openzeppelin/test-environment');
```

yarn.lock
```diff
@@ -10605,14 +10514,6 @@ truffle@^4.1.14:
     original-require "1.0.1"
     solc "0.4.26"
 
-truffle@^5.0.3:
```
nventuro: :excited:

spalladino:
@nventuro removed build artifacts and added compile step to test. Please re-review!

PR #1369, opened 2019-12-30
Trim ZWeb3

frangio:
This PR trims ZWeb3 by removing all the functions which are just wrappers around web3 functionality. I additionally made `ZWeb3.eth` a getter (it used to be a function) so that the syntax for using it is nicer (without parentheses).

For example `ZWeb3.accounts()` has been removed and it now becomes `ZWeb3.eth.getAccounts()`.

I've also removed many tests that were only testing the wrapped functionality, because they were just testing the web3 dependency. We might want to keep them though, given that web3 is kind of unstable, but IMO they would not be enough to detect any significant issues.

Do we consider this a breaking change?


packages/lib/test/src/artifacts/ZWeb3.test.js
```diff
@@ -114,34 +114,34 @@ describe('ZWeb3', function() {
...
-                  /is not a valid Ethereum address or it has not been checksummed correctly/,
-                );
-            });
-          });
-
-          context('when address has been checksummed properly', function() {
-            it('returns checksummed address', function() {
-              ZWeb3.toChecksumAddress(accounts[0]).should.be.eq(accounts[0]);
-            });
+    describe('checksum address', function() {
```
frangio: This is unrelated to the PR purpose but I was editing this file and noticed that this describe block was incorrectly placed within the previous describe block.

spalladino:
I don't consider this a breaking change, but I'm reluctant to removing many of the methods from ZWeb3 that would allow us to switch to a different library in the future (if we need to). Let's discuss it live.

packages/cli/src/utils/constants.ts
```diff
@@ -1,4 +1,6 @@
-export const ERC20_PARTIAL_ABI = [
+import { AbiItem } from 'web3-utils';
```
spalladino: I love how in every PR you keep adding more typings

PR #1375, opened 2020-01-06
Remove .`gitkeep` as it's not needed. Move `.gitattributes` to the root of the directory

cgcardona:
1. Remove `.gitkeep` as it's no longer needed since that directory isn't empty. 
2. Move `.gitattributes` to the root of the directory structure. I learned about this from [your Contracts repo](https://github.com/OpenZeppelin/openzeppelin-contracts/blob/master/.gitattributes). You can see a [working example here](https://github.com/OpenZeppelin/openzeppelin-contracts/blob/master/contracts/cryptography/ECDSA.sol)


cgcardona:
I see that the `ci/circleci: test-cli` failed w/ a bunch of errors like this:

```
SyntaxError: /home/circleci/openzeppelin/packages/cli/build/contracts/GreeterImpl.json: Unexpected string in JSON at position 23061
```

AFAIK that's unrelated to my change. Can you confirm? Thanks for a OpenZeppelin! I'm truly amazed at how high quality your platform and tooling are. Great work! 🎩

spalladino:
Good catch, thanks for the contribution!

PR #1376, opened 2020-01-07
Non-upgradeable deploys, round 2

### LINKED ISSUE #802: Creating non-upgradeable contracts
A pattern that is coming up from user interviews is that part of their smart contracts are upgradeable, while the core of their system is not. This means that a user that relies on zOS needs to switch to a different tool to work with non-upgradeable contracts.

We can implement three different approaches to this:
1. **Creating a non-upgradeable proxy (ie minimal proxy) to the logic contract**. This is the less disruptive option, since we just work with different "flavors" of proxies, and plays nicely with EVM packages. It does require the user to deploy the logic contract first (`zos push`) and to switch all constructors to initializers in the implementation.
2. **Duplicating the code of the logic contract into a new instance.** This is the most costly option, since it requires two deployments per contract (one of the logic one, which goes unused, and another of the actual instance). However, it is partly implemented already in #32.
3. **Simply deploying the contract from the local artifacts.** This is the most direct approach, allows users to keep working with constructors, and does not require trust in the proxy pattern.

frangio:
Supercedes #1240. Fixes #802.

This is quite a large pull request that adds a new `deploy` command that can be used for deploying both upgradeable and "regular" contracts.

The new command was built following a different architecture than all the current commands, as an experiment to see how we can refactor the entire commands layer to have better separation of concerns and better maintainability. The idea would be to refactor the rest of the codebase if we agree that this is an improvement.

In the current architecture there are command and script layers. The command layer provides a command's specification and action, and the action handles interactivity very explicitly by making use of the prompts module in the action flow. The script layer is very thin and almost all it does is error handling, and printing to stdout, then delegating to a controller.

The proposed architecture removes the scripts layer, and separates the command layer more explicitly into command specification and action. Interactivity is no longer a concern of the action, instead the command specification provides details for the behavior of each positional argument and option, and when the command is registered in the commander program these details are used to run interactive prompts prior to running the action.

The code for registering a command using this new architecture can be found in the `register-command.ts` file, including also the code for running the interactive prompts based on a command spec. There are no tests for this, but I think I should add a few before merging this.

The two parts of the new command are found in `commands/deploy/spec.ts` and `commands/deploy/action.ts`. This separation will allow for optimizing the loading time of the CLI because the dependencies of the action will not be imported unless the command is actually called. There is also `commands/deploy/index.ts` that provides the glue between the proposed architecture and the current architecture.

One of the things that I think should be improved is the naming around `prompt`, `Question`, etc., in the args and opts specification. I kind of picked these names quickly and kept them but they're not good.


packages/cli/src/prompts/prompt.ts
```diff
@@ -232,10 +207,17 @@ function contractMethods(
 ): any[] {
   const { contract: contractAlias, package: packageName } = fromContractFullName(contractFullName);
   const contractManager = new ContractManager(projectFile);
-  if (!contractManager.hasContract(packageName, contractAlias)) return [];
-  const contract = contractManager.getContractClass(packageName, contractAlias);
 
-  return contractMethodsFromAbi(contract, constant);
+  try {
```
frangio: This may seem unrelated, but the reason this was changed is that now `deploy` can be used with contracts from _unlinked_ packages, and there won't be an implicit `add` as in `create`. Because of this, `contractManager.hasContract` is no longer reliable, and I thought it was better to just try to read the artifact from disk and catch the failure. I created the `ContractNotFound` custom error class to make this catch nicer, instead of matching on the error message.

packages/cli/src/prompts/prompt.ts
```diff
@@ -136,45 +137,13 @@ export function proxiesList(
...
-
-    return inquirerQuestion(name, message, type, [...contractsFromBuild, ...flatten(packageContracts)]);
-  } else return [];
+export function contractsList(
+  name: string,
+  message: string,
+  type: string,
+  source?: choices.ContractsSource,
+): { [key: string]: any } {
+  return inquirerQuestion(name, message, type, choices.contracts(source));
```
frangio: I refactored out the part that actually builds the list of contracts to reuse it.

packages/cli/src/prompts/prompt.ts
```diff
@@ -202,8 +171,14 @@ export function methodsList(
     });
 }
 
-export function argLabel(arg: MethodArg): string {
-  return arg.name ? `${arg.name}: ${ABI.getArgTypeLabel(arg)}` : ABI.getArgTypeLabel(arg);
+export function argLabel(arg: MethodArg, index?: number): string {
```
frangio: Previously `arg.name` was being set to `#${index}` in some other part of the prompts module. Now this function is sometimes used for plain `MethodArg` extracted from the ABI without any preprocessing.

packages/cli/src/register-command.ts
```diff
@@ -0,0 +1,211 @@
...
+
+    yield [name, opt];
+  }
+
+  for (const arg of spec.args) {
+    yield [arg.name, arg];
+  }
+}
+
+async function promptForMissing(cmd: Command, spec: CommandSpec, argsAndOpts: ArgsAndOpts) {
```
frangio: I'm sorry for this huge function. The functionality is quite complex and I gave up on trying to organize it better. I'm sure there is a way to refactor it into something nicer. Suggestions will be appreciated.

packages/cli/src/register-command.ts
```diff
@@ -0,0 +1,220 @@
...
+          return e.callback();
+        } else {
+          throw e;
+        }
+      }
+      const [cmd, argsAndOpts] = getCommandArgsAndOpts(...actionArgs);
+      await promptForMissing(cmd, spec, argsAndOpts);
+      // TODO: normalize network name and include txParams (ConfigManager)
+      Telemetry.report(cmd.name(), argsAndOpts, !!argsAndOpts.interactive);
+      await action(...generateActionArgs(cmd, argsAndOpts));
```
frangio: Here `argsAndOpts` is an object with key-value pairs of all positional arguments and options, and `generateActionArgs` converts this back into format that commander passes arguments to an action. However, I would consider skipping this conversion and writing our actions to receive a single `argsAndOpts` argument.

packages/cli/src/prompts/prompt.ts
frangio: I had to undo this because it was being used in a place that I hadn't foreseen and it caused a test to fail.

packages/cli/src/commands/deploy/spec.ts
```diff
@@ -0,0 +1,89 @@
...
+    async prompt(): Promise<Question> {
+      const choices = await import('../../prompts/choices');
+
+      // TODO: Make this include contracts from _unlinked_ dependencies.
+      const contracts = choices.contracts('all');
+
+      return {
+        message: 'Pick a contract to deploy',
+        choices: contracts,
+        validate: value => contracts.includes(value),
```
spalladino: Shouldn't we return a string with the error message, or have a `validationError` property?

packages/cli/src/commands/deploy/spec.ts
```diff
@@ -0,0 +1,89 @@
...
+      const { fromContractFullName } = await import('../../utils/naming');
+      const { default: ContractManager } = await import('../../models/local/ContractManager');
+      const { argLabelWithIndex } = await import('../../prompts/prompt');
+      const { parseArg, getSampleInput } = await import('../../utils/input');
+
+      const contractName = argsAndOpts.contract as string;
+
+      const { package: packageName, contract: contractAlias } = fromContractFullName(contractName);
+      const contract = new ContractManager().getContractClass(packageName, contractAlias);
+      const constructorInputs: MethodArg[] = contract.schema.abi.find(f => f.type === 'constructor')?.inputs ?? [];
```
spalladino: Same here as above, I'd refactor this line out. I'd also consider a helper function that returns the contract object directly from the `fullContractName`. And I'd love to turn the `ContractManager` into a set of standalone functions - why everything needs to be an object?!

packages/cli/src/commands/deploy/spec.ts
```diff
@@ -0,0 +1,89 @@
...
+          message: `${argLabelWithIndex(arg, index)}:`,
+          validate: (value: string) => {
+            try {
+              parseArg(value, arg);
+              return true;
+            } catch (err) {
+              return false;
+            }
+          },
+          validationError,
```
spalladino: I'm curious: why a static property instead of a message generated within the `validate` function, so it can reuse `value` and any calculations done by the validation?

packages/cli/src/commands/deploy/spec.ts
```diff
@@ -0,0 +1,89 @@
...
+  {
+    format: '-n, --network <network>',
+    description: 'network to use',
+    async prompt(): Promise<Question> {
+      const { default: ConfigManager } = await import('../../models/config/ConfigManager');
+      const networks = ConfigManager.getNetworkNamesFromConfig();
+      return {
+        message: 'Pick a network',
+        choices: networks,
+        validate: value => networks.includes(value),
```
spalladino: Do we want this be case insensitive, when the user inputs the network from the command line? Just wondering.

packages/cli/src/models/files/ProjectFile.ts
```diff
@@ -215,6 +215,10 @@ export default class ProjectFile {
       manager === 'trufle' ? { manager: 'truffle' } : pickBy({ ...this.data.compiler, ...configOptions });
   }
 
+  public translateMaybeAlias(nameOrAlias: string): string {
```
spalladino: While accurate, I'm not sure this name is immediately clear. How about `contractOrAlias`?

packages/cli/src/models/network/NetworkController.ts
```diff
@@ -239,8 +249,15 @@ export default class NetworkController {
...
     Loggy.spin(__filename, '_uploadSolidityLib', `upload-solidity-lib${libName}`, `Uploading ${libName} library`);
-    const libInstance = await this.project.setImplementation(libClass, libName);
-    this.networkFile.addSolidityLib(libName, libInstance);
+
+    if (this.project !== undefined) {
+      const libInstance = await this.project.setImplementation(libClass, libName);
+      this.networkFile.addSolidityLib(libName, libInstance);
+    } else {
+      const libInstance = await Transactions.deployContract(libClass);
+      this.networkFile.addSolidityLib(libName, libInstance);
```
spalladino: Given the second line is the same on both branches, I'd move it outside the conditional

packages/cli/src/models/network/NetworkController.ts
```diff
@@ -239,8 +249,15 @@ export default class NetworkController {
...
     await this._setSolidityLibs(libClass); // Libraries may depend on other libraries themselves
     Loggy.spin(__filename, '_uploadSolidityLib', `upload-solidity-lib${libName}`, `Uploading ${libName} library`);
-    const libInstance = await this.project.setImplementation(libClass, libName);
-    this.networkFile.addSolidityLib(libName, libInstance);
+
+    if (this.project !== undefined) {
+      const libInstance = await this.project.setImplementation(libClass, libName);
+      this.networkFile.addSolidityLib(libName, libInstance);
+    } else {
+      const libInstance = await Transactions.deployContract(libClass);
```
spalladino: Can you add a comment explaining _when_ there is no project? Outside the context of this PR it's not clear.

packages/cli/src/models/network/NetworkController.ts
```diff
@@ -671,6 +690,35 @@ export default class NetworkController {
...
   }
 
+  public async createInstance(packageName: string, contractAlias: string, initArgs: unknown[]): Promise<Contract> {
+    await this.migrateManifestVersionIfNeeded();
+
+    if (!packageName) {
+      packageName = this.projectFile.name;
+    }
+
+    // TODO: Fix for contracts from dependencies that may use libraries.
```
spalladino: Can you create an issue for this?

packages/cli/src/models/network/NetworkController.ts
```diff
@@ -671,6 +690,35 @@ export default class NetworkController {
...
+    const contract = this.contractManager.getContractClass(packageName, contractAlias);
+    await this._setSolidityLibs(contract);
+
+    const instance = await Transactions.deployContract(contract, initArgs, this.txParams);
+
+    if (packageName === this.projectFile.name) {
+      await this._updateTruffleDeployedInformation(contractAlias, instance);
+    }
+
+    this.networkFile.addProxy(packageName, contractAlias, {
```
spalladino: WDYT about storing the contract bytecode hash as well? It may come in handy in the future if we want to tell the user whether a deployed instance matches the current source code or not.

packages/cli/src/utils/input.ts
```diff
@@ -134,11 +144,8 @@ function requireInputString(arg: string | string[]): arg is string {
  * unquoted strings in the input, or quotes using both simple and double quotes.
  * @param input string to parse
  * @returns parsed ouput.
- * The return type is a lie! This function returns a recursive type,
- * with arbitrarily nested string arrays, but ts has a hard time handling that,
- * so we're fooling it into thinking it's just one level deep.
```
spalladino: Nice!

packages/lib/src/errors.ts
```diff
@@ -0,0 +1,9 @@
+export class ContractNotFound extends Error {
```
spalladino: Nice!

packages/cli/test/commands/deploy.test.ts
```diff
@@ -0,0 +1,125 @@
...
+    projectFile = new ProjectFile('test/mocks/packages/package-empty.zos.json');
+    networkFile = new NetworkFile(projectFile, network);
+  });
+
+  it('should deploy a simple contract with no constructor', async function() {
+    const contract = 'SimpleNonUpgradeable';
+    await deploy(contract, [], {
+      network,
+      txParams,
+      networkFile: networkFile,
```
spalladino:
```suggestion
      networkFile,
```

spalladino: Repeat in each occurrence the file

packages/cli/test/commands/deploy.test.ts
```diff
@@ -0,0 +1,125 @@
...
+    });
+    const instances = networkFile.getProxies({ contract });
+    instances.should.have.lengthOf(1);
+
+    const instanceInfo = instances[0];
+    instanceInfo.kind.should.equal(ProxyType.NonProxy);
+    const instance = WithConstructorNonUpgradeable.at(instanceInfo.address);
+    (await instance.methods.answer().call()).should.equal('42');
```
spalladino: Sneaky, I did not anticipate that 42

packages/cli/src/commands/deploy/action.ts
```diff
@@ -0,0 +1,63 @@
...
+  contractName: string | undefined,
+  deployArgs: string[],
+  options: Options,
+): Promise<void> {
+  if (!options.skipCompile) {
+    await compile();
+  }
+
+  if (options.upgradeable === true) {
+    throw new AbortAction(async () => {
```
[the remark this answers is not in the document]
frangio: Yes, good catch. I don't see a way to make that happen without building it in in a very ad-hoc way. However, aborting early is only important now because we have to delegate to `create`. Once we add the transpiler it will be built in to `deploy`, and it won't be necessary to do this abort thing. In fact, I would remove the abort feature once we don't need it anymore.

packages/cli/src/commands/deploy/action.ts
```diff
@@ -0,0 +1,63 @@
...
+      // Translate arguments to syntax expected by create.
+      options['args'] = deployArgs.join(',');
+      await createAction(contractName, options);
+    });
+  }
+}
+
+export async function action(contractName: string, deployArgs: string[], options: Options): Promise<void> {
+  // this is an interactive prompt, so should be handled by the prompts
+  // if (!(await hasToMigrateProject(network))) process.exit(0);
```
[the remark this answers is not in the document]
frangio: It's in `commands/create`. 😅 

https://github.com/OpenZeppelin/openzeppelin-sdk/blob/a851d4ce1cc475e1b227c0decfe891b90b90a35a/packages/cli/src/commands/create.ts#L92

I think the way we should handle this is by adding a CLI flag `--no-migrate` that interactively results in this same prompt, and then we pass the boolean value to some other function. But the way in which this is happening in create is very indirect and it gave me a headache so I left it for later. I should probably implement it for this PR.

packages/cli/src/commands/deploy/action.ts
```diff
@@ -0,0 +1,63 @@
...
+
+  const { network, txParams } = options.txParams ? options : await ConfigManager.initNetworkConfiguration(options);
+
+  const { package: packageName, contract: contractAlias } = fromContractFullName(contractName);
+
+  const controller = new NetworkController(network, txParams, options.networkFile);
+
+  const contract = controller.contractManager.getContractClass(packageName, contractAlias);
+  const constructorInputs: MethodArgType[] = contract.schema.abi.find(f => f.type === 'constructor')?.inputs ?? [];
```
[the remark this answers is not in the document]
frangio: I agree that it's weird for `action` to have this concern, but I was reluctant to add anything to "`lib`", a.k.a. `upgrades`, and I don't think it's a good idea to use `upgrades` as the "utils" module for CLI. What do you think?

It's true that currently we are using `upgrades` in this way, so arguably I should just stick to that for now, but I would've liked to go in the direction of improving that situation.

packages/cli/src/commands/deploy/index.ts
```diff
@@ -0,0 +1,13 @@
...
+import { generateSignature, register as _register } from '../../register-command';
+
+import * as spec from './spec';
+
+export { name, description } from './spec';
+
+export const signature = generateSignature(spec.name, spec.args);
+
+export function register(program: Command) {
+  _register(program, spec, () => import('./action'));
```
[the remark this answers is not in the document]
frangio: Register is always called, but its action is only called when the command is invoked on the CLI. I used lazy imports so that command-specific code (prompts and actions) is only loaded if needed. I expect this to improve the loading times of the CLI, once everything is loaded lazily. I haven't measured it though and we can discuss whether it's worth the effort.

https://github.com/OpenZeppelin/openzeppelin-sdk/blob/a851d4ce1cc475e1b227c0decfe891b90b90a35a/packages/cli/src/register-command.ts#L53-L54

packages/cli/src/commands/deploy/spec.ts
frangio: There is a default validation error message based on the parameter name, but it can probably be improved.

https://github.com/OpenZeppelin/openzeppelin-sdk/blob/a851d4ce1cc475e1b227c0decfe891b90b90a35a/packages/cli/src/register-command.ts#L167

packages/cli/src/commands/deploy/spec.ts
frangio: Yes, this validation function needs to be improved, it's not replicating the behavior of `create`.

packages/cli/src/commands/deploy/action.ts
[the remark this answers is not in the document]
spalladino: I'd add this to `upgrades`, so when we split out the `utils` modules, this gets moved there automatically. Otherwise, when we make the split, this code will still be left there.

And yes, we _will_ be doing that code split; probably right after transpiler gets integrated (or even right before!).

packages/cli/src/commands/deploy/index.ts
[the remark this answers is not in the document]
spalladino: Makes sense, thanks!

packages/cli/src/register-command.ts
```diff
@@ -0,0 +1,220 @@
...
+      }
+    }
+  }
+}
+
+async function askQuestion(name: string, question: Question): Promise<string> {
+  const { message, choices, validate: rawValidate, validationError } = question;
+  const type = choices === undefined ? 'input' : 'list';
+
+  const validate = (value: string) => {
```
spalladino: Looping back to a comment above: I think `rawValidate` should definitely be optional, and here we should set default validations based on the `type` of the question. We can easily set up a default inclusion validation for `choices`, or a check for boolean-ish values for yes/no questions.

packages/cli/src/register-command.ts
```diff
@@ -0,0 +1,220 @@
...
+    type,
+    message,
+    choices,
+    validate,
+  });
+
+  return answers.question;
+}
+
+export type ArgsAndOpts = Record<string, boolean | string | string[]>;
```
spalladino: Can you push this definition closer to the beginning of the file, since it's used in several places before?

packages/cli/src/register-command.ts
spalladino: I'd go the easy way and just break it on the main conditionals inside the loop: sth like promptForUnset, promptForUnsetVariadic, validateArgOpt, validateVariadic.

packages/cli/src/register-command.ts
```diff
@@ -0,0 +1,220 @@
...
+// Converts the arguments that Commander passes to an action into an object
+// where the key-value pairs correspond to positional arguments and options,
+// and extracts the Command object.
+function getCommandArgsAndOpts(...args: unknown[]): [Command, ArgsAndOpts] {
+  const cmd = args.pop() as Command;
+
+  const argsAndOpts = {};
+
+  for (let i = 0; i < cmd._args.length; i++) {
+    argsAndOpts[cmd._args[i].name] = args[i];
```
spalladino: I'm surprised that Commander doesn't provide this out of the box

packages/cli/src/register-command.ts
```diff
@@ -0,0 +1,220 @@
...
+        continue;
+      }
+      if (param.variadic === true) {
+        if (!Array.isArray(value)) {
+          throw new Error(`Expected multiple values for ${name}`);
+        }
+        const prompt = await param.prompt?.(argsAndOpts);
+        if (prompt) {
+          if (prompt.length !== value.length) {
+            throw new Error(`Expected ${prompt.length} values for ${name} but got ${value.length}`);
```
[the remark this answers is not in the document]
frangio: It was half-intentional. It was a consequence of the design but I decided to keep it.

packages/cli/src/prompts/choices.ts
```diff
@@ -0,0 +1,61 @@
...
+
+import difference from 'lodash.difference';
+import flatten from 'lodash.flatten';
+import ProjectFile from '../models/files/ProjectFile';
+import ContractManager from '../models/local/ContractManager';
+import Dependency from '../models/dependency/Dependency';
+
+export type Choice = string | { name: string; value: string } | { type: 'separator' };
+
+export type ContractsSource = 'built' | 'notAdded' | 'added' | 'all';
```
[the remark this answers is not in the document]
frangio: My impression is that string literals are idiomatic. I couldn't find a source for this, but it's quite clear that the type system was designed around this idiom and it supports it really well. Both enums and symbols force you to import the value before using it, adding ceremony that isn't necessary and is prone to be worked around and cause errors. For example, if we use an enum it wouldn't be possible to use a string literal in its place as a function argument, because the string constants aren't subtypes of the enum, and it may lead a developer to write `'all' as ContractSource` which sneakily loses type safety since you can also do `'invalid' as ContractSource`.

Symbols are interesting but only add value when strong encapsulation is needed, since they allow to restrict access to a property to the module that created the symbol.

packages/cli/src/register-command.ts
frangio: With some work we can make it check the conversion, but it would approach black magic. Definitely interesting though, TypeScript is super powerful!

Leaving that aside though, I'm not sure that using `argsAndOpts` has stronger type guarantees (unless we also put in that extra work). I just think it would be nicer.

packages/cli/src/commands/deploy/action.ts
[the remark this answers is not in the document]
frangio: You're right. I intuitively went for an exception because I was thinking it would be necessary to jump over several levels of function calls but it isn't.

packages/cli/src/models/network/NetworkController.ts
```diff
@@ -206,6 +212,16 @@ export default class NetworkController {
     return pipeline.reduce((xs, f) => f(xs), this.projectFile.contracts);
   }
 
+  private getLibsToDeploy(libNames: string[], onlyChanged = false): Contract[] {
+    return libNames
+      .map(libName => Contracts.getFromLocal(libName))
+      .filter(libClass => {
+        const hasSolidityLib = this.networkFile.hasSolidityLib(libClass.schema.contractName);
+        const hasChanged = this._hasSolidityLibChanged(libClass);
+        return !hasSolidityLib || !onlyChanged || hasChanged;
```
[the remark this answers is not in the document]
frangio: Hm, so it seems I forgot to set `onlyChanged = true` in the call to this function in `deployChangedSolidityLibs`, right?

packages/cli/src/prompts/choices.ts
```diff
@@ -0,0 +1,61 @@
...
+      contractsFromLocal.map(({ value }) => value),
+    );
+  }
+
+  if (source === 'added') {
+    return contractsFromLocal;
+  }
+
+  // generate a list of built contracts and package contracts
+  if (source === 'all') {
```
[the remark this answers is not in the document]
frangio: I totally avoid them because I never remember which languages have fallthrough semantics, and the possibility of forgetting to put in `break` makes me panic. We may be able to add a linter rule to force always using `break`...

In this case though I moved most of this code from `prompts.contractsList` and it was using `if` statements.

packages/cli/src/commands/deploy/action.ts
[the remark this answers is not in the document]
ylv-io: > I wouldn't say exceptions necessarily mean that something went wrong
I agree. It depends on language and code style, for example in Python developers throw exceptions all the time and use them as a way to return from a function instead of returning `null`.

packages/cli/src/models/network/NetworkController.ts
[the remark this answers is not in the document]
ylv-io: I think so but I couldn't find any specific tests around that call and I am not that knowledgeable around how Solidity libs are deployed based on a project type. What concerns me is that it seems current code will redeploy libs all the time yet tests are totally fine with that.

packages/cli/src/prompts/choices.ts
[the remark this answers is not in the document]
ylv-io: Thanks for clarifications. Just want to mention that [airbnb eslint config](https://eslint.org/docs/rules/no-fallthrough)  doesn't allow [fallthrough](https://eslint.org/docs/rules/no-fallthrough).

packages/cli/src/commands/deploy/spec.ts
frangio: Leaving this case sensitive since it seems to be the behavior of the rest of the CLI.

packages/cli/src/prompts/choices.ts
```diff
@@ -0,0 +1,61 @@
...
+
+  // get contracts from project.json file
+  if (source === 'notAdded') {
+    return difference(
+      contractsFromBuild,
+      contractsFromLocal.map(({ value }) => value),
+    );
+  }
+
+  if (source === 'added') {
```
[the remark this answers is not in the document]
frangio: OCD Driven Development

frangio:
I still haven't finished resolving all the review comments, but I'm signing off for today. Feel free to wait until I'm done with everything to re-review, or if you want you can also start taking a look now.

packages/cli/src/commands/deploy/spec.ts
frangio: I didn't want to build a more general solution without seeing the need for it, and this separation felt more intuitive to me initially. Let's wait and see if this is enough?

frangio: I've changed my mind on this.

frangio:
`deploy` is not sending the concealed `txParams` in the telemetry reports, because it wasn't easy to fit it into the new command framework, and I didn't think the effort was worth it since the tx params that end up being used can be different because of overrides at different points.

packages/cli/src/telemetry/index.ts
```diff
@@ -43,11 +43,11 @@ export default {
...
     if (telemetryOptions === undefined || !telemetryOptions.optIn) return;
 
-    // extract network name if present
-    let network;
-    if ('network' in options) {
-      network = options.network;
-      if (network.match(/dev-/)) network = 'development';
+    // normalize network name
+    const { ZWeb3 } = await import('@openzeppelin/upgrades');
+    let network = await ZWeb3.getNetworkName();
```
spalladino: This normalization was being handled in ConfigManager#initNetworkConfiguration, which means that all calls after the initNetworkConfiguration got the normalized network name. If we need to do this at the telemetry level, this _might_ mean that we are failing to normalize at the command level. Can you double check?

packages/cli/src/models/network/NetworkController.ts
```diff
@@ -250,14 +250,17 @@ export default class NetworkController {
     await this._setSolidityLibs(libClass); // Libraries may depend on other libraries themselves
     Loggy.spin(__filename, '_uploadSolidityLib', `upload-solidity-lib${libName}`, `Uploading ${libName} library`);
 
-    if (this.project !== undefined) {
-      const libInstance = await this.project.setImplementation(libClass, libName);
-      this.networkFile.addSolidityLib(libName, libInstance);
+    let libInstance;
+
+    if (this.project === undefined) {
```
spalladino: Why not a ternary operator, to avoid the use of `let`?

```ts
const libInstance = (this.project === undefined) ? await Transactions.deployContract(libClass) : await this.project.setImplementation(libClass, libName);
```

packages/cli/src/register-command.ts
```diff
@@ -109,16 +109,16 @@ async function promptForMissing(cmd: Command, spec: CommandSpec, params: object)
     if (value === undefined || (Array.isArray(value) && value.length === 0)) {
       // Seems redundant but it helps the type system.
       if (param.variadic === true) {
-        const prompt = await param.prompt?.(params);
+        const details = await param.details?.(params);
         const values = [];
-        for (const p of prompt) {
-          values.push(await askQuestion(name, p));
+        for (const d of details) {
```
spalladino: `p -> d` - nothing gets past you, eh?

packages/cli/src/telemetry/index.ts
```diff
@@ -111,18 +111,22 @@ async function checkOptIn(interactive: boolean): Promise<GlobalTelemetryOptions
...
-    cliVersion: require('../../package.json').version,
+    cliVersion: await getCLIVersion(),
     upgradesVersion: getDependencyVersion('@openzeppelin/upgrades'),
     truffleVersion: getDependencyVersion('truffle'),
     web3Version: getDependencyVersion('web3'),
   };
 }
 
+async function getCLIVersion(): Promise<string> {
+  return JSON.parse(await fs.readFile(__dirname + '/../../package.json', 'utf8')).version;
```
spalladino: Why not [readJson](https://github.com/jprichardson/node-fs-extra/blob/HEAD/docs/readJson.md) from fs-extra?

packages/cli/src/register-command.ts
```diff
@@ -110,12 +111,18 @@ async function promptForMissing(cmd: Command, spec: CommandSpec, params: object)
         const details = await param.details?.(params);
         const values = [];
         for (const d of details) {
+          if (!params.interactive || DISABLE_INTERACTIVITY) {
+            throw new Error(`Missing required parameters ${name}`);
```
spalladino: I'd prefer if we collect all missing parameters and throw a single validation error - it's less frustrating to the user, as they know all at once every parameter they need to add

packages/cli/src/commands/deploy/spec.ts
```diff
@@ -4,6 +4,9 @@ import { TxParams } from '@openzeppelin/upgrades';
 import NetworkFile from '../../models/files/NetworkFile';
 import { DEFAULT_TX_TIMEOUT } from '../../models/network/defaults';
 
+const kinds = ['regular', 'upgradeable'] as const;
```
spalladino: Let's add minimal here as well, and forward it to `create` if set

packages/cli/src/register-command.ts
```diff
@@ -16,7 +16,7 @@ interface CommonParams {
 export interface ParamDetails {
   prompt?: string;
   promptType?: 'confirm' | 'list' | 'input';
-  choices?: Choice[];
+  choices?: readonly Choice[];
```
spalladino: BRAIN EXPLODES

packages/cli/src/register-command.ts
frangio: I like this idea but I'm going to implement it in a different PR.

packages/cli/src/register-command.ts
frangio: Hehe. I had to add this btw because `const kinds = ['regular', 'upgradeable', 'minimal'] as const;` creates a `readonly` array, because of the `as const`, and if I don't make it so the type will be `string[]`.

packages/cli/src/models/network/NetworkController.ts
frangio: I don't know, I was influenced by a recent post by Dan Abramov to not be afraid of `let`.

https://overreacted.io/on-let-vs-const

packages/cli/src/telemetry/index.ts
frangio: It got past me!

packages/cli/src/register-command.ts
frangio: https://github.com/OpenZeppelin/openzeppelin-sdk/issues/1398

packages/cli/src/models/network/NetworkController.ts
frangio: Changed to ternary operator.

PR #1382, opened 2020-01-13
Stop interactive create or update on validation errors

### LINKED ISSUE #1348: Command `create` bypasses contract validations
Contracts are validated to ensure they are upgradeable (eg by checking they have no ctors) at `push` time. If they fail the validation, the push command fails. However, if the contract is never `push`ed manually, but is handled instead automatically during a `create`, then the warnings will be displayed **but the contract will be pushed and created anyway**.

We should halt the `create` process if a validation error is thrown.

spalladino:
The force flag was being passed to push from both create and upgrade to force pushing the implementation contracts. This caused the contracts to be pushed even if they had validation errors. This commit removes that flag so the create or upgrade fails if there are any errors.

Fixes #1348

### DIFF
--- a/packages/cli/src/commands/push.ts
+++ b/packages/cli/src/commands/push.ts
@@ -6,12 +6,9 @@ import add from './add';
 import push from '../scripts/push';
 import Session from '../models/network/Session';
 import { compile } from '../models/compiler/Compiler';
-import { fromContractFullName } from '../utils/naming';
 import Dependency from '../models/dependency/Dependency';
-import ProjectFile from '../models/files/ProjectFile';
 import ConfigManager from '../models/config/ConfigManager';
 import { promptIfNeeded, networksList, InquirerQuestions } from '../prompts/prompt';
-import NetworkFile from '../models/files/NetworkFile';
 import Telemetry from '../telemetry';
 
 const name = 'push';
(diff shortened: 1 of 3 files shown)


packages/cli/src/commands/push.ts
```diff
@@ -94,20 +91,9 @@ async function runActionIfRequested(externalOptions: any): Promise<void> {
...
-    if (
-      force ||
-      (!packageName && !networkFile.hasContract(contractAlias)) ||
-      (packageName && !networkFile.hasDependency(packageName))
-    ) {
-      await action({ ...options, dontExitProcess: true, skipTelemetry: true });
-    }
-  }
+  const { interactive } = options;
+  if (!interactive) return;
```
frangio: I personally prefer 1) avoiding early returns, and 2) always using braces. The latter should be a linter rule if we even agree to it, so I won't insist on it.

spalladino: 1) I like returns either at the beginning or the end, though not in the middle. Returns at the very first lines of the function feel natural as they are kind of "validations" of whether we want to run or not. Agree on avoiding them in the middle.

2) I particularly line single-line ifs when it's a return. Seems like we're on opposite sides on this one!

spalladino: Actually, just changed this so the return is on the very first line.

PR #1384, opened 2020-01-13
Store relative paths in project file

### LINKED ISSUE #1372: Absolute path to artifacts and sources in project file
When using the current `master` (not in 2.6.0), running `oz compile` in a new project with a single contract hardcodes the absolute path to artifacts and sources in the project json file. 

```json
{
  "manifestVersion": "2.2",
  "contracts": {},
  "dependencies": {},
  "name": "foo",
  "version": "1.0.0",
  "compiler": {
    "compilerSettings": {
      "optimizer": {
        "enabled": false,
        "runs": "200"
      }
    },
    "typechain": {
      "enabled": false
    },
    "manager": "openzeppelin",
    "solcVersion": "0.5.15",
    "artifactsDir": "/home/spalladino/Projects/sdk/examples/node-12/build/contracts",
    "contractsDir": "/home/spalladino/Projects/sdk/examples/node-12/contracts"
  },
  "telemetryOptIn": false
}
```

Since this file is to be added to version control and shared among all devs in the project, the path should always be relative to the project root, and not absolute.

spalladino:
Fixes #1372

### DIFF
--- a/packages/cli/src/models/files/ProjectFile.ts
+++ b/packages/cli/src/models/files/ProjectFile.ts
@@ -49,7 +49,7 @@ function getProjectPath(file: string, dir: string): [string, string] {
 
 function getRootFromFilePath(file: string): string {
   const dir = path.dirname(file);
-  return dir === OPEN_ZEPPELIN_FOLDER ? path.dirname(dir) : dir;
+  return path.basename(dir) === OPEN_ZEPPELIN_FOLDER ? path.dirname(dir) : dir;
 }
 
 function getFileFromRoot(dir: string): string {
--- a/packages/cli/src/models/files/ProjectFile.ts
+++ b/packages/cli/src/models/files/ProjectFile.ts
@@ -48,6 +48,7 @@ export const LEGACY_PROJECT_FILE_NAME = 'zos.json';
 
 export default class ProjectFile {
   public filePath: string;
+  public root: string;
   public data: ProjectFileData;
 
   public static getLinkedDependencies(filePath: string = null): string[] {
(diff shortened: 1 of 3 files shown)


packages/cli/src/models/files/ProjectFile.ts
```diff
@@ -84,7 +84,8 @@ export default class ProjectFile {
   }
 
   public get root(): string {
-    return path.dirname(this.filePath);
+    // We go to the parent of the containing directory (.openzeppelin) to get the root
+    return path.dirname(path.dirname(this.filePath));
```
frangio: I don't think this is correct as long as we support the old `zos.json`. Or is migration mandatory?

spalladino: Good catch! Just force pushed a fix. By the way, only after I pushed it I noted I shouldn't have amended, so you could see the new diff. Apologies for that! Still, being a small change, I hope you won't mind!

packages/cli/src/models/files/ProjectFile.ts
```diff
@@ -42,12 +41,29 @@ interface ProjectFileData {
...
 
+function getProjectPath(file: string, dir: string): [string, string] {
+  if (!dir) dir = file ? getRootFromFilePath(file) : process.cwd();
+  if (!file) file = getFileFromRoot(dir);
+  return [file, dir];
+}
+
+function getRootFromFilePath(file: string): string {
+  const dir = path.dirname(file);
+  return dir === OPEN_ZEPPELIN_FOLDER ? path.dirname(dir) : dir;
```
frangio: If I'm not mistaken `file` could be an absolute path.

```suggestion
  return path.basename(dir) === OPEN_ZEPPELIN_FOLDER ? path.dirname(dir) : dir;
```

PR #1385, opened 2020-01-13
Validate constructors in ancestor contracts

### LINKED ISSUE #1332: SDK fails to identify inherited constructors with no arguments
Consider the following Solidity source file:

```solidity
pragma solidity ^0.5.0;

contract Deployer {
    address public deployer;

    constructor() public {
        deployer = msg.sender;
    }
}

contract Foo is Deployer { }
```

Due to inheritance, `Foo`'s default constructor must call `Deployer`'s constructor. Because `Deployer`'s receives no arguments, `Foo` doesn't need to make its constructor explicit: the default constructor will suffice. However, **the SDK doesn't realize that Foo has a non-trivial constructor and raises no warnings**.

When running `oz create Deployer`, the SDK correctly identifies `Deployer` as a contract with a constructor, and emits a warning. 

```
$ npx oz create Deployer
✓ Added contract Deployer
- Contract Deployer has an explicit constructor. Change it to an initializer function. See https://docs.openzeppelin.com/sdk/2.5/writing-contracts#initializers.
...
```

However, no such warning exists for `Foo`:

```bash
$ npx oz create Foo
✓ Added contract Foo
✓ Contract Foo deployed
All contracts have been deployed
✓ Instance created at 0x4bf749ec68270027C5910220CEAB30Cc284c7BA2
```

We can easily check that `Foo` is indeed inheriting from `Deployer`:

```bash
$ npx oz call
? Pick an instance Foo at 0x4bf749ec68270027C5910220CEAB30Cc284c7BA2
? Select which function deployer()
✓ Method 'deployer()' returned: 0x0000000000000000000000000000000000000000
```
----

Multiple useful contracts have an argumentless constructor and are used via inheritance: see [Ownable](https://docs.openzeppelin.com/contracts/2.x/api/ownership#Ownable) and [ReentrancyGuard](https://docs.openzeppelin.com/contracts/2.x/api/utils#ReentrancyGuard).

spalladino:
Fixes #1332

### DIFF
--- a/packages/lib/src/validations/Constructors.ts
+++ b/packages/lib/src/validations/Constructors.ts
@@ -1,19 +1,17 @@
 import Contract from '../artifacts/Contract';
 import { BuildArtifacts } from '..';
-import ContractAST, { ContractDefinitionFilter } from '../utils/ContractAST';
-import { Artifact } from '../artifacts/BuildArtifacts';
+import ContractAST, { FunctionDefinitionFilter } from '../utils/ContractAST';
 
 export function hasConstructor(contract: Contract, buildArtifacts: BuildArtifacts): boolean {
-  if (hasConstructorInABI(contract.schema)) return true;
-
-  const baseContracts = new ContractAST(contract, buildArtifacts, ContractDefinitionFilter)
+  return new ContractAST(contract, buildArtifacts, FunctionDefinitionFilter)
     .getLinearizedBaseContracts()
-    .map(node => buildArtifacts.getArtifactByName(node.name));
-
-  const baseContractsWithConstructors = baseContracts.filter(hasConstructorInABI);
-  return baseContractsWithConstructors.length > 0;
+    .filter(hasNonEmptyConstructorInAST)
+    .length > 0;
 }
 
-function hasConstructorInABI(contract: Artifact): boolean {
-  return !!contract.abi.find(fn => fn.type === 'constructor');
+function hasNonEmptyConstructorInAST(contractNode: any): boolean {
+  return contractNode.nodes
+    .filter((n: any) => n.nodeType === 'FunctionDefinition' && n.kind === 'constructor')
+    .filter((n: any) => n.body.statements.length > 0)
+    .length > 0;
 }
--- a/packages/lib/src/validations/Constructors.ts
+++ b/packages/lib/src/validations/Constructors.ts
@@ -12,6 +12,6 @@ export function hasConstructor(contract: Contract, buildArtifacts: BuildArtifact
 function hasNonEmptyConstructorInAST(contractNode: any): boolean {
   return contractNode.nodes
     .filter((n: any) => n.nodeType === 'FunctionDefinition' && n.kind === 'constructor')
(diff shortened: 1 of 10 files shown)


spalladino:
Funny thing. The `GSNRecipient` from contracts-ethereum-package extends from [`Context`](https://github.com/OpenZeppelin/openzeppelin-contracts-ethereum-package/blob/master/contracts/GSN/Context.sol), which has an internal constructor. 

And apparently Solidity generates an ABI entry for the constructor (even if it's internal), which is being picked up by the new validation, and causes the integration test to fail.
```
$ jq .abi build/contracts/Context.json 
[{
    "inputs": [],
    "payable": false,
    "stateMutability": "nonpayable",
    "type": "constructor"
}]
```

We'll have to think how to get around this. Perhaps adding an exception for the Context contract? Or checking if the ctor body is empty?

packages/lib/test/src/validations/Validations.test.js
```diff
@@ -6,49 +6,67 @@ import Contracts from '../../../src/artifacts/Contracts';
 import { setVanillaContractsPackageName } from '../../../src/validations/VanillaContracts';
 
 describe('Validations', function() {
-  it('should warn when adding a contract with a constructor', async function() {
-    validate('WithConstructor').hasConstructor.should.be.true;
+  describe.only('#hasConstructor', function () {
```
frangio:
```suggestion
  describe('#hasConstructor', function () {
```

packages/lib/src/validations/Constructors.ts
```diff
@@ -1,4 +1,17 @@
...
+import ContractAST, { FunctionDefinitionFilter } from '../utils/ContractAST';
+
+export function hasConstructor(contract: Contract, buildArtifacts: BuildArtifacts): boolean {
+  return new ContractAST(contract, buildArtifacts, FunctionDefinitionFilter)
+    .getLinearizedBaseContracts()
+    .filter(hasNonEmptyConstructorInAST)
+    .length > 0;
+}
+
+function hasNonEmptyConstructorInAST(contractNode: any): boolean {
```
frangio: We should also check that there are no modifiers!

spalladino: Good catch! Added, please re-review!

packages/lib/src/validations/Constructors.ts
```diff
@@ -12,6 +12,6 @@ export function hasConstructor(contract: Contract, buildArtifacts: BuildArtifact
 function hasNonEmptyConstructorInAST(contractNode: any): boolean {
   return contractNode.nodes
     .filter((n: any) => n.nodeType === 'FunctionDefinition' && n.kind === 'constructor')
-    .filter((n: any) => n.body.statements.length > 0)
+    .filter((n: any) => n.body.statements.length > 0 || n.modifiers.length > 0)
```
frangio: There can be false positives if one of the "modifiers" is a super call to a parent contract with an empty constructor. But that's an edge case that I don't think will happen.

PR #1386, opened 2020-01-14
Add url and path properties to network config

### LINKED ISSUE #1338: Provider URL is misconstructed when the host string includes a path
`NetworkConfig` setups the provider URL in the following way:

```javascript
provider = `${protocol ? protocol : 'http'}://${host}:${port}`;
```

This is incorrect when `host` contains a path, since the port should be added at the end of the domain, not the entire string.

For example, given `host = my.domain.com/my/endpoint` and `port = 80`, the provider url should be `my.domain.com:80/my/endpoint` and not `my.domain.com/my/endpoint:80`. 

In particular this means that Infura hosts cannot be used, since they look like `mainnet.infura.io/v3/apiKey`.

As a side note, we may want to consider not making the `port` parameter mandatory, since web3 is able to deduce it for `http` and `https`.

spalladino:
Add support for specifying an url entry with the full URL to the node. Also adds a new entry, path, that can be combined with host, protocol, and port, and makes port optional. Note that #1338 suggested accepting a path as part of the host, though this is not technically correct (hence the new path property).

Fixes #1338

### DIFF
--- a/packages/cli/src/models/config/NetworkConfig.ts
+++ b/packages/cli/src/models/config/NetworkConfig.ts
@@ -25,15 +25,26 @@ interface NetworkSnakeCase<T> {
 
 type NetworkId<T> = NetworkCamelCase<T> | NetworkSnakeCase<T> | (NetworkCamelCase<T> & NetworkSnakeCase<T>);
 
-type Network = {
+type NetworkURL = {
+  url: string;
+};
+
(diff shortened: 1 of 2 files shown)


nventuro:
With this change we can now use the CLI on public networks by connecting to Infura with no mnemonics (read-only) :confetti_ball:

PR #1387, opened 2020-01-16
Update with CLI docsite URL

abcoathup:


spalladino:
Created https://github.com/OpenZeppelin/openzeppelin-sdk/issues/1389 to tackle it

PR #1394, opened 2020-01-17
Catch empty networks in network configuration files

### LINKED ISSUE #1339: Malformed networks files are not detected
The SDK fails to properly detect scenarios where the object exported by the network config file doesn't have a `networks` key.

To reproduce, replace the 'networks' key with any other string.

If the config file is `networks.js`, the following error message is displayed:
>You must provide a `choices` parameter

If instead it is a `truffle.js` file, this other message is displayed:
>A network name must be provided to execute the requested action.

spalladino:
Fixes #1339

### DIFF
--- a/packages/cli/src/models/config/ConfigManager.ts
+++ b/packages/cli/src/models/config/ConfigManager.ts
@@ -78,6 +78,11 @@ const ConfigManager = {
     return config && config.networks ? Object.keys(config.networks) : undefined;
   },
 
+  getConfigFileName(root: string = process.cwd()): string {
+    this.setBaseConfig(root);
+    return this.config.getConfigFileName(root);
+  },
+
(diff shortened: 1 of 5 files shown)


packages/cli/src/models/config/ConfigManager.ts
```diff
@@ -78,6 +78,11 @@ const ConfigManager = {
     return config && config.networks ? Object.keys(config.networks) : undefined;
   },
 
+  getConfigFileName(root: string = process.cwd()): string {
```
nventuro: It seems like this new method is only used for the nice error message? I'd expect for this PR to also use it when reading from the network file.

spalladino: @nventuro fixed in `NetworkConfig`! I did not change it in TruffleConfig though, but because it was not a direct match.

PR #1408, opened 2020-01-22
Fix regular deploy issues with ZWeb3 initialization

frangio:
The recent #1393 introduced some changes to `ZWeb3` initialization that made evident an issue with the deploy command: `ZWeb3` was being implicitly accessed too early by using methods in `ContractManager`. Too early means before the it had been initialized properly.

To solve this I added an optional post-processing step to the spec for every parameter, so once we have gathered all the required parameters, we use this step to initialize the config. Now that this is initialized at this time, I've also reverted the changes I had had to introduce to telemetry (to work around some of the config information not being available at the time the telemetry report was produced).

### DIFF
--- a/packages/cli/src/models/config/ConfigManager.ts
+++ b/packages/cli/src/models/config/ConfigManager.ts
@@ -9,7 +9,6 @@ import isNil from 'lodash.isnil';
 
 const ConfigManager = {
   config: undefined,
-  cache: undefined,
 
   initialize(root: string = process.cwd()): void {
     if (!TruffleConfig.exists() && !NetworkConfig.exists()) {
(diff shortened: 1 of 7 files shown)


packages/cli/test/telemetry.test.js
```diff
@@ -128,35 +127,29 @@ describe('telemetry', function() {
                 stringField: 'foo',
                 numberField: 3,
                 objectField: { stringField: 'foo', arrayField: [1, 2, 3] },
+                network: 'development',
               };
             });
 
             context('when a non-development network is specified', function() {
-              it('conceals all options recursively except for the command name', async function() {
+              it('conceals all options recursively except for the network name', async function() {
```
spalladino:
```suggestion
              it('conceals all options recursively except for the command and network name', async function() {
```

packages/cli/src/models/config/ConfigManager.ts
```diff
@@ -9,7 +9,6 @@ import isNil from 'lodash.isnil';
 
 const ConfigManager = {
   config: undefined,
-  cache: undefined,
```
spalladino: Where was this cache being used? I cannot find where it was read, only where it was set.

frangio: It was used for testing telemetry only, as a way of stubbing the thing. Now that I think of it it would've been better to just stub the init function using sinon. My bad!

PR #1412, opened 2020-01-27
Add @types for lodash packages in CLI; Fix errors

ylv-io:
A lot of Lodash functions doesn't have typing which leads to loss of types. This PR adds typings for Lodash functions and fix errors caused by missing types.


ylv-io:
`lodash.pickBy` is particularly bad because it completely wipes out all they type information in the way we use. Its usage is also questionable. As far as I understand we use to remove `falsy` values from objects but I am not sure what is the purpose of this. Also it is error prone in the way that any falsy value will be removed which devs up and down callstack may not expect.

spalladino:
> Its usage is also questionable. As far as I understand we use to remove falsy values from objects but I am not sure what is the purpose of this. Also it is error prone in the way that any falsy value will be removed which devs up and down callstack may not expect.

You are not the only one to have spotted that: https://github.com/OpenZeppelin/openzeppelin-sdk/issues/1231

packages/cli/src/models/dependency/Dependency.ts
```diff
@@ -97,22 +97,18 @@ export default class Dependency {
...
 
     const libraries = fromPairs(
       await Promise.all(
         map(libraryNames, async libraryName => {
-          const implementation = await project.setImplementation(
-            Contracts.getFromNodeModules(this.name, libraryName),
-            libraryName,
-          );
-          return [libraryName, implementation.address];
+          if (typeof libraryName === 'string') {
```
spalladino: Can `libraryName` be something that is not a string? 
If yes, I'd handle it explicitly.
If not, I'd remove this `if` clause and just type it.

packages/cli/src/models/network/NetworkController.ts
```diff
@@ -190,25 +190,18 @@ export default class NetworkController {
...
-            this.hasContractChanged(contractAlias, contract) ||
-            this._hasChangedLibraries(contract, changedLibraries),
-        ),
-    ];
-
-    return pipeline.reduce((xs, f) => f(xs), this.projectFile.contracts);
+
+    return toPairs(this.projectFile.contracts)
+      .map(
+        ([contractAlias, contractName]) => [contractAlias, Contracts.getFromLocal(contractName)] as [string, Contract],
```
spalladino: I think this line is not the same as above. With `([contractAlias, contractName]): [string, Contract] => [ ... ]`, you are telling the compiler the expected return value, and it will check that the implementation will conform to it. By changing it to `... as [string, Contract]`, you are manually casting it to what you want it to be, and losing the compiler type check.

packages/cli/src/scripts/interfaces.ts
```diff
@@ -68,10 +68,10 @@ export type Params =
...
 export interface CreateParams extends Proxy {
-  salt?: string;
-  signature?: string;
-  admin?: string;
-  kind?: ProxyType;
+  salt: string;
+  signature: string;
+  admin: string;
+  kind: ProxyType;
 }
```
spalladino: This change reminded me of [this talk](https://www.youtube.com/watch?v=YR5WdGrpoug), which I coincidentally watched yesterday. TLDR: an object does not have optional fields, it has fields which may be optional depending on the context.

packages/lib/src/validations/index.ts
```diff
@@ -100,22 +100,6 @@ function validateStorage(
   };
 }
 
-function tryGetUninitializedBaseContracts(contract: Contract): string[] {
```
spalladino: Should we remove `packages/lib/src/validations/Initializers.ts` as well then, if we are removing this whole function?

packages/cli/src/models/dependency/Dependency.ts
ylv-io: To my understanding `libraryName` is always a string. Its type is lost somewhere up at the callstack. I didn't have time to fix that. By just typing you mean `libraryName as string`? I think keeping is better because it will prevent errors if `libraryName` type will change due to code modifications when `libraryName as string` will overwrite any type information.

packages/cli/src/models/network/NetworkController.ts
ylv-io: In fact we should tune linter to give warning on `as` usage and reduce them to zero eventually.

packages/cli/src/scripts/interfaces.ts
ylv-io: Particularly in TypeScript it makes no sense to have a type with all the fields optional because then `Partial<NonOptional>` gives you an optional version of it.

packages/lib/src/validations/index.ts
ylv-io: Certainly. SDK is hard enough to navigate even without dead code.

packages/cli/src/models/dependency/Dependency.ts
spalladino: I was thinking of typing `libraryNames` above, and also changing:
```diff
- map(libraryNames, async libraryName => {
+ map(libraryNames, async libraryName: string => {
```

PR #1413, opened 2020-01-28
Support solidity 0.6 in Initializable contract

spalladino:
Modifies Initializable so it is compatible with Solidity 0.5 and 0.6. Adds a set of tests that run on the 0.6 version to verify its behaviour (note that the 0.6 compiled contracts are committed into source control to simplify the setup).

I suggest reviewing commits separately. The 2nd is somewhat unrelated to this PR, but required for the tests to pass (it was either that, or manually killing the CircleCI cache).


packages/lib/test/mocks/mock-solc-0.6/contracts/initializable/MultipleInheritanceInitializableMocks.sol
```diff
@@ -0,0 +1,79 @@
...
+// 
+//         Human
+//       /       \
+//      |       Gramps
+//      |         |
+//    Mother    Father
+//      |         |
+//      -- Child --
+
+/**
```
ylv-io: Love the art and the comment.

packages/lib/test/mocks/mock-solc-0.6/contracts/initializable/MultipleInheritanceInitializableMocks.sol
```diff
@@ -0,0 +1,79 @@
...
+ * Child extends from mother, father (gramps)
+ */
+contract SampleChild is Initializable, SampleMother, SampleFather {
+  uint256 public child;
+
+  function initialize(uint256 value) initializer override(SampleGramps, SampleMother) public {
+    SampleGramps.initialize(value);
+    SampleMother.initialize(value);
+  }
```
ylv-io: Why do we need to override it here? Having two methods called `initialize` is confusing. It is like having two constructors.

packages/lib/test/mocks/mock-solc-0.6/contracts/initializable/MultipleInheritanceInitializableMocks.sol
```diff
@@ -0,0 +1,79 @@
...
+
+  function initialize(uint256 _mother, uint256 _gramps, uint256 _father, uint256 _child) initializer public {
+    SampleMother.initialize(_mother);
+    SampleFather.initialize(_gramps, _father);
+    child = _child;
+  }
+}
```
ylv-io: Again, couldn't run the code so my question is why this doesn't cause to a duplicated call at `SampleHuman. initialize`? Since both `SampleMother` and `SampleFather` are children of `SampleHuman`.

packages/lib/test/mocks/mock-solc-0.6/contracts/initializable/MultipleInheritanceInitializableMocks.sol
spalladino: Heh, it was there from before!

packages/lib/test/mocks/mock-solc-0.6/contracts/initializable/MultipleInheritanceInitializableMocks.sol
spalladino: We don't actually need it, but it's helpful for testing a nasty case

packages/lib/test/mocks/mock-solc-0.6/contracts/initializable/MultipleInheritanceInitializableMocks.sol
spalladino: Actually, it does, and it's one of the things tested here (note that this test existed from before, I just moved it and migrated it to 0.6). There was a bug in Initializable that caused it to fail in these scenarios with an _already initialized_ error. This test checks that it's ok to call a base initializer more than once during the initialization process, to accommodate for structures like this one.

spalladino:
> It seems symlink Initializable.sol is broken. Does it works for you?

Hmm it works for me, and also works on the CI. Do you know if OSX handles symlinks differently?

packages/lib/test/mocks/mock-solc-0.6/contracts/initializable/MultipleInheritanceInitializableMocks.sol
ylv-io: Hmm. Didn't know that our current approach is that your `init` method might be called twice. This is certainly tricky and not many devs would expect that. I like the direction we are moving with Transpiler.

ylv-io:
OSX should handle them fine but if it works on CI and your machine it is fine. I've got all the answer I wanted. Thanks!

PR #1415, opened 2020-01-28
Do not ask for typechain support on compile outside a project

### LINKED ISSUE #1414: Prompt for typechain support is always asked when out of a project
Command `oz compile` is set up so it can work outside an `.openzeppelin/project.json`. However, in these cases, if there is a `tsconfig.json`, the CLI will ask to _Enable typechain support_, and since there is no `project.json` to record the answer, will keep asking every single time `oz compile` is run.

Do not interactively ask for typechain support if there is no `project.json`.

spalladino:
This PR adds typechain options for the command line, and avoid interactive questions whenever the user runs compile outside an OZ project (fixes #1414). It also fixes a bug that caused typechain wrappers not to be generated when there are multiple contracts.


packages/cli/src/models/compiler/Compiler.ts
```diff
@@ -47,11 +48,22 @@ export async function compile(
...
   // Run typechain if requested
-  if (resolvedOptions.typechain && resolvedOptions.typechain.enabled) {
+  if (resolvedOptions.typechain?.enabled) {
+    Loggy.spin(__filename, 'compile', 'compile-typechain', 'Generating typechain artifacts...');
     const result = compileResult as ProjectCompileResult;
-    const filesList = result && result.artifacts ? result.artifacts.map(c => c.contractName).join(',') : '*';
-    const filesGlob = `${resolvedOptions.outputDir}/${filesList}.json`;
-    await typechain(filesGlob, resolvedOptions.typechain.outDir, resolvedOptions.typechain.target);
+    let filesGlob: string;
+    if (!result?.artifacts) {
```
ylv-io: The line `const { compileWithTruffle, compileWithSolc } = exports;` cause to lose all the type information and then `const result = compileResult as ProjectCompileResult;` wrongly resets it because real type of `compileResult` is `void | ProjectCompileResult` which leads to `if (!result?.artifacts) {` not checking for `result` being `undefined`.
It can be fixed by using the following statements:
```
const withTruffle: typeof compileWithTruffle = exports.compileWithTruffle;
const withSolc:  typeof compileWithSolc = exports.compileWithSolc;
```
instead of `const { compileWithTruffle, compileWithSolc } = exports;` which do maintain type information.
Also you cant use optional chaining operator on `void` and this `const compilePromise = useTruffle ? compileWithTruffle() : compileWithSolc(resolvedOptions);` doesn't make much sense given `compileWithTruffle` returns `void`. I suggest to change return value to `undefined` or not to capture it return value at all.

packages/cli/test/commands/compile.test.js
```diff
@@ -36,4 +37,22 @@ describe('compile command', function() {
       });
     },
   );
+
+  itShouldParse(
+    'should call compile with typechain',
+    'compiler',
+    'zos compile --typechain=web3-v1 --typechain-outdir=foo --no-interactive',
```
ylv-io:
```suggestion
    'oz compile --typechain=web3-v1 --typechain-outdir=foo --no-interactive',
```
You can actually type anything instead of `zos` for example `sdfsdf` and it still will work. I think we should improve that.

packages/cli/src/models/compiler/Compiler.ts
spalladino: Good comments, thanks!!

spalladino:
Ready to re-review @ylv-io!

PR #1425, opened 2020-02-05
Fix deploy bug for upgradeable and minimal deploys

### LINKED ISSUE #1417: `oz deploy` fails to deploy empty contract using `upgradeable` and `minimal` types.
Given a simple contract:
```
pragma solidity ^0.5.0;

contract Foo {
    function hello() public pure returns(string memory) {
        return "hello";
    }
}
```
in empty project, running `oz deploy` with either `upgradeable` or `minimal` would produce the following error `Given network 'dev-1580295554917' is not defined in your networks.js file`.
image

frangio:
Fixes #1417 

The surface problem was that we were replacing the `network` option in the parameters with the normalized network name, and when forwarding to the `create` command it failed because the normalized name was not one of the user's networks.

The deeper problem was that we were initializing and then re-initializing the ConfigManager, first in `deploy` and then in `create`, in these cases. I made the initialization happen only on `deploy`.

### DIFF
--- a/packages/cli/src/commands/deploy/spec.ts
+++ b/packages/cli/src/commands/deploy/spec.ts
@@ -45,7 +45,7 @@ export const args: Arg[] = [
     async details(params: Options & Args): Promise<ParamDetails[]> {
       // If the user requests an upgradeable deploy, we will internally call
       // the create command and let it handle its own argument parsing.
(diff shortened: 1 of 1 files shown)


ylv-io:
Eventually we need to write tests for new arguments/options parser module.

PR #1426, opened 2020-02-06
Push only required contract during create command

### LINKED ISSUE #1250: Unrelated contracts get deployed when using create
I have a simple project with two unrelated contracts: `Receiver` and `Empty`. After running `oz create` and choosing to deploy `Receiver`, `Empty` was also deployed:

```bash
$ npx oz create 
? Pick a contract to instantiate Receiver
? Pick a network ropsten
✓ Contract Receiver deployed
✓ Contract Empty deployed
All contracts have been deployed
```

I understand this may be related to an underlying `push` operation, but it is an unexpected outcome which may result in the user's funds being used in ways they didn't intend or agree to (there was no prompt asking whether or not I wanted to deploy the other contract on my project).

ylv-io:
Fixes #1250. Will push only request contract by `create` command. The same functionality theoretically can be implemented for `update` command as well but it will require quite a bit of rearchitecting due to `update` command accepting options like `all/alias/address` as well as these options being resolved down the call stack of the `update` command.

### DIFF
--- a/packages/cli/src/commands/push.ts
+++ b/packages/cli/src/commands/push.ts
@@ -3,6 +3,7 @@ import { ZWeb3 } from '@openzeppelin/upgrades';
 
 import add from './add';
 import push from '../scripts/push';
+import { PushParams } from '../scripts/interfaces';
 import Session from '../models/network/Session';
 import { compile } from '../models/compiler/Compiler';
 import Dependency from '../models/dependency/Dependency';
@@ -42,6 +43,7 @@ async function commandActions(options: any): Promise<void> {
 
 async function action(options: any): Promise<void> {
   const {
+    contractAlias,
     force,
     deployDependencies,
     reset: reupload,
@@ -66,7 +68,7 @@ async function action(options: any): Promise<void> {
   });
   const promptDeployDependencies = await promptForDeployDependencies(deployDependencies, network, interactive);
 
-  const pushArguments = {
+  const pushArguments: PushParams = {
     deployProxyAdmin,
     deployProxyFactory,
     force,
@@ -76,7 +78,10 @@ async function action(options: any): Promise<void> {
     ...promptDeployDependencies,
   };
 
-  if (!options.skipTelemetry) await Telemetry.report('push', pushArguments, interactive);
+  if (contractAlias) pushArguments.contractAliases = [contractAlias];
+
+  if (!options.skipTelemetry)
+    await Telemetry.report('push', (pushArguments as unknown) as Record<string, unknown>, interactive);
   await push(pushArguments);
   if (!options.dontExitProcess && process.env.NODE_ENV !== 'test') process.exit(0);
 }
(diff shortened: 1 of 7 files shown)


packages/cli/src/commands/push.ts
```diff
@@ -89,9 +94,9 @@ async function runActionIfRequested(externalOptions: any): Promise<void> {
   return action(options);
 }
 
-async function runActionIfNeeded(contractName: string, network: string, options: any): Promise<void> {
+async function runActionIfNeeded(contractAlias: string, network: string, options: any): Promise<void> {
```
spalladino: This `contractName` was not even used, right?

packages/cli/src/commands/push.ts
```diff
@@ -76,7 +78,10 @@ async function action(options: any): Promise<void> {
     ...promptDeployDependencies,
   };
 
-  if (!options.skipTelemetry) await Telemetry.report('push', pushArguments, interactive);
+  if (contractAlias) pushArguments.contractAliases = [contractAlias];
+
+  if (!options.skipTelemetry)
+    await Telemetry.report('push', (pushArguments as unknown) as Record<string, unknown>, interactive);
```
spalladino: I'm curious: why was this cast needed when changing from `{ [key: string]: unknown }` to `Record<string, unknown>` in the `report` method definition? I thought that both were semantically the same.

packages/cli/src/commands/push.ts
ylv-io: Correct.

packages/cli/src/commands/push.ts
ylv-io: The cast is needed not because of the change of `report` method but because of change `pushParams` to type `PushParams` which doesn't have required index signature.

packages/cli/src/models/network/NetworkController.ts
```diff
@@ -122,13 +122,13 @@ export default class NetworkController {
   }
 
   // DeployerController
-  public async push(reupload = false, force = false): Promise<void | never> {
+  public async push(aliases: string[], { reupload = false, force = false } = {}): Promise<void | never> {
```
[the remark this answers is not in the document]
ylv-io: Yes. I was thinking there are all the null checks. I completely forgot about that our `tsconfig` is not in a strict mode. We should bump it to it as fast as possible.

darshan0071990:
I have updated cli sdk & upgrades to version 2.7.1. I have 2 different contracts let say Contract A {} and Contract B {}. Both the contracts are added and part of the contracts object of the project.json file. When I run oz create in interactive mode or oz create ContractA, it always deploys ContractA and ContractB together even after specifying only 1 contract which I need to deploy. 

Looks like it still not fixed with the new release or am I doing something wrong with create.

frangio:
Hi @darshan0071990. This was not included in the 2.7 release. It will be part of the 2.8 release.

PR #1427, opened 2020-02-06
Update solc-wrapper version

### LINKED ISSUE #1406: Compile command fails with Solidity 0.6+
Compiling a project with solc 0.6.1:

```bash
$ ozdev compile
✖ Compiling contracts with solc 0.6.1 (commit.e6f7d5a4)
soljson.Pointer_stringify is not a function
```

spalladino:
Prevents "soljson.Pointer_stringify is not a function" error

Fixes #1406

### DIFF
--- a/packages/cli/solc/solc
+++ b/packages/cli/solc/solc
@@ -14,6 +14,7 @@ var path = require('path');
 var wrapper = require('solc-wrapper');
 var binary = require('./soljson-v0.5.9+commit.e560f70d');
 var solc = wrapper(binary);
+var solcCompile = solc.compile ? solc.compile : solc.compileStandardWrapper;
 
 // FIXME: remove annoying exception catcher of Emscripten
 //        see https://github.com/chriseth/browser-solidity/issues/167
@@ -61,7 +62,7 @@ function abort (msg) {
 
 if (argv['standard-json']) {
   var input = fs.readFileSync(process.stdin.fd).toString('utf8');
-  var output = solc.compileStandardWrapper(input);
+  var output = solcCompile(input);
 
   console.log(output);
   process.exit(0);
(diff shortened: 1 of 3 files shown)


frangio:
I'm not familiar with the compiler part of the CLI. Does this continue to work with solc 0.5? Would appreciate a short explanation of how this works.

spalladino:
@frangio `solc-js` is a project maintained by the Solidity team that does two things:
- Ships with a javascript version of the compiler, generated with emscripten
- Provides a user-friendly wrapper for interacting with it

We have forked this project into our own `solc-wrapper`, which is exactly the same but without the emscripten compiler. Instead, we wait for the user to decide on a compiler version (or we pick the latest one), download the emscripten version, and load it using the wrapper.

It seems that something changed in 0.6 which required an update on the wrapper. It's still compatible with 0.5 (I tested it manually, and based on the integration tests we have, should also work with 0.4).

frangio:
The CI is failing. It seems like we should change `compileStandardWrapper` into just `compile`.

See https://github.com/ethereum/solc-js/commit/656a40e880d693a34d695bd95b915f2114935ff5

spalladino:
Right on it! By the way, I accidentally dismissed your review. Mind re-approving?

packages/cli/solc/solc
```diff
@@ -14,6 +14,7 @@ var path = require('path');
 var wrapper = require('solc-wrapper');
 var binary = require('./soljson-v0.5.9+commit.e560f70d');
 var solc = wrapper(binary);
+var solcCompile = solc.compile ? solc.compile : solc.compileStandardWrapper;
```
frangio: Is this necessary? From reading the source it seems like the name is always `compile`.

https://github.com/spalladino/solc-js/blob/e1645d9140133e3f7ba42b29b91c9863103e9b21/wrapper.js#L326

spalladino: Probably not, but wanted to be covered.

PR #1433, opened 2020-02-06
Remove clashing contract

frangio:
One of our tests requires having two contracts with clashing names. Our own compiler doesn't handle this gracefully, and this sometimes causes the CI to fail.

I've removed the file that causes the clash and skipped this test for now.

I will fix the compiler in a separate PR.

Note: fixing the compiler will not immediately fix this issue in the CI because we use an older version of the CLI to compile. This is why I went with this approach.


spalladino:
Since 2.7, our compiler will handle repeated contract names (see https://github.com/OpenZeppelin/openzeppelin-sdk/pull/1411). I'm fine with commenting out this test for now, but I'd add it back once we start using 2.7 for bootstrap.

PR #1437, opened 2020-02-06
Fix check command

### LINKED ISSUE #1436: Command `oz check` fails
Running `oz check` on 2.7 RC2 fails with:

```
$ npx oz check
Cannot destructure property `blockTimeout` of 'undefined' or 'null'.
```

spalladino:
Allow commands that do not require a web3 connection to create Contract instances.

Fixes #1436

This PR is **not** part of #1438, will need a separate merge once we release 2.7 final.

### DIFF
--- a/packages/lib/src/artifacts/ZWeb3.ts
+++ b/packages/lib/src/artifacts/ZWeb3.ts
@@ -37,7 +37,7 @@ declare module 'web3-eth' {
 export default class ZWeb3 {
   public static provider: provider;
   public static web3instance: Web3;
-  public static web3Options: Web3Options;
+  public static web3Options: Web3Options = {};
 
   public static initialize(provider: provider, web3Options: Web3Options = {}): void {
     ZWeb3.provider = provider;


frangio:
This may work but it's not at all good idea. Since the `web3Instance` is cached, if at _any_ point we access `ZWeb3.web3` before initializing the options they will never be set. We already ran into this in the context of `deploy`, and the fix we came up with was #1408, which was an ugly compromise.

We can merge this as a temporary fix but we have to fix this soon. Let me know if you've considered what I mention in the previous paragraph, and whether you think it's okay to merge anyway.

spalladino:
@frangio thanks for the review. Yes, I had evaluated #1408 before implementing this solution. Given I wanted this to be fixed in 2.7, I wanted to introduce the least invasive change. And pre-initializing `web3options` ensures we have exactly the same behaviour as in 2.6 (where we did not have `web3options` at all). Also, note that the cached `web3instance` depends on the `provider` as well, not just `web3options`.

In other words, with this change, if it worked in 2.6, it should work in 2.7 (and vice-versa!). I agree on including a more robust fix in either 2.8 or 3.0.

Issue #1399, opened 2020-01-17
Implement verify for non upgradeable deploys


frangio:
The `verify` command only verifies proxy implementations. We should extend it to also verify non-upgradeable deploys.

PR #1443, opened 2020-02-07
Migrate verify to the new CLI framework

nventuro:
This is a first step in also extending `verify` to regular deploys (#1399).

The migration exposed an inconsistency in the command: on interactive mode, the CLI asks if optimizations were enabled if the option isn't present, but on non interactive mode, `--optimizer` not being passed is interpreted as optimizations being disabled.

In the new framework, any scenario that would trigger a prompt in interactive mode causes an error when interactivity is disabled. Under this conditions, it was no longer possible to specify the optimizer being off with `--no-interactive`, which is why `--optimizer` now has a boolean value.

Additionally, the signature of the `verify` action (the closest thing to the old verify script) changed from `verify(contract, options)` to `verify(options)`, as required by the new command framework. I adapted the tests to reflect this.

Other than that, I tried to not modify too much the backend of the verifier. `verify` now has better typings, reflecting CLI behavior (such as `apiKey` only being asked for if the `remote` is Etherscan), but I didn't propagate this change all the way into `Verifier.ts` to keep this PR focused.

Finally, I had quite a bit of trouble understanding why the call to `ConfigManager.initNetworkConfiguration()` needed to be placed as an `after` action and how the tests rely on this fact, but that is likely me not being familiar with the CLI codebase.


frangio:
>Finally, I had quite a bit of trouble understanding why the call to `ConfigManager.initNetworkConfiguration()` needed to be placed as an after action and how the tests rely on this fact, but that is likely me not being familiar with the CLI codebase.

It's not you, it's poorly documented.

It doesn't need to be placed as an `after` action. In `deploy`, the config init needs to run before the `details` function of the `arguments` argument, so we put it as an `after` action once all the required parameters are collected. How it relates to the tests is that we should not initialize the config during tests. In your case, I think you should favor putting the init config call in the action, skipping it if in tests.

See the way it was done in `deploy` before the `arguments` thing turned out to be an issue.

https://github.com/OpenZeppelin/openzeppelin-sdk/blob/e789b2271789fe80687114459988c232d27b37e5/packages/cli/src/commands/deploy/action.ts#L41-L42

https://github.com/OpenZeppelin/openzeppelin-sdk/blob/e789b2271789fe80687114459988c232d27b37e5/packages/cli/src/commands/deploy/spec.ts#L20-L22

nventuro:
>In your case, I think you should favor putting the init config call in the action, skipping it if in tests.

Ah, I initially considered doing precisely this, but then followed `deploy`'s approach due to not understanding precisely the implications of `initNetworkConfig`. Thanks!

packages/cli/src/commands/verify/spec.ts
```diff
@@ -0,0 +1,129 @@
+import { ParamDetails, Arg, Option } from '../../register-command';
+
+import { TxParams } from '@openzeppelin/upgrades';
+import NetworkFile from '../../models/files/NetworkFile';
+import { DEFAULT_TX_TIMEOUT } from '../../models/network/defaults';
+
+import ProjectFile from '../../models/files/ProjectFile';
```
frangio: The spec file is always loaded even if the `verify` command wasn't the one invoked. To avoid loading unnecessary files why we should do dynamic `await import`s inside the `async` functions below.

packages/cli/src/register-command.ts
```diff
@@ -17,8 +17,8 @@ export interface ParamDetails {
   prompt?: string;
   promptType?: 'confirm' | 'list' | 'input';
   choices?: readonly Choice[];
-  preselect?: string;
-  validationError?: (value: string | boolean) => string | undefined;
+  preselect?: string | boolean;
+  validationError?: (value: string | boolean) => string;
```
frangio: Why did you remove `undefined`?

packages/cli/src/commands/verify/spec.ts
```diff
@@ -0,0 +1,129 @@
...
+
+      return {
+        prompt: 'Pick a contract to verify',
+        choices: contracts,
+      };
+    },
+  },
+];
+
+export const options: Option[] = [
```
frangio: This is missing the `--no-interactive` option. Can you take the one from `deploy` and put it in a place where it can be reused here?

packages/cli/src/commands/verify/action.ts
```diff
@@ -0,0 +1,26 @@
...
+  }
+
+  const controller = new NetworkController(params.network, params.txParams, params.networkFile);
+
+  controller.checkLocalContractDeployed(params.contract, true);
+
+  await controller.verifyAndPublishContract(
+    params.contract,
+    params.optimizer,
+    (params as any).optimizerRuns ?? 200,
```
frangio: Is the `as any` necessary because the property only exists if `optimizer` is true? This is undoing all the benefits of having that sophisticated type, so I think it would be better to use a simpler interface.

packages/cli/src/commands/verify/spec.ts
nventuro: This is used to create a new `ProjectFile`, which is in turn used to provide the defaults for the `optimizer` and `optimizer-runs` settings. I placed it here since this is the only common scope to both of their `details` functions.

We could refactor this into importing the file in both `details`, creating the new `ProjectFile` in each if needed, or putting this into a function that memoizes these values. WDYT?

packages/cli/src/register-command.ts
nventuro: This was likely a leftover of my exploration with `validationError`. Thanks for spotting it, and sorry!

packages/cli/src/commands/verify/spec.ts
nventuro: It's not missing it, it's just at the very end (which I find leads to a better `--help` message). I'll add a file with command utils.

nventuro: I added a simple `utils` file with this declaration: I suspect we'll want to reuse not only options but also common validations, such as checking if a value is an address, a number, etc.

Note we could also almost remove the duplication of the `network` option, but they are different due to the `after` in deploy. I thought about making this parametrized, but didn't want to complicate things unnecessarily. Up to you!

packages/cli/src/commands/verify/action.ts
nventuro: This is what I meant by not wanting to propagate these changes in the front end all the way to the `Verifier` module. 

The new types nicely reflect the guarantees placed by the new command framework: I'd rather make these initial steps in the right direction and then use them while refactoring the rest of the verify callstack.

packages/cli/src/commands/verify/spec.ts
frangio: I think it's okay to import in both and create it in both separately.

packages/cli/src/commands/verify/action.ts
frangio: Okay, sounds good. Ideally we would say `params as OptimizerTrueOptions` or something. Either way though I'm fine with this. We should put in a comment explaining that we expect the rest of verifier to be refactored to accept the fancier types.

PR #1447, opened 2020-02-10
Add nicer error message to catch verify of regular instances

frangio:
`controller.isContractDeployed` checks if a contract is deployed as an _implementation_ for a proxy. If a contract isn't deployed, there is already an error produced by `checkLocalContractDeployed`. I added a different error which contains a warning about regular instances not being supported in verify yet.

### DIFF
--- a/packages/cli/src/commands/verify/action.ts
+++ b/packages/cli/src/commands/verify/action.ts
@@ -16,7 +16,7 @@ export async function action(params: Options & Args & { dontExitProcess: boolean
 
   if (!controller.isContractDeployed(params.contract)) {
     throw new Error(
-      `Contract '${params.contract}' has no proxies in network '${userNetworkName}'.\n\nVerification of regular instances is not yet supported.`,
+      `Contract '${params.contract}' is not deployed to '${userNetworkName}'.\n\nVerification of regular instances is not yet supported.`,
     );
   }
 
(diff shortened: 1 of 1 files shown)


packages/cli/src/commands/verify/action.ts
```diff
@@ -5,13 +5,21 @@ import ConfigManager from '../../models/config/ConfigManager';
...
   if (process.env.NODE_ENV !== 'test') {
     const { network } = await ConfigManager.initNetworkConfiguration(params);
     Object.assign(params, { network });
   }
 
   const controller = new NetworkController(params.network, params.txParams, params.networkFile);
 
+  if (!controller.isContractDeployed(params.contract)) {
+    throw new Error(
+      `Contract '${params.contract}' has no proxies in network '${userNetworkName}'.\n\nVerification of regular instances is not yet supported.`,
```
nventuro: >Contract '${params.contract}' has no proxies in network '${userNetworkName}

Isn't this inaccurate? The issue isn't that there are no proxies: it's that the implementation was not deployed (`push`ed).

What if we kept the old message, but added the new comment about regular deploys?

>`Contract ${contractAlias} is not deployed to ${this.network}.\n\nVerification of regular deployments is not yet supported`

frangio: It's true that it's inaccurate in the general case. I was thinking about the common case where `push` is done implicitly before `create`. I think your suggestion is fine.

```suggestion
      `Contract '${params.contract}' is not deployed to '${userNetworkName}'.\n\nVerification of regular instances is not yet supported.`,
```

PR #1448, opened 2020-02-11
Feature/remove aliases

### LINKED ISSUE #1489: Remove aliases
Since day 1 the CLI has had the concept of aliases. The idea is that you can assign an alternative name to a contract and its instances. We don’t expect a lot of people to know about this feature because it’s nowadays hidden in the `oz create` workflow, and contracts are almost always referred to by their name in the source code.

We want to fully remove aliases now, since we don’t think they really contribute to a good experience, and they add a lot of complexity internally. We do recognize that aliases as a building block can be used for many purposes, but in my opinion they are probably never the right abstraction.

As an alternative we will add features that we think make better abstractions, prioritizing them based on user needs: #1490, #1491.

ylv-io:
Fixes #1489.

### Commands
This PR removes `alias` feature from CLI. The following commands are affected:
* Add
* Remove
* Create
* Link
* Push
* Update

From now on it all CLI commands take only contract names as inputs.

### ProjectFile
Project file field `contracts` have been changed to array of strings.
```diff
-contracts: { [alias: string]: string };
+contracts: string[]; 
```

### Network file 
Network file field 'contracts' have been changed to contain contract names not aliases.
```diff
-contracts: { [contractAlias: string]: ContractInterface };
+contracts: { [name: string]: ContractInterface };
```

### Tests
Many tests were dependent on aliases for simulating contracts upgrades. This behaviour was replaced by [Sinon mocks](https://github.com/OpenZeppelin/openzeppelin-sdk/blob/134f481cef7997ce54b0b2da03c999ed42f783e1/packages/cli/test/scripts/update.test.js#L119).

### Backward compatibility
Since we changed ProjectFile and NetworkFile we have to provide a way to migrate users from 2.x versions to 3.x. One way to do it is to convert old project files to a new ones. The problem with this approach is that different aliases might be used for the same contract which is not supported with current model. We have to provide an alternative mechanism in 3.x to do that.


ylv-io:
Integration and examples tests do fail right now but I don't want to fix them right now because Transpiler and other 3.0 feature would bring more changes. I'd rather prefer to keep cli/lib tests working and fix all integration tests before first 3.0 release.

spalladino:
> Integration and examples tests do fail right now but I don't want to fix them right now because Transpiler and other 3.0 feature would bring more changes. I'd rather prefer to keep cli/lib tests working and fix all integration tests before first 3.0 release.

I'm fine with this as long as we don't merge into master. Otherwise, if we want to make a fix on a different issue, we have no way to know if it breaks the integration tests or not. Let's create a new branch (could be a `develop` one, or a `transpiler` one) and merge there.

packages/cli/test/prompts/prompt.test.js
```diff
@@ -162,16 +162,16 @@ describe('prompt', function() {
...
         });
       });
 
       context('when looking for not yet added but built contracts', function() {
         it('returns an object with not added contracts', function() {
           const contracts = contractsList('keyName', 'Im a message', 'listy', 'notAdded');
-          contracts.keyName.choices.should.include.members(['Bar', 'Buz']);
-          contracts.keyName.choices.should.not.include('Foo');
+          contracts.keyName.choices.should.include.members(['Buz']);
+          contracts.keyName.choices.should.not.include.members(['Foo', 'Bar']);
```
spalladino: Do we know why `Bar` changed in this test?

packages/cli/src/models/files/NetworkFile.ts
```diff
@@ -37,9 +37,8 @@ interface SolidityLibInterface {
 }
 
 export interface ProxyInterface {
-  alias?: string;
+  contractName?: string;
```
spalladino: As with the `ProjectFile`, if we are changing the scheme of the JSON files, we need to bump the `manifestVersion` and provide an automatic migration. Let's make sure we collect all breaking changes we are doing.

packages/cli/test/scripts/query-signed-deployment.test.js
```diff
@@ -11,15 +11,13 @@ import { helpers } from '@openzeppelin/upgrades';
 import push from '../../src/scripts/push';
 import queryDeployment from '../../src/scripts/query-deployment';
 
-const should = require('chai').should();
-
 describe('query-signed-deployment script', function() {
-  const [owner, another, admin] = accounts;
+  const [owner, , admin] = accounts;
```
spalladino: I didn't even know this was valid ts!

packages/cli/test/scripts/query-signed-deployment.test.js
```diff
@@ -79,7 +77,7 @@ describe('query-signed-deployment script', function() {
       const address = await querySignedDeployment({
         signature,
         salt,
-        contractAlias,
+        contractName: contractName,
```
spalladino:
```suggestion
        contractName,
```

packages/cli/test/scripts/verify.test.js
```diff
@@ -26,7 +26,7 @@ describe('verify script', function() {
 
   describe('validations', function() {
     describe('with invalid package or network files', function() {
-      it('throws error if zOS project is not yet initialized', async function() {
+      it('throws error if oz project is not yet initialized', async function() {
```
spalladino: Nothing gets past you!

packages/lib/src/project/ProxyAdminProject.ts
```diff
@@ -47,11 +47,7 @@ class BaseProxyAdminProject extends BaseSimpleProject {
     contractParams: ContractInterface = {},
   ): Promise<Contract> {
     const { initMethod: initMethodName, initArgs } = contractParams;
-    const { implementationAddress, pAddress, initCallData } = await this._setUpgradeParams(
-      proxyAddress,
-      contract,
-      contractParams,
-    );
+    const { implementationAddress, pAddress } = await this._setUpgradeParams(proxyAddress, contract, contractParams);
```
spalladino: Are we renaming `proxyAddress` into `pAddress`? I'd keep the full name to be honest, it's easier to understand while reading the code.

packages/transpiler/.eslintignore
```diff
@@ -0,0 +1,2 @@
+/lib
```
spalladino: I'd include the transpiler in a separate PR

packages/cli/test/scripts/update.test.js
```diff
@@ -320,14 +364,20 @@ describe('update script', function() {
       it('should upgrade multiple proxies and migrate them', async function() {
         // add non-migratable implementation for WithLibraryImpl contract
         await add({
-          contractsData: [{ name: 'UnmigratableImplV2', alias: 'WithLibraryImpl' }],
+          contracts: ['UnmigratableImplV2'],
           projectFile: this.projectFile,
         });
         await push({ network, txParams, networkFile: this.networkFile });
 
+        const UnmigratableImplV2 = this.networkFile.contract('UnmigratableImplV2').address;
```
spalladino:
```suggestion
        const unmigratableImplV2 = this.networkFile.contract('UnmigratableImplV2').address;
```
I'd reserve upper-case for contract classes, and use lower case for instances (or addresses)

packages/cli/test/scripts/update.test.js
```diff
@@ -320,14 +364,20 @@ describe('update script', function() {
...
+
+        restoreUpdate();
+        stubUpdate(
+          { ImplV1: 'UnmigratableImplV2', WithLibraryImplV1: 'WithLibraryImplV2' },
+          { ImplV1: UnmigratableImplV2, WithLibraryImplV1: this.withLibraryImplV2Address },
```
spalladino:
```suggestion
          { ImplV1: unmigratableImplV2, WithLibraryImplV1: this.withLibraryImplV2Address },
```

ylv-io:
> Also, from what I got to review, it does not seem like we managed to delete so much code (perhaps in `upgrades` it's a different story). Can I ask you for impression, now that you've finished this, on whether the removal of aliases was worth the effort with regards to reducing code complexity (and not regarding DX)?

It certainly reduces complexity because now we have fewer abstractions. It is completely another question would it take less complexity to provide the same features using different abstractions. Regardless, Fran has a strong opinion that aliases have to be removed. My initial suggestion was to move aliases to an isolated layer so they would not block Transpiler integration. 

> Finally, we have been wanting to do smaller PRs, and this one clearly does not fit the bill. However, I'm unsure how it could have been split. Do you have any ideas regarding that?

It is certainly not a small PR and I don't think there is obvious way to make it a small. I'd say every rule has an exception. To change something in a big way would take a big PR. Also huge part of why it is so big is the fact that aliases were spread across entire codebase and not isolated at one place.

packages/cli/src/commands/push.ts
```diff
@@ -78,7 +78,7 @@ async function action(options: any): Promise<void> {
     ...promptDeployDependencies,
   };
 
-  if (contractAlias) pushArguments.contractAliases = [contractAlias];
+  if (contracts) pushArguments.contracts = [contracts];
```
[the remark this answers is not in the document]
ylv-io: Good catch!

packages/cli/src/commands/verify/action.ts
```diff
@@ -10,7 +10,7 @@ export async function action(params: Options & Args & { dontExitProcess: boolean
 
   const controller = new NetworkController(params.network, params.txParams, params.networkFile);
 
-  controller.checkLocalContractDeployed(params.contract, true);
+  controller.throwOrLogErrorForContract(params.contract, true);
```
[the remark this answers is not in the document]
ylv-io: Yeah, I am dissatisfied with `throwOrLogErrorForContract` as well. It can be simplified to `logErrorForContract` because a parameter `throwIfFail` implies that it also can throw.
Funny thing. I only now realized that `checkLocalContractDeployed` stands for check if a contract is local and deployed. It is confusing because it is missing `if/is` parts. The method called `checkIfContractIsLocalOrDeployed` would read much better. To be precise this method checks for three things:
* Contract is in ProjectFile (is local)
* Contract is in NetworkFile (is deployed)
* Deployed contract is up to date `currentBytecode === localBytecode;`
And the method name is missing the third part. I also think using the term `local` for being present in ProjectFile is not the best because if it is local, what is a global contract then? Or remote?
My new take on that is `logErrorIfContractChangedOrNotDeployed`. I know it sounds long but I still have fresh memories of how confusing these methods names can be. If there are many of them it becomes really hard to build a mental map of a project.
It is certainly quite an opinionated topic so I value input on that issue. Let me know what you think about `logErrorIfContractChangedOrNotDeployed`.

packages/cli/src/models/dependency/Dependency.ts
```diff
@@ -89,9 +88,9 @@ export default class Dependency {
     // this should all be handled at the Project level. Consider adding a setImplementations (plural) method
     // to Projects, which handle library deployment and linking for a set of contracts altogether.
 
-    const contracts = toPairs(this.projectFile.contracts).map(([contractAlias, contractName]): [Contract, string] => [
+    const contracts = this.projectFile.contracts.map((contractName): [Contract, string] => [
```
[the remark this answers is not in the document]
ylv-io: There is also a third option to migrate to a new format upon detecting old format. My question is how we should handle aliases if there are two aliases for the same contract?

packages/cli/src/models/local/LocalController.ts
```diff
@@ -127,15 +120,10 @@ export default class LocalController {
...
-    if (contractName) {
-      const contractDataPath = Contracts.getLocalPath(contractName);
-      const { compiler, sourcePath } = fs.readJsonSync(contractDataPath);
-      return { sourcePath, compilerVersion: compiler.version };
-    } else {
-      throw Error(`Could not find ${contractAlias} in contracts directory.`);
-    }
+  public getContractSourcePath(contractName: string): { sourcePath: string; compilerVersion: string } | never {
+    const contractDataPath = Contracts.getLocalPath(contractName);
+    const { compiler, sourcePath } = fs.readJsonSync(contractDataPath);
```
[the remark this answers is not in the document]
ylv-io: Surprisingly I can't recall why it happened.

packages/cli/src/models/network/NetworkController.ts
```diff
@@ -644,14 +636,14 @@ export default class NetworkController {
...
+      const contract = this.contractManager.getContractClass(packageName, contractName);
+      await this.setSolidityLibs(contract);
       this.checkInitialization(contract, initMethod);
-      if (salt) await this._checkDeploymentAddress(salt);
+      if (salt) await this.checkDeploymentAddress(salt);
 
       const createArgs = {
         packageName,
-        contractName: contractAlias,
+        contractName: contractName,
```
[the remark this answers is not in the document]
ylv-io: [Certainly.](https://eslint.org/docs/2.0.0/rules/object-shorthand). Let's make it a separate PR.

packages/cli/test/scripts/update.test.js
```diff
@@ -71,142 +74,188 @@ describe('update script', function() {
...
+          ? Promise.resolve(mappingToAddresses[contractName])
+          : ProxyAdminProject.prototype._getOrDeployOwnImplementation.wrappedMethod.apply(this, [
+              contract,
+              contractName,
+              redeployIfChanged,
+            ]);
+      });
+  };
+
+  const restoreUpdate = () => {
```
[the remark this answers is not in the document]
ylv-io: It is actually [depreciated](https://github.com/sinonjs/sinon/issues/1390) but we use old version of Sinon. If we upgrade Sinon our tests would fail 😅

packages/cli/test/scripts/update.test.js
```diff
@@ -71,142 +74,188 @@ describe('update script', function() {
...
 
-    this.implV2Address = this.networkFile.contract('Impl').address;
-    this.withLibraryImplV2Address = this.networkFile.contract('WithLibraryImpl').address;
+    await push({ contracts, network, txParams, networkFile: this.networkFile });
+
+    this.implV2Address = this.networkFile.contract('ImplV2').address;
+    this.withLibraryImplV2Address = this.networkFile.contract('WithLibraryImplV2').address;
+  };
+
+  const stubUpdate = (mappingToNames, mappingToAddresses) => {
```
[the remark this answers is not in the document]
ylv-io: Thanks. I hope it help use to mock other commands in case we need!

packages/cli/test/prompts/prompt.test.js
ylv-io: Because old implementation used aliases ` sinon.stub(ProjectFile.prototype, 'contracts').get(() => ({ Foo: 'Foo', BarAlias: 'Bar' }));` so `Bar` wouldn't be found because its alias is  `BarAlias`.

packages/cli/src/commands/verify/action.ts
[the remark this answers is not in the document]
spalladino: Maybe `logErrorIfContractNotDeployedLatest`? I'd try to go for the positive: if a contract passes this check, it means it has been deployed and is up to date.

packages/cli/src/models/dependency/Dependency.ts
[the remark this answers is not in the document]
spalladino: > My question is how we should handle aliases if there are two aliases for the same contract?

I'd fail, at least until we decide what to do regarding adding names to instances.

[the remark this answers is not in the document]
frangio: Agree on failing for now. This also depends on what we do about support for ethereum packages.

packages/cli/src/models/files/ProjectFile.ts
```diff
@@ -33,7 +33,7 @@ interface ProjectFileData {
   manifestVersion?: string;
   zosversion?: string;
   dependencies: { [name: string]: string };
-  contracts: { [alias: string]: string };
+  contracts: string[];
```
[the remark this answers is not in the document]
frangio: Lossless automatic migration is not possible, but I think we should be able to migrate consolidating all aliases to the actual contract name. We have to do some research on how this will affect people and what we can do to mitigate it.

packages/cli/src/models/network/NetworkController.ts
```diff
@@ -410,24 +409,31 @@ export default class NetworkController {
...
+    if (err) this.logErrorMessage(err, throwIfFail);
   }
 
   // Contract model
-  private _errorForLocalContractsDeployed(): string {
-    const [contractsDeployed, contractsMissing] = partition(this.projectFile.contractAliases, alias =>
-      this.isContractDeployed(alias),
+  private getDeploymentErrorForProject(): string {
+    const contractsMissing = this.projectFile.contracts.filter(
+      o => this.isProjectFileContract(o) && !this.isNetworkFileContract(o),
```
frangio: I'm really curious why you're using the variable name `o`. :thinking:

packages/cli/src/models/network/NetworkController.ts
```diff
@@ -464,40 +463,34 @@ export default class NetworkController {
...
     return !this.networkFile.hasSameBytecode(libClass.schema.contractName, libClass);
   }
 
   // Contract model
-  public hasContractChanged(contractAlias: string, contract?: Contract): boolean {
-    if (!this.isLocalContract(contractAlias)) return false;
-    if (!this.isContractDeployed(contractAlias)) return true;
+  public hasContractChanged(contractName: string, contract?: Contract): boolean {
+    if (!this.isProjectFileContract(contractName)) return false;
+    if (this.isProjectFileContract(contractName) && !this.isNetworkFileContract(contractName)) return true;
```
frangio: The first part of the second condition is implied by the previous line.

```suggestion
    if (!this.isProjectFileContract(contractName)) return false;
    if (!this.isNetworkFileContract(contractName)) return true;
```

It may be clearer using an `if/else`.
```suggestion
    if (!this.isProjectFileContract(contractName)) {
      return false;
    } else if (!this.isNetworkFileContract(contractName)) {
      return true;
    }
```

frangio:
Once we merge #1472 there are likely going to be conflicts in the test files, which were moved to a different directory. I can help resolve them.

packages/cli/test/scripts/update.test.js
frangio: This approach is cool. I don't know if it will work 100% though.

For example, `getContractClass` probably returns something that contains the actual name, rather than the faked one, and I don't know if this can cause issues. On the other hand, this is very similar to what aliases did, so if aliases worked I guess this will work too.

I just think we should keep an eye on this in case it may break, and eventually think of a different way to test upgrades.

packages/cli/test/scripts/update.test.js
```diff
@@ -484,6 +530,7 @@ describe('update script', function() {
   const shouldHandleUpdateOnDependency = function() {
     describe('updating on dependency', function() {
       beforeEach('setup', async function() {
+        this.timeout(10000);
```
frangio: Did you add this in because of the timeouts in CI? Did it fix things? I did not think that the problem was actual lack of time, I was assuming there was something preventing progress.

examples/creating-instances-from-solidity/index.js
```diff
@@ -1,4 +1,4 @@
-'use strict';
+"use strict";
```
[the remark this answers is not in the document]
ylv-io: Yes. My VSCode config has auto-linter enabled on save. I'll be more careful in future.

packages/cli/src/models/network/NetworkController.ts
ylv-io: Personal preference. It is common to use `i` but I find `o` to be better suited because it is not take by loops and and can't be confused with `l`.

packages/cli/test/scripts/update.test.js
ylv-io: Since these are unit tests we have to mock their dependencies. The only question which dependencies and how we mock. I've tried a few options before arriving at this set of mocks.

packages/cli/test/scripts/update.test.js
ylv-io: Yes. It was consistently timing out on my laptop. I think it takes a lot of time because it does multiple `push` and `createProxy` commands which take time on Ganache.

PR #1450, opened 2020-02-11
Update firebase dependency

### LINKED ISSUE #1419: Install fails because of grpc@1.23.3 in node 13.2.0
404 error here

`node-pre-gyp WARN Tried to download(404): https://node-precompiled-binaries.grpc.io/grpc/v1.23.3/node-v79-linux-x64-glibc.tar.gz `

`node-pre-gyp WARN Pre-built binaries not found for grpc@1.23.3 and node@13.2.0 (node-v79 ABI, glibc) (falling back to source compile with node-gyp) `

so it tries to build it and than it has lot of error

```
/home/vikram/.cache/node-gyp/13.2.0/include/node/v8.h:1576:58: error: ‘deprecated’ attribute directive ignored [-Werror=attributes]
Local<Value> export_value);
...
gyp ERR! stack Error: `make` failed with exit code: 2
...
```

..
many more such errors

spalladino:
Updates grpc transitive dependency, which was causing install issues in node 13

Fixes #1419

### DIFF
--- a/packages/cli/package.json
+++ b/packages/cli/package.json
@@ -71,7 +71,7 @@
     "env-paths": "^2.2.0",
     "ethereumjs-util": "^6.1.0",
     "find-up": "^3.0.0",
-    "firebase": "^6.6.0",
+    "firebase": "^7.8.0",
     "fs-extra": "^7.0.1",
     "glob": "^7.1.6",
     "inquirer": "^6.4.1",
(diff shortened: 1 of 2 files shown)


spalladino:
I manually tested the push to firebase and it worked fine. I'm doing this PR on release 2.7 so we can release a 2.7.1 with this fix.

frangio:
Shouldn't this be merged to `master`?

PR #1461, opened 2020-02-14
Additions to deploy integration test

frangio:
I have extended the `deploy` integration test to also test non-regular deployment kinds, and to test constructor/initializer arguments.


tests/deploy/contracts/Answer.sol
```diff
@@ -0,0 +1,7 @@
+pragma solidity ^0.5.0;
+
+contract Answer {
+  function answer() public view returns (uint) {
+    return 42;
```
spalladino: It's funny how I'd have flagged as a change request anything but a 42 here

Issue #1457, opened 2020-02-12
Write tests for new command framework


frangio:
The new command framework implemented in [`register-command.ts`](https://github.com/frangio/openzeppelin-sdk/tree/remove-ts-node) that is now in use by `deploy` and `verify` doesn't have any tests.

It essentially takes care of converting from our own command spec format into commander API calls, and also takes care of handling interactivity. We should write tests for those two things.

Issue #1336, opened 2019-12-13
Implement lightweight offline signing provider


spalladino:
To have support for offline signing, users currently need to use truffle's hdwallet-provider. However, this has lead to bugs like #997 #907 #891 and zeppelinos/zos-cli#334. We should consider implementing a lightweight version of an offline signing provider, that is automatically created given a `mnemonic` or `privateKey` parameter in `network.js`.

PR #1477, opened 2020-02-19
Exit process after running a command

### LINKED ISSUE #1469: `npx oz deploy` hangs when deploying regular contract to Rinkeby
`npx oz deploy` hangs when deploying regular contract to Rinkeby

Deploying [Box.sol](https://docs.openzeppelin.com/learn/developing-smart-contracts#setting-up-a-solidity-project) as a regular contract to Rinkeby hangs.  `ctrl-c` after five minutes to end.
```
$ npx oz deploy
? Choose the kind of deployment regular
? Pick a network rinkeby
? Pick a contract to deploy Box
0xd2990A11ad5AE856059d89bf982B4a83C0E21942
```

cli: `2.8.0-rc.1`
node: `v10.19.0`
npm: `6.13.7`
Operating System: `Windows 10 - WSL`
@truffle/hdwallet-provider: `1.0.31`

## networks.js
```
const { projectId, mnemonic } = require('./secrets.json');
const HDWalletProvider = require('@truffle/hdwallet-provider');

module.exports = {
  networks: {
    development: {
      protocol: 'http',
      host: 'localhost',
      port: 8545,
      gas: 5000000,
      gasPrice: 5e9,
      networkId: '*',
    },
    rinkeby: {
      provider: () => new HDWalletProvider(
        mnemonic, `https://rinkeby.infura.io/v3/${projectId}`
      ),
      networkId: 4,
      gasPrice: 10e9
    }
  },
};
```

Contract appears to be deployed successfully:
https://rinkeby.etherscan.io/address/0xd2990A11ad5AE856059d89bf982B4a83C0E21942

Can interact with contract:
```
$ npx oz send-tx
? Pick a network rinkeby
? Pick an instance Box at 0xd2990A11ad5AE856059d89bf982B4a83C0E21942
? Select which function store(newValue: uint256)
? newValue: uint256: 23
✓ Transaction successful. Transaction hash: 0xf19bcee329d8883872d8929e7737783b5c7efd00fd9d61231fd38ae0ee52ca21
Events emitted:
 - ValueChanged(23)
```
Can deploy upgradeable version successfully:
```
$ npx oz deploy
? Choose the kind of deployment upgradeable
? Pick a network rinkeby
? Pick a contract to deploy Box
All implementations are up to date
? Call a function to initialize the instance after creating it? No
✓ Instance created at 0x2551BB7C13425aCaf9CAA17Ae0EA8caEfdbE734c
0x2551BB7C13425aCaf9CAA17Ae0EA8caEfdbE734c
```
Can deploy minimal version successfully:
```
$ npx oz deploy
? Choose the kind of deployment minimal
? Pick a network rinkeby
? Pick a contract to deploy Box
Minimal proxy support is still experimental.
All implementations are up to date
? Call a function to initialize the instance after creating it? No
✓ Instance created at 0xD3b254165746FDf47a17A707d3e8eC7c5F89e7bC
✓ Deployed ProxyFactory at 0x9306995E929a0615D06f7f20687C667115cfbaE3
0xD3b254165746FDf47a17A707d3e8eC7c5F89e7bC
```
Can deploy regular version to development network:
```
$ npx oz deploy
? Choose the kind of deployment regular
? Pick a network development
? Pick a contract to deploy Box
0x2612Af3A521c2df9EAF28422Ca335b04AdF3ac66
```

spalladino:
Fixes #1469

### DIFF
--- a/packages/cli/src/register-command.ts
+++ b/packages/cli/src/register-command.ts
@@ -76,6 +76,7 @@ export function register(program: Command, spec: CommandSpec, getAction: () => P
       await promptOrValidateAll(cmd, spec, params);
       Telemetry.report(cmd.name(), params as { [key: string]: unknown }, !!params.interactive);
       await action(params);
+      process.exit(0);
     });
 
   for (const opt of spec.options) {


spalladino:
I've already cherry-picked this one to `release/2.8` to publish a new RC with this fix.

ylv-io:
Nice. I was wonder that was the reason it is missing. I guess there was no reason. Also why it is just a `process.exit(0)` but not `  if (!options.dontExitProcess && process.env.NODE_ENV !== 'test') process.exit(0);`?

spalladino:
> Also why it is just a `process.exit(0)` but not `if (!options.dontExitProcess && process.env.NODE_ENV !== 'test') process.exit(0)`?

`dontExitProcess` was used when one action called into another (like `create` calling into `push`), which is no longer the case in the new implementation. And the NODE_ENV check is not needed because this code has no tests at all (see https://github.com/OpenZeppelin/openzeppelin-sdk/issues/1457).

By the way, I'd like for @frangio to check this PR before merging.

frangio:
What is the underlying reason why it's necessary to call `process.exit`? I don't think it's good practice to exit the process like this. There could be pending work in the event queue.

I think someone mentioned once that the problem is in the HD wallet provider. Have we looked into fixing that? Because this is not a problem in CLI, but rather in the user setup, caused by a dependency that they installed and configured (granted, recommended by our documentaiton).

spalladino:
> I don't think it's good practice to exit the process like this. There could be pending work in the event queue.

The bulk of the work is currently awaited in the CLI. But it's true that there could be something else we are not considering which gets aborted.

> Because this is not a problem in CLI, but rather in the user setup, caused by a dependency that they installed and configured (granted, recommended by our documentaiton).

While I agree, it is perceived as a problem with the CLI by our users, especially since other CLIs do include the `process.exit` to handle this provider. I would love to actually build a custom solution (see https://github.com/OpenZeppelin/openzeppelin-sdk/issues/1336) which is more lightweight, but it's on the backlog now. In the meantime, we need to provide a built-in solution.

> Have we looked into fixing that?

In the hdwallet-case, I _think_ the issue is with [`provider-engine`](https://github.com/MetaMask/web3-provider-engine). The `hdwallet-provider` is built using `provider-engine`, which requires to be `stop()`'d once it finishes processing. We could add an action at the end of the CLI that inspects if the current provider implements a `stop` method, and if so, calls it.

However, we had run into issues with other libraries in the past that also have dangling connections (I think a version of truffle-contracts or web3 was one, which came up again during test-env). All in all, just calling `process.exit` was the cleanest way to wrap things up and ensure that the process would not hang, regardless of issues with our (or our users') dependencies.

To sum up, I'm fine with not adding this `process.exit` and fixing this particular case (by adding the `stop()` call), since we have reduced the number of dependencies we need to deal with within the CLI.

frangio:
Thanks for explaining @spalladino. I've created an issue to review this situation. https://github.com/OpenZeppelin/openzeppelin-sdk/issues/1483

PR #1484, opened 2020-02-26
Disable automatic help text in new command framework

frangio:
We are using custom code for the help text (are we using commander for anything at all? :stuck_out_tongue:) and forgot to disable the default in the new command framework. It looked like this:

image

The same commands are already present later in the output.


frangio:
We should backport to 2.8.

Issue #1481, opened 2020-02-25
`Contract undefined not found` with `npx oz deploy` of `@openzeppelin/contracts-ethereum-package/StandaloneERC20`


abcoathup:
Deploying `@openzeppelin/contracts-ethereum-package` with OpenZeppelin CLI 2.8 RC2 gives error `Contract undefined not found` for both a development network and to a public testnet.

Works with OpenZeppelin CLI 2.71 using `oz create`.

node `v10.19.0`
npm `6.13.7`
oz `2.8.0-rc.2`
os: Windows 10 WSL

## Setup
```
$ npx oz init
$ npm i @openzeppelin/cli@rc
$ npx oz link @openzeppelin/contracts-ethereum-package
```
## Deploy (fails)
```
$ npx oz deploy
Nothing to compile, all contracts are up to date.
? Choose the kind of deployment upgradeable
? Pick a network development
? Pick a contract to deploy @openzeppelin/contracts-ethereum-package/StandaloneERC20
Contract undefined not found
```

frangio:
This seems like an error in `oz create` in 2.8 RC as well.

PR #1486, opened 2020-02-28
Fix push functionality when deploying a dependency contract

frangio:
Fixes https://github.com/OpenZeppelin/openzeppelin-sdk/issues/1481. This bug had been introduced in #1426.

Should be cherry-picked into 2.8 release branch once merged.


frangio:
I did think about a test but this error happens at a layer that isn't currently tested I think.

PR #1487, opened 2020-02-29
Reduce asm in Proxy.sol

k06a:


packages/lib/contracts/upgradeability/Proxy.sol
```diff
@@ -28,23 +28,10 @@ contract Proxy {
    * @param implementation Address to delegate.
    */
   function _delegate(address implementation) internal {
+    (bool success, bytes memory data) = implementation.delegatecall(msg.data);
+    require(success, string(data));
```
frangio: This is slightly more complicated because Solidity will add a prefix to the reason in this `require`, but `data` already contains this prefix. The way to work around this is using `revert` in assembly, which should look the same as the `return` statement in the assembly block below.

k06a: Can we use ‘abi.decode(data, (string))`?

frangio: No that wouldn't work. Solidity encodes the data as if it were a function call, with a 4-byte selector at the beginning.

k06a: @frangio wait, I expected `data` to contain revert reason. Can we do parse the prefix like this `abi.decode(data, (bytes4,string))`?

frangio: IIRC That doesn't work because it tries to decode it as if it were function arguments, which are padded differently. It's possible to use the new slices feature from 0.6 but these contracts are still on 0.5. See https://solidity.readthedocs.io/en/latest/060-breaking-changes.html#new-features.

k06a: Yep, found it: https://ethereum.stackexchange.com/a/66404/3032

k06a:
What about a new version? Or you think performance will be different and in this place it is more matter?

frangio:
I'm not sure I understod the question. A new version of what? Do you mean upgrading to Solidity 0.6?

k06a:
@frangio I mean I added new commit to handle this 😁

frangio:
I really like this! I would like someone else to look at it because it's quite critical.

packages/lib/contracts/upgradeability/Proxy.sol
```diff
@@ -28,23 +28,12 @@ contract Proxy {
...
-      returndatacopy(0, 0, returndatasize)
-
-      switch result
-      // delegatecall returns 0 on error.
-      case 0 { revert(0, returndatasize) }
-      default { return(0, returndatasize) }
+      switch success
+        // delegatecall returns 0 on error.
+        case 0 { revert(add(data, 32), returndatasize) }
+        default { return(add(data, 32), returndatasize) }
```
frangio: A small optimization, although the compiler may catch this too.

```suggestion
        let data_start := add(data, 32)
        case 0 { revert(data_start, returndatasize) }
        default { return(data_start, returndatasize) }
```

I also considered using `mload(data)` instead of `returndatasize` but it seems to be more expensive.

k06a: I tried `mload(data)`, but tests failed

frangio: That's interesting. Did you understand why?

k06a: @frangio not yet, will check

PR #1496, opened 2020-03-03
Add proxyInfo test

ylv-io:
More tests!

### DIFF
--- a/packages/cli/src/test/prompts/prompt.test.js
+++ b/packages/cli/src/test/prompts/prompt.test.js
@@ -275,7 +275,7 @@ describe('prompt', function() {
       });
     });
 
-    describe.only('#proxyInfo', function() {
+    describe('#proxyInfo', function() {
       beforeEach('initialize projectFile', function() {
         this.projectFile = new ProjectFile('mocks/mock-stdlib-2/zos.json');
       });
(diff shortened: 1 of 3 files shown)


packages/cli/src/test/prompts/prompt.test.js
```diff
@@ -273,5 +274,37 @@ describe('prompt', function() {
...
       });
     });
+
+    describe('#proxyInfo', function() {
+      beforeEach('initialize projectFile', function() {
+        this.projectFile = new ProjectFile('mocks/mock-stdlib-2/zos.json');
+      });
+
+      this.afterEach('restore stubs', function() {
+        sinon.restore();
```
frangio: This was deprecated, right?

packages/cli/src/test/prompts/prompt.test.js
```diff
@@ -273,5 +274,37 @@ describe('prompt', function() {
         });
       });
     });
+
+    describe('#proxyInfo', function() {
+      beforeEach('initialize projectFile', function() {
+        this.projectFile = new ProjectFile('mocks/mock-stdlib-2/zos.json');
+      });
+
+      this.afterEach('restore stubs', function() {
```
frangio: Did you add `this.` for any reason in particular?

```suggestion
      afterEach('restore stubs', function() {
```

PR #1515, opened 2020-03-25
Configure lerna version bumping behavior

frangio:
This makes persistent the versioning settings that some of us have been using.


spalladino:
Sweet!

PR #1532, opened 2020-04-30
Add async option for send-tx command

### LINKED ISSUE #1531: We need async option for send-tx command
Please add `--async` option for `send-tx` command to avoid awaiting for tx gets mined.

k06a:
Resolves #1531

### DIFF
--- a/packages/cli/src/bin/options.ts
+++ b/packages/cli/src/bin/options.ts
@@ -31,3 +31,7 @@ program.Command.prototype.withNonInteractiveOption = function(): Command {
 program.Command.prototype.withSkipCompileOption = function(): Command {
   return this.option('--skip-compile', 'skips contract compilation');
 };
+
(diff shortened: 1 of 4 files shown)


packages/cli/src/bin/options.ts
```diff
@@ -31,3 +31,7 @@ program.Command.prototype.withNonInteractiveOption = function(): Command {
 program.Command.prototype.withSkipCompileOption = function(): Command {
   return this.option('--skip-compile', 'skips contract compilation');
 };
+
+program.Command.prototype.withAsyncOption = function(): Command {
+  return this.option('--async-tx', 'avoid waiting for tx gets mined');
```
spalladino, line 36:
```suggestion
  return this.option('--async-tx', 'do not wait for the transaction to be mined');
```

PR #1545, opened 2020-05-25
Support quorum network

kigawas:
Not a standard, but widely adopted in its documentation.

https://docs.goquorum.com/en/latest/Getting%20Started/running/#configurable-transaction-size


frangio:
Thanks for the PR @kigawas! Unfortunately I'm going to reject it since we don't have the bandwidth at the moment to properly ensure Quorum compatibility and I feel this may give a false impression in that direction.

kigawas:
@frangio 

NP.
At least it's necessary to make it possible to config custom networks, do you agree?

frangio:
That sounds reasonable if you want to open an issue about it.

PR #1559, opened 2020-06-20
Use posix.relative() instead of path.relative()

### LINKED ISSUE #1558: relatively import path does not work on Windows without WSL
use relatively import on Windows without WSL does not work, for example: https://github.com/yurenju/oz-compile-failed-on-windows

Contract2.sol is imported via `import "./Conntract2.sol"` in Contract1.sol:
```solidity
pragma solidity ^0.5.0;

import "./Contract2.sol";

contract Contract1 {
    uint256 public value;

    function increase() public {
      value++;
    }
}
```

but if run `npm run compile`  on WIndows without WSL, it shows `File import callback not supported`

```
> oz compile
✖ Compiling contracts with solc 0.5.17 (commit.d19bba13)
Compilation errors: 
contracts\Contract1.sol:3:1: ParserError: Source "Contract2.sol" not found: File import callback not supported
import "./Contract2.sol";
^-----------------------^
```

I know you guys recommend developing smart contract on Windows with WSL on [openzeppelin forum](https://forum.openzeppelin.com/t/file-import-callback-not-supported/1341/10), but deploy contract on dev network via openzeppelin cli is not really stable in my experience, I usually need to shutdown WSL to deploy contract if it complain connection refused. 

so it would be great if we can have native windows support for openzeppelin cli without WSL and I actually found a solution, pull request is coming.

yurenju:
fixed #1558 

in solc-js, file path is used as key to manage dependencies, the contract body is retrieved via file path, so if: `import "./Contract2.sol";` appears, solc-js will use "./Contract2.sol" to lookup the body of contract which need to be place into it.

but when we use path.relative() to handle file path, it will transform from "./Contract2.sol" to ".\Contract2.sol" then solc-js cannot find correct contract body, finally throw "File import callback not supported".

so path.posix.relative() should be used in this case to ensure file path follows POSIX rule for a key to index contract body.

### DIFF
--- a/packages/cli/src/models/compiler/solidity/SourcesGatherer.ts
+++ b/packages/cli/src/models/compiler/solidity/SourcesGatherer.ts
@@ -1,5 +1,5 @@
 import { Loggy } from '@openzeppelin/upgrades';
-import { dirname, isAbsolute, join, relative, resolve } from 'path';
+import { dirname, isAbsolute, join, resolve, posix } from 'path';
 import fs from 'fs';
 import { promisify } from 'util';
 import { getImports } from '../../../utils/solidity';
@@ -35,7 +35,7 @@ export async function gatherSources(rootContracts: string[], workingDir: string)
 
   for (const contract of rootContracts) {
     queue.push({
-      name: relative(workingDir, contract),
+      name: posix.relative(workingDir, contract),
       url: resolve(workingDir, contract),
       root: workingDir,
     });


FrozenKiwi:
bump?

abcoathup:
Hi @FrozenKiwi,

We’ve decided it’s best to focus our upgradeability efforts on the Upgrades Plugins exclusively, and have halted development on the OpenZeppelin CLI.

See: https://forum.openzeppelin.com/t/building-for-interoperability-why-we-re-focusing-on-upgrades-plugins/4088

I recommend using OpenZeppelin Upgrades Plugins: https://docs.openzeppelin.com/upgrades-plugins/1.x/

